PR #13, opened 2014-12-08
Cherry-pick chroot updates for 2.10

nh2:
The following commits are updates to `devel/build_chroot` on master but were not picked into 2.10:

- 2cbee1: "Don't check certificates if we check the content"
  - because we don't have a wget invocation yet in 2.10
- a124bb0: "Update hlint in chroot"
  - because the updated `cabal install` invocations are not yet present in 2.10 and it (most probably) doesn't need the newer hlint


bpfoley:
These were committed on stable-2.10. Closing.

PR #16, opened 2014-12-08
HTools: Pass --ignore-soft-errors all the way through

nh2:
Fixes hail silently ignoring the --ignore-soft-errors flag.

See issue 1006.

For hspace and tests, the default is to use defaultOptions
(do not force, like it was before).

The core change is in `allocateOnSingle`/`allocateOnPair`,
which now call `addPriEx force` instead of `addPri` (that one
defaults to `force = False`).


bpfoley:
Committed in Jan 2015. Closing.

PR #17, opened 2014-12-12
Basic support for resource reservations in htools

nh2:
See the patch with subject "HTools: Node violation checks for forthcoming instances" for a description.

Includes a fix by Klaus for making checkData idempotent; let us discuss those in the corresponding thread that he sent earlier. I will then rebase this work on it.

There are a few TODOs to be discussed in the last patch, please let me know your suggestions regarding those.


bpfoley:
Committed in December 2014. Closing.

PR #24, opened 2016-03-06
Add Ed25519 SSH Key Type support (disabled for hostkey currently)

azet:
OpenSSH has, for a while now, supported SSH keys with Ed25519. Host keys
are currently unsupported in Debian Jessie, but that might change in the
future. Thus these functions have been commented-out.

Signed-off-by: Aaron Zauner <azet@azet.org>


bpfoley:
Closing, as discussion moved to the ganeti-devel mailing list.

PR #25, opened 2016-09-29
kvm: use_guest_agent: QEMU Guest Agent support

robbat2:
Implement the QEMU Guest Agent sockets, so that code/scripts on the
hypervisors can communicate with guest operating systems easily.

--
CLA already filed.


bpfoley:
Discussion moved to the ganeti-devel mailing list.

PR #34, opened 2017-03-27
Add optStaticKvmNodeMemory to HTools and IAllocator

vbachraty:
This is an attempt to fix balancing on 2.16 using KVM hypervisor. The problem
with KVM is, that it reports 'active' memory from /proc/memstat for node
memory a.k.a. nMem a.k.a. memory_dom0.

This is plain wrong, as 'active' memory in Linux includes memory from all
running processes, including instances (qemu processes). We can't calculate
the node memory on the hypervisor side, simply subtracting RES used by qemu
processes would give us only approximate value and still it would be the actively
used memory rather than a upper hard limit as in dom0 size in Xen. As KVM
does not provide proper isolation for node memory, the only thing we can do is
to instruct htools to reserve a static value for node memory.

This is solved in 2.17 in patch f971341471 implementing the obtainNodeMemory
function that takes the live dom0 size in Xen, but uses 'hv_state' for exporting
the node size configurable parameter for KVM. I've opted for _not_ backporting
the cluster parameter from 2.17 for these reasons:

* this way the change is isolated to hbal, no need to change cluster config
* the obtainNodeMemory is for Luxi/Rapi only, this flag works with any backend
* the goal is to fix balancing as it is, without adding support for memory
overcommitment, doing so would require backporting a bunch of other dependencies:
d2bfc50608, f971341471 and possibly more.

The flag is ignored on Xen, and on KVM is set by default to 4G, so the change
should be backward compatible (no need to specify this flag unless you need
a different value).

### DIFF
--- a/src/Ganeti/HTools/Backend/IAlloc.hs
+++ b/src/Ganeti/HTools/Backend/IAlloc.hs
@@ -203,8 +203,9 @@ parseGroup u a = do
 -- value.
 parseData :: ClockTime -- ^ The current time
           -> String -- ^ The JSON message as received from Ganeti
+          -> Int -- ^ Static node memory size, see optStaticKvmNodeMemory
           -> Result ([String], Request) -- ^ Result tuple
-parseData now body = do
+parseData now body static_n_mem = do
   decoded <- fromJResult "Parsing input IAllocator message" (decodeStrict body)
   let obj = fromJSObject decoded
       extrObj x = tryFromObj "invalid iallocator message" obj x
@@ -233,8 +234,19 @@ parseData now body = do
       dsrd_loc_tags = extractDesiredLocations ctags
       updateTags = updateExclTags ex_tags .
                    updateDesiredLocationTags dsrd_loc_tags
-  cdata1 <- mergeData [] [] [] [] now (ClusterData gl nl il ctags defIPolicy)
-  let (msgs, fix_nl) = checkData (cdNodes cdata1) (cdInstances cdata1)
+  -- hypervisor
+  enabled_hypervisors <- extrObj "enabled_hypervisors"
+  -- This is ugly, unfortunately the  IAllocator "API" (which is a CLI not an
+  -- API to be precise) is different from other htools backends that receive
+  -- the 'default_hypervisor' field. Normally there is no reason to have
+  -- more than one hypervisor so in practice this should be fine. Ganeti 2.17
+  -- exports the hv_stat per node field with the missing information.
+  -- TODO: Remove, once the caller can get the correct nodes size on KVM.
+  let nl2 = case enabled_hypervisors of
+              hv : _ -> Container.map (`Node.setHypervisor` hv) nl
+              _ -> nl
+  cdata1 <- mergeData [] [] [] [] now (ClusterData gl nl2 il ctags defIPolicy)
+  let (msgs, fix_nl) = updateMissing (cdNodes cdata1) (cdInstances cdata1) static_n_mem
       cdata = cdata1 { cdNodes = fix_nl }
       map_n = cdNodes cdata
       map_i = cdInstances cdata
@@ -484,13 +496,15 @@ processRequest opts request =
            >>= formatAllocateSecondary il
 
 -- | Reads the request from the data file(s).
-readRequest :: FilePath -> IO Request
-readRequest fp = do
+readRequest :: FilePath -- ^ Path to IAllocator input file
+            -> Int -- ^ Static node memory size, see optStaticKvmNodeMemory
+            -> IO Request
+readRequest fp static_n_mem = do
   now <- getClockTime
   input_data <- case fp of
                   "-" -> getContents
                   _   -> readFile fp
-  case parseData now input_data of
+  case parseData now input_data static_n_mem of
     Bad err -> exitErr err
     Ok (fix_msgs, rq) -> maybeShowWarnings fix_msgs >> return rq
 
@@ -512,8 +526,9 @@ runIAllocator :: AlgorithmOptions
 runIAllocator opts request = formatIAllocResult $ processRequest opts request
 
 -- | Load the data from an iallocation request file
-loadData :: FilePath -- ^ The path to the file
+loadData :: FilePath -- ^ Path to IAllocator input file
+         -> Int -- ^ Static node memory size, see optStaticKvmNodeMemory
          -> IO (Result ClusterData)
-loadData fp = do
-  Request _ cdata <- readRequest fp
+loadData fp static_n_mem = do
+  Request _ cdata <- readRequest fp static_n_mem
   return $ Ok cdata
--- a/src/Ganeti/HTools/CLI.hs
+++ b/src/Ganeti/HTools/CLI.hs
@@ -112,6 +112,7 @@ module Ganeti.HTools.CLI
   , oShowVer
   , oShowComp
   , oSkipNonRedundant
+  , oStaticKvmNodeMemory
   , oStdSpec
   , oTargetResources
   , oTieredSpec
@@ -203,6 +204,7 @@ data Options = Options
   , optShowNodes   :: Maybe [String] -- ^ Whether to show node status
   , optShowVer     :: Bool           -- ^ Just show the program version
   , optSkipNonRedundant :: Bool      -- ^ Skip nodes with non-redundant instance
+  , optStaticKvmNodeMemory :: Int    -- ^ Use static value for node memory on KVM
   , optStdSpec     :: Maybe RSpec    -- ^ Requested standard specs
   , optTargetResources :: Double     -- ^ Target resources for squeezing
   , optTestCount   :: Maybe Int      -- ^ Optional test count override
@@ -259,6 +261,7 @@ defaultOptions  = Options
   , optNodeSim     = []
   , optNodeTags    = Nothing
   , optSkipNonRedundant = False
+  , optStaticKvmNodeMemory = 4096
   , optOffline     = []
   , optRestrictToNodes = Nothing
   , optOfflineMaintenance = False
@@ -659,7 +662,7 @@ oNodeTags =
    (ReqArg (\ f opts -> Ok opts { optNodeTags = Just $ sepSplit ',' f })
     "TAG,...") "Restrict to nodes with the given tags",
    OptComplString)
-     
+
 oOfflineMaintenance :: OptType
 oOfflineMaintenance =
   (Option "" ["offline-maintenance"]
@@ -761,6 +764,14 @@ oSkipNonRedundant =
    (NoArg (\ opts -> Ok opts { optSkipNonRedundant = True }))
     "Skip nodes that host a non-redundant instance",
     OptComplNone)
+
+oStaticKvmNodeMemory :: OptType
+oStaticKvmNodeMemory =
+  (Option "" ["static-kvm-node-memory"]
+   (reqWithConversion (tryRead "static node memory")
+    (\i opts -> Ok opts { optStaticKvmNodeMemory = i }) "N")
+   "use static node memory [in MB] on KVM instead of value reported by hypervisor.",
+   OptComplInteger)
 
 oStdSpec :: OptType
 oStdSpec =
--- a/src/Ganeti/HTools/CLI.hs
+++ b/src/Ganeti/HTools/CLI.hs
@@ -770,7 +770,8 @@ oStaticKvmNodeMemory =
   (Option "" ["static-kvm-node-memory"]
    (reqWithConversion (tryRead "static node memory")
     (\i opts -> Ok opts { optStaticKvmNodeMemory = i }) "N")
-   "use static node memory [in MB] on KVM instead of value reported by hypervisor.",
+   "use static node memory [in MB] on KVM instead of value reported by hypervisor.\
+   \ Use 0 to take value reported from hypervisor.",
    OptComplInteger)
 
 oStdSpec :: OptType
--- a/src/Ganeti/HTools/ExtLoader.hs
+++ b/src/Ganeti/HTools/ExtLoader.hs
@@ -60,7 +60,7 @@ import qualified Ganeti.HTools.Backend.Text as Text
 import qualified Ganeti.HTools.Backend.IAlloc as IAlloc
 import qualified Ganeti.HTools.Backend.MonD as MonD
 import Ganeti.HTools.CLI
-import Ganeti.HTools.Loader (mergeData, checkData, ClusterData(..)
+import Ganeti.HTools.Loader (mergeData, updateMissing, ClusterData(..)
                             , commonSuffix, clearDynU)
 import Ganeti.HTools.Types
 import Ganeti.Utils (sepSplit, tryRead, exitIfBad, exitWhen)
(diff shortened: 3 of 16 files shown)


src/Ganeti/HTools/Backend/IAlloc.hs
```diff
@@ -233,8 +234,19 @@ parseData now body = do
...
-  let (msgs, fix_nl) = checkData (cdNodes cdata1) (cdInstances cdata1)
+  -- hypervisor
+  enabled_hypervisors <- extrObj "enabled_hypervisors"
+  -- This is ugly, unfortunately the  IAllocator "API" (which is a CLI not an
+  -- API to be precise) is different from other htools backends that receive
+  -- the 'default_hypervisor' field. Normally there is no reason to have
+  -- more than one hypervisor so in practice this should be fine. Ganeti 2.17
+  -- exports the hv_stat per node field with the missing information.
+  -- TODO: Remove, once the caller can get the correct nodes size on KVM.
+  let nl2 = case enabled_hypervisors of
```
RafaelMarinheiro: I'd prefer if this was some sort of "plugin" or middleware instead of being hardcoded within the function: e.g.: `parseData :: ClockTime -> String -> (JSRecord -> a? -> b?) -> Result ...`. That plugin would then be applied here. However, the return types at this function (Result [String], Request]) are fairly broad, so it is hard to figure out what a? and b? are.

Do you have any ideas?



 The return types are not helping here, however

src/Ganeti/HTools/Backend/IAlloc.hs
```diff
@@ -484,13 +496,15 @@ processRequest opts request =
            >>= formatAllocateSecondary il
 
 -- | Reads the request from the data file(s).
-readRequest :: FilePath -> IO Request
-readRequest fp = do
+readRequest :: FilePath -- ^ Path to IAllocator input file
+            -> Int -- ^ Static node memory size, see optStaticKvmNodeMemory
```
RafaelMarinheiro: Same thing applies here

src/Ganeti/HTools/CLI.hs
```diff
@@ -259,6 +261,7 @@ defaultOptions  = Options
   , optNodeSim     = []
   , optNodeTags    = Nothing
   , optSkipNonRedundant = False
+  , optStaticKvmNodeMemory = 4096
```
RafaelMarinheiro: So it is enabled and set to 4GB by default?

src/Ganeti/HTools/CLI.hs
```diff
@@ -762,6 +765,14 @@ oSkipNonRedundant =
     "Skip nodes that host a non-redundant instance",
     OptComplNone)
 
+oStaticKvmNodeMemory :: OptType
+oStaticKvmNodeMemory =
+  (Option "" ["static-kvm-node-memory"]
+   (reqWithConversion (tryRead "static node memory")
+    (\i opts -> Ok opts { optStaticKvmNodeMemory = i }) "N")
+   "use static node memory [in MB] on KVM instead of value reported by hypervisor.",
```
RafaelMarinheiro: Is there a way to disable this behavior? If so, please add the documentation here.

src/Ganeti/HTools/ExtLoader.hs
```diff
@@ -117,11 +117,15 @@ loadExternalData opts = do
         | setLuxi -> wrapIO . Luxi.loadData $ fromJust lsock
         | setSim -> Simu.loadData simdata
         | setFile -> wrapIO . Text.loadData $ fromJust tfile
-        | setIAllocSrc -> wrapIO . IAlloc.loadData $ fromJust iallocsrc
+        -- IAlloc.loadData calls updateMissing internally because Hail does not
+        -- loadExternalData for loading the JSON config (see wrapReadRequest).
+        -- Here we just pass a 0 as the 'generic' call to updateMissing follows.
+        | setIAllocSrc -> wrapIO . flip IAlloc.loadData 0 $ fromJust iallocsrc
```
RafaelMarinheiro: Why 0? Why not the value specified by the flag?

src/Ganeti/HTools/Loader.hs
```diff
@@ -372,29 +373,49 @@ clearDynU cdata@(ClusterData _ _ il _ _) =
...
-                                  nname delta_mem (delta_dsk `div` 1024):msgs
-                      else msgs
-             in (umsg1, newn)
-        ) [] nl
+-- | Update cluster data to use static node memory on KVM.
+setStaticKvmNodeMem :: Node.List -- ^ Nodes to update
+                       -> Int -- ^ Static node size
+                       -> Node.List -- ^ Updated nodes
+setStaticKvmNodeMem nl static_node_mem =
+  let updateNM n = if Node.hypervisor n == Just Kvm
```
RafaelMarinheiro: Wouldn't guards be better?
```
let updateNM n =
        | Node.hypervisor n == Just Kvm = { Node.nMem = static_node_mem }
        | otherwise = n
```

src/Ganeti/HTools/Loader.hs
```diff
@@ -372,29 +373,49 @@ clearDynU cdata@(ClusterData _ _ il _ _) =
...
+                   then n { Node.nMem = static_node_mem }
+                   else n
+  in Container.map updateNM nl
+
+-- | Update node memory stat based on instance list.
+updateMemStat :: Node.Node -> Instance.List -> Node.Node
+updateMemStat node il =
+    let missing = truncate (Node.tMem node)
```
RafaelMarinheiro: Do you think floor would be better than truncate here? Rounding down is probably safer than rounding up here.

RafaelMarinheiro: Please write this as its own function, i.e. `missingMem :: Node.Node -> Int`

src/Ganeti/HTools/Loader.hs
```diff
@@ -372,29 +373,49 @@ clearDynU cdata@(ClusterData _ _ il _ _) =
...
+    -- This overrides node mem on KVM as loaded from backend. Ganeti 2.17
+    -- handles this using obtainNodeMemory.
+    let nl2 = if static_node_mem > 0
+              then setStaticKvmNodeMem nl static_node_mem
+              else nl
+        updateSingle msgs node =
+            let nname = Node.name node
+                newn = updateMemStat node il
+                delta_mem = Node.xMem newn
+                delta_dsk = truncate (Node.tDsk node)
```
RafaelMarinheiro: Please write this as its own function, i.e. `missingDsk :: Node.Node -> Int`

src/Ganeti/HTools/Loader.hs
```diff
@@ -372,29 +373,49 @@ clearDynU cdata@(ClusterData _ _ il _ _) =
...
+    in node { Node.xMem = missing }
+
+-- | Check the cluster for memory/disk allocation consistency and update stats.
+updateMissing :: Node.List -- ^ All nodes in the cluster
+              -> Instance.List  -- ^ All instances in the cluster
+              -> Int -- ^ Static node memory for KVM
+              -> ([String], Node.List) -- ^ Pair of errors, update node list
+updateMissing nl il static_node_mem =
+    -- This overrides node mem on KVM as loaded from backend. Ganeti 2.17
+    -- handles this using obtainNodeMemory.
```
RafaelMarinheiro: Can we move the static_node_mem > 0 behavior to setStaticKvmNodeMem?

RafaelMarinheiro: Also, can we apply this function in another place?

src/Ganeti/HTools/Loader.hs
```diff
@@ -372,29 +373,49 @@ clearDynU cdata@(ClusterData _ _ il _ _) =
...
+updateMemStat :: Node.Node -> Instance.List -> Node.Node
+updateMemStat node il =
+    let missing = truncate (Node.tMem node)
+                  - Node.nMem node
+                  - Node.fMem node
+                  - nodeImem node il
+    in node { Node.xMem = missing }
+
+-- | Check the cluster for memory/disk allocation consistency and update stats.
+updateMissing :: Node.List -- ^ All nodes in the cluster
```
RafaelMarinheiro: This function is both updating xMem and checking some conditions. Can we split this behavior in two or more functions? i.e.:

`updateMissing :: Node.List -> Instance.List -> Node.List`
`checkData :: Node.List -> Instance.List -> Either [String] () or GenericResult [String] ()`

test/hs/Test/Ganeti/HTools/Node.hs
```diff
@@ -134,6 +134,7 @@ genEmptyOnlineNode =
                        , Node.rMemForth = 0
                        , Node.pRem = 0
                        , Node.pRemForth = 0
+                       , Node.xMem = 0
```
RafaelMarinheiro: Can we completely get rid of xMem? It is calculated from other measured values, so this could be replaced by a function.

src/Ganeti/HTools/CLI.hs
vbachraty: Yes indeed. This provides backward compatibility - i.e. no need to add the flag to every execution of hail, hbal etc. which could be problematic if you have a mixed fleet of 2.16 (i.e. you'd need to detect which cluster was upgraded and set the flag accordingly).

RafaelMarinheiro: I don't know if using this as a default is a good idea, though. If someone is using KVM with machines with a low amount of memory, this change might break their use case. Do we have an estimate of how much memory is needed in order to run KVM? If that number is somewhat close to 4GB, I think I'd be fine with this.

src/Ganeti/HTools/CLI.hs
vbachraty: The value overrides hypervisor reported values if it is greater than 0. Documentation was extended.

src/Ganeti/HTools/Backend/IAlloc.hs
vbachraty: I have followed the same pattern as with other 'sane' htools backends setting the hypervisor parameter just before creating the ClusterData record in parseData. The parsing is backend specific so that plugin would be Json backend specific. 

The ugly thing is that IAllocator interface is broken by design. IAllocator takes a json file for input that has limited (different) semantics compared to Htools backends (Rapi/Luxi/Text etc.) While other backends separate parsing and merging the data, IAllocator does it in one step. I think the support for Text backend in IAllocator was carelessly hacked in. First it parses the json  (r1 in Hail.wrapReadRequest) then if optDataFile (Text backend) or optNodeSim (Simu backend) is specified, the json loaded cluster data is thrown away and a new one is loaded using the standard htools backends. 

I think fixing this would require redesigning IAllocator interface and a significant effort that I think should be a separate fix.

src/Ganeti/HTools/Backend/IAlloc.hs
vbachraty: See my above comment.

src/Ganeti/HTools/ExtLoader.hs
vbachraty: Because it doesn't really matter what value is specified there. IAlloc has a different interface that doesn't use loadExternalData (see readRequest) that's why IAlloc's loadData does not only loads but also calls mergeData/updateMissing. Here however, mergeData/updateMissing follows a few lines below, so the values calculated by IAlloc's loadData are overwritten. This is the consequence of what I described in the first comment.

src/Ganeti/HTools/Loader.hs
vbachraty: I'm not sure what is the preferred style, I don't mind either.

src/Ganeti/HTools/Loader.hs
vbachraty: I'd prefer to use the same formula as before, this is just a small refactoring. A complete rewrite of the body of the function follows in a separate patch.

vbachraty: Please see my above comment.

src/Ganeti/HTools/Backend/IAlloc.hs
RafaelMarinheiro: Ack

src/Ganeti/HTools/Backend/IAlloc.hs
RafaelMarinheiro: Ack

src/Ganeti/HTools/ExtLoader.hs
RafaelMarinheiro: Ack

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: Ack

RafaelMarinheiro: Ack

src/Ganeti/HTools/Loader.hs
vbachraty: The point of this function is to validate missing memory and disk. The side effect is that we need to store the missing memory for later calculations, while we don't store the missing disk. If you insist I could split it down, to multiple smaller functions.

src/Ganeti/HTools/Loader.hs
vbachraty: Moved the condition. I don't think we have many choices where to call this function, bear in mind that it has to work for IAllocator as well that does not use loadExternalData in the generic case, but it does call updateMissing. Also we need to put it somewhere, where we have access to options. This sounded like the best possible compromise. Do you have any ideas?

src/Ganeti/HTools/Loader.hs
vbachraty: I think that should be a separate change, it is not related to KVM memory fixes.

test/hs/Test/Ganeti/HTools/Node.hs
vbachraty: Unfortunately we can't. I tried to, but then I realized, that hbal does a lot of placement simulations when finding optimal cluster allocation. During these steps the state of record is updated, but the state of world obviously isn't. At that point the function would give us different values. That's why we need to calculate the difference between state-of-world/state-of-record beforehand and use this value to predict what the state-of-world would be after those placements happen.

src/Ganeti/HTools/CLI.hs
vbachraty: Maybe we could have an inhouse only patch for the value? I think one would need at least 2G of RAM for the base Linux OS. Also the current node memory as reported by the hypervisor backend is just wrong, even if 4G is too much for someone, at least it is a sane value.

vbachraty: Just rememberd, the default value in 2.17 is also 4G: https://github.com/ganeti/ganeti/blob/stable-2.17/src/Ganeti/ConstantUtils.hs#L236

RafaelMarinheiro: Since 2.16 is still RC, then this is probably fine.

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: Ack.

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: Let's leave it as is for now. I'll try to come up with a refactor for that later.

test/hs/Test/Ganeti/HTools/Node.hs
RafaelMarinheiro: But isn't xMem the difference between the free memory according to the record and the actual free memory? If we update the state of record, the function xMem would change without needing any update, right?

Anyway, it would probably be a better idea to change that in another patch. So let's keep it as is for now.

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: Ack.

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: Let's keep it as is for now. I'll try to revisit this in a follow-up patch.

test/hs/Test/Ganeti/HTools/Node.hs
vbachraty: > xMem would change without needing any update, right?

But we don't want this to change. The name 'missing' is ambigous here (to me it was confusing at least at the beginning). It isn't the memory that is missing in order to accomodate something. It is the memory that went missing - it is being used by some unknown process on the node. During placements we assume that this memory doesn't change as we shuffle instances.

PR #35, opened 2017-03-28
Consider state-of-the-record free memory in htools memory model

vbachraty:
The existing memory model in htools was reporting incorrect node memory,
missing memory and also available (free) memory on KVM that resulted in some
cases in overcramming nodes (overcommiting memory without having swap to evict
pages from RAM) leading to OOM kills.

This model is equivalent to the model proposed in issue 127, but doesn't
introduce overcommitment (swap or balooning) support nor does it depend
on Ganeti 2.17 features.

The goal is to distinguish between reported (state-of-world) values and recorded
(state-of-record) values and not to mix them arbitrarily. To estimate free
memory we always need to take the minimum of measured/calculated values using
those two states (root causes of the differences explained in missing memory) in
order to be on the safe side.

### DIFF
--- a/src/Ganeti/HTools/Cluster.hs
+++ b/src/Ganeti/HTools/Cluster.hs
@@ -226,11 +226,9 @@ updateCStats cs node =
                csFspn = x_fspn, csIspn = x_ispn, csTspn = x_tspn
              }
         = cs
-      inc_amem = Node.fMem node - Node.rMem node
-      inc_amem' = if inc_amem > 0 then inc_amem else 0
+      inc_amem = Node.availMem node
       inc_adsk = Node.availDisk node
-      inc_imem = truncate (Node.tMem node) - Node.nMem node
-                 - Node.xMem node - Node.fMem node
+      inc_imem = Node.iMem node
       inc_icpu = Node.uCpu node
       inc_idsk = truncate (Node.tDsk node) - Node.fDsk node
       inc_ispn = Node.tSpindles node - Node.fSpindles node
@@ -238,13 +236,13 @@ updateCStats cs node =
       inc_acpu = Node.availCpu node
       inc_ncpu = fromIntegral (Node.uCpu node) /
                  iPolicyVcpuRatio (Node.iPolicy node)
-  in cs { csFmem = x_fmem + fromIntegral (Node.fMem node)
+  in cs { csFmem = x_fmem + fromIntegral (Node.unallocatedMem node)
         , csFdsk = x_fdsk + fromIntegral (Node.fDsk node)
         , csFspn = x_fspn + fromIntegral (Node.fSpindles node)
-        , csAmem = x_amem + fromIntegral inc_amem'
+        , csAmem = x_amem + fromIntegral inc_amem
         , csAdsk = x_adsk + fromIntegral inc_adsk
         , csAcpu = x_acpu + fromIntegral inc_acpu
-        , csMmem = max x_mmem (fromIntegral inc_amem')
+        , csMmem = max x_mmem (fromIntegral inc_amem)
         , csMdsk = max x_mdsk (fromIntegral inc_adsk)
         , csMcpu = max x_mcpu (fromIntegral inc_acpu)
         , csImem = x_imem + fromIntegral inc_imem
@@ -774,13 +772,12 @@ iterateAllocSmallStep opts nl il limit newinst allocnodes ixes cstats =
 -- | Guess a number of machines worth trying to put on the cluster in one step.
 -- The goal is to guess a number close to the actual capacity of the cluster but
 -- preferrably not bigger, unless it is quite small (as we don't want to do
--- big steps smaller than 10).
+-- big steps smaller than 20).
 guessBigstepSize :: Node.List -> Instance.Instance -> Int
 guessBigstepSize nl inst =
   let nodes = Container.elems nl
-      totalUnusedMemory = sum $ map Node.fMem nodes
-      reserved = round . maximum $ map Node.tMem nodes
-      capacity = (totalUnusedMemory - reserved) `div` Instance.mem inst
+      totalAvail = sum $ map Node.availMem nodes
+      capacity = totalAvail `div` Instance.mem inst
       -- however, at every node we might lose almost an instance if it just
       -- doesn't fit by a tiny margin
       guess = capacity - Container.size nl
--- a/src/Ganeti/HTools/Loader.hs
+++ b/src/Ganeti/HTools/Loader.hs
@@ -388,11 +388,10 @@ setStaticKvmNodeMem nl static_node_mem =
 -- | Update node memory stat based on instance list.
 updateMemStat :: Node.Node -> Instance.List -> Node.Node
 updateMemStat node il =
-    let missing = truncate (Node.tMem node)
-                  - Node.nMem node
-                  - Node.fMem node
-                  - nodeImem node il
-    in node { Node.xMem = missing }
+  let node2 = node { Node.iMem = nodeImem node il }
+      node3 = node2 { Node.xMem = Node.missingMem node2 }
+  in node3 { Node.pMem = fromIntegral (Node.unallocatedMem node3)
+                         / Node.tMem node3 }
 
 -- | Check the cluster for memory/disk allocation consistency and update stats.
 updateMissing :: Node.List -- ^ All nodes in the cluster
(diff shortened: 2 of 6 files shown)


src/Ganeti/HTools/Cluster.hs
```diff
@@ -774,13 +772,12 @@ iterateAllocSmallStep opts nl il limit newinst allocnodes ixes cstats =
 -- | Guess a number of machines worth trying to put on the cluster in one step.
 -- The goal is to guess a number close to the actual capacity of the cluster but
 -- preferrably not bigger, unless it is quite small (as we don't want to do
--- big steps smaller than 10).
+-- big steps smaller than 20).
```
RafaelMarinheiro: What changed here? Why 20 instead of 10?

src/Ganeti/HTools/Loader.hs
```diff
@@ -388,11 +388,10 @@ setStaticKvmNodeMem nl static_node_mem =
 -- | Update node memory stat based on instance list.
 updateMemStat :: Node.Node -> Instance.List -> Node.Node
 updateMemStat node il =
-    let missing = truncate (Node.tMem node)
-                  - Node.nMem node
-                  - Node.fMem node
-                  - nodeImem node il
-    in node { Node.xMem = missing }
+  let node2 = node { Node.iMem = nodeImem node il }
```
RafaelMarinheiro: Do we have lens support for these objects? It'd be so much cleaner to use that instead.

src/Ganeti/HTools/Node.hs
```diff
@@ -1095,14 +1096,86 @@ availDisk t =
...
+--     - on KVM this might be due to the overhead per qemu process
+--     - on Xen manually upsized domains (xen mem-set)
+-- (+) on KVM non-qemu processes might be using more memory than what is
+--     reserved for node (no isolation)
+-- (-) on KVM qemu processes allocate memory on demand, thus an instance grows
+--     over its lifetime until it reaches state-of-record (+overhead)
+-- (-) on KVM KSM might be active
+-- (-) on Xen manually downsized domains (xen mem-set)
+missingMem :: Node -> Int
+missingMem t =
```
RafaelMarinheiro: I'd just use `missingMem t = recordedFreeMem t - reportedFreeMem t`.

src/Ganeti/HTools/Node.hs
```diff
@@ -1095,14 +1096,86 @@ availDisk t =
...
+-- in the real world the moves would shuffle parts of missing memory as well.
+-- Unfortunately as long as we don't have a more finegrained model that can
+-- better explain missing memory (split down based on root causes), we can't
+-- do better.
+-- NOTE 3: This is a hard limit based on available bytes and our bookkeeping.
+-- In case of memory overcommitment, both recordedFreeMem and reportedFreeMem
+-- would be extended by swap size on KVM or baloon size on Xen (their nominal
+-- and reported values).
+unallocatedMem :: Node -> Int
+unallocatedMem t =
```
RafaelMarinheiro: Your doc mentions that this is the minimum of what is reported by the node and our calculation. So I think it'd be better to write `unnalocatedMem t = min (recordedFreeMem t) (reportedFreeMem t)`.

src/Ganeti/HTools/Node.hs
```diff
@@ -1095,14 +1096,86 @@ availDisk t =
...
 availMem :: Node -> Int
 availMem t =
-  let _f = fMem t
-      _l = rMem t
-  in if _f < _l
-       then 0
-       else _f - _l
+  let reserved = rMem t
+      unallocated = unallocatedMem t
+  in max 0 (unallocated - reserved)
```
RafaelMarinheiro: I think I asked you before, but why bounded to 0? If I did ask you that before, feel free to just ack.

src/Ganeti/HTools/Node.hs
```diff
@@ -1194,9 +1263,10 @@ showField t field =
     "nmem" -> printf "%5d" $ nMem t
     "xmem" -> printf "%5d" $ xMem t
     "fmem" -> printf "%5d" $ fMem t
+    "umem" -> printf "%5d" $ unallocatedMem t
     "imem" -> printf "%5d" $ iMem t
     "rmem" -> printf "%5d" $ rMem t
-    "amem" -> printf "%5d" $ fMem t - rMem t
+    "amem" -> printf "%5d" $ availMem t
```
RafaelMarinheiro: This is not a equivalent change. The previous one was not bounded by 0. It is probably ok considering that this is just a log, but it still good to check.

src/Ganeti/HTools/Cluster.hs
vbachraty: The constant in the code is 20, so I just updated the comment.

src/Ganeti/HTools/Node.hs
vbachraty: Ack

src/Ganeti/HTools/Node.hs
vbachraty: That's how I had it first. But that is wrong because of the same reasons xMem is part of the state not just dynamically calculated. As the state-of-record changes during instance shuffling simulations this would report incorrect values (see Note 1). And verifying this I see I had a bug there - instead of missingMem I should directly use xMem that was calculated at load time.

src/Ganeti/HTools/Cluster.hs
RafaelMarinheiro: Ack.

src/Ganeti/HTools/Node.hs
RafaelMarinheiro: Ack. Can you add that as a comment in the function?

src/Ganeti/HTools/Node.hs
vbachraty: I think it's just by definition. I preserved the definition as it was. I think in the scoring algorithm you want just a sum of non-negative numbers (i.e. if a node doesn't have enough memory for reservations and another one that has some extra available memory they should not cancel out)

src/Ganeti/HTools/Node.hs
vbachraty: The definition in the doc says available memory is fmem - rmem, but the availMem was bounded by 0. Since this is just in a log and the value bounded by 0 is used in score calculations, I tried to unify this. I've also updated the docs (see doc/design-resource-model.rst)

src/Ganeti/HTools/Loader.hs
vbachraty: We don't have lens support for these objects, but I think we could add it. I'm not entirely sure if they would help here.

The problem here is that all those functions (nodeImem, missingMem, unallocatedMem) take a node as an argument but they also depend on each other - i.e. missingMem expects that iMem was updated, unallocatedMem expects xMem to be updated beforehand. Maybe I should have defined those functions taking multiple integer arguments instead?

src/Ganeti/HTools/Node.hs
vbachraty: Actually Note 1 should be it, should I rephrase it? I've updated the respective design doc as well.

vbachraty:
PTAL

src/Ganeti/HTools/Loader.hs
RafaelMarinheiro: I was thinking about lenses because it would simplify each Node -> Node function you have here, so eventually you would be able to write `updateMemStat = fc . fb. fa`. However this is not as streamlined because of the dependency between each one of those. If you wanted to you would be able to use lambdas, but it would be as big as the code here. Feel free to ignore it.

src/Ganeti/HTools/Node.hs
RafaelMarinheiro: Ack

src/Ganeti/HTools/Node.hs
RafaelMarinheiro: Ack

RafaelMarinheiro:
If I understand it correctly, the core difference here is that now the placement algorithm will now take the state-of-the-record free memory into account by using `min(free_state_world, free_state_record)` instead of blindly trusting `free_state_world`. Could you reflect that into the Pull Request title message, i.e. "Consider state-of-the-record free memory in allocation algorithm" or something like that.

vbachraty:
> If I understand it correctly, the core difference here is that now the placement algorithm will now take the > state-of-the-record free memory into account by using min(free_state_world, free_state_record) instead
> of blindly trusting free_state_world. 

Yes, this is the main point of the change. We can't blindly trust neither of state-of-world/state-of-record exclusively as there might be both positive and negative differences termed as 'missing memory'.

src/Ganeti/HTools/Loader.hs
vbachraty: Ack

PR #36, opened 2017-03-28
Merge branch 'stable-2.15' into stable-2.16

vbachraty:
* stable-2.15
  Add test for removing disk on a live instance
  Prohibit disk removal w/o hotplug on live instance
  Fix tuple-unpacking from QueryInstances result
  Fix idx in RemoveDisks warning
  Make man pages more consistent in parameter format
  Re-add 2 imports incorrectly removed during cleanup
  Add status and "version implemented" fields to design docs
  Fix typo in cli_opts.py IGNORE_HVVERSIONS_OPT flag
  Fix coexistence of location tags and non-DRBD instances
  Fix backup export in case of ext disk template
  Fix miscellaneous typos found when reading design docs

* Merge conflicts:

  src/Ganeti/HTools/Cluster.hs # removed block, kept 2.16 version
  man/gnt-instance.rst # slash in new_name
  man/gnt-cluster.rst # dash in cluster-name
  lib/client/gnt_cluster.py # angle brackets
  doc/design-shared-storage-redundancy.rst # removed block, kept 2.16 version
  doc/design-draft.rst # kept 2.16 specific list

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1759,7 +1759,7 @@ Version 2.10.6
 
 *(Released Mon, 30 Jun 2014)*
 
-- Make Ganeti tolerant towards differnt openssl library
+- Make Ganeti tolerant towards different openssl library
   version on different nodes (issue 853).
 - Allow hspace to make useful predictions in multi-group
   clusters with one group overfull (isse 861).
@@ -1900,7 +1900,7 @@ New features
   specified if the destination cluster has a default iallocator.
 - Users can now change the soundhw and cpuid settings for XEN hypervisors.
 - Hail and hbal now have the (optional) capability of accessing average CPU
-  load information through the monitoring deamon, and to use it to dynamically
+  load information through the monitoring daemon, and to use it to dynamically
   adapt the allocation of instances.
 - Hotplug support. Introduce new option '--hotplug' to ``gnt-instance modify``
   so that disk and NIC modifications take effect without the need of actual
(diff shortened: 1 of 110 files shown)


RafaelMarinheiro:
About the CLA: All CLAs were signed. All those commits are already in the Ganeti project here at GitHub, so there is no need to worry about that.

I had to momentarily mark "CLA required" as disabled because, as googlebot pointed out, googlebot will not change its commit status. I will turn that on after that.

vbachraty:
> Note to project maintainer: This is a terminal state, meaning the cla/google commit status will not change > from this state. It's up to you to confirm consent of the commit author(s) and merge this pull request when > appropriate.

I have looked for the error and found that other projects had it and it should be possible to override this. It should be treated more like a warning that github wasn't able to identify the authors. In this particular case I wouldn't worry too much about it - the patches from dimara/yiannis are already in our source tree, this pull request only merges our internal branches. In case we get new patches via github, the pull request submitter will be the same as the author of the patch, so it should be fine as well.

RafaelMarinheiro:
What is commit 8048e563?

vbachraty:
Commit 8048e56 is a revert of 8315239, which was a upmerge from 2.15 to 2.16 mistakenly submitted to 2.15. This means the revert is removing 2.16 features from the incorrectly modified 2.15 - we certainly don't want that in 2.16.

PR #38, opened 2017-03-30
Add cleanup of stale OS hvp data on cluster modify.

Morgawr:
Cluster config would keep track of old/stale OS data even if their
hypervisor parameters are already gone. This check makes sure to wipe
them from the config dictionary if they are empty.

Reviewed-on: https://github.com/ganeti/ganeti/pull/38

### DIFF
--- a/lib/cmdlib/cluster/__init__.py
+++ b/lib/cmdlib/cluster/__init__.py
@@ -1367,6 +1367,12 @@ class LUClusterSetParams(LogicalUnit):
             else:
               self.new_os_hvp[os_name][hv_name].update(hv_dict)
 
+      # Cleanup any OS that has an empty hypervisor parameter list, as we don't
+      # need them in the cluster config anymore.
+      for os_name, hvs in self.new_os_hvp.items():
+        if not hvs:
+          self.new_os_hvp.pop(os_name, None)
+
     # os parameters
     self._BuildOSParams(cluster)
 
--- a/test/py/cmdlib/cluster_unittest.py
+++ b/test/py/cmdlib/cluster_unittest.py
@@ -805,6 +805,22 @@ class TestLUClusterSetParams(CmdlibTestCase):
 
     assert constants.HT_FAKE not in self.cluster.os_hvp["mocked_os"]
 
+  def testRemoveOsFromOsHvpList(self):
+    os_hvp = {
+        "mocked_os_1": {
+            constants.HT_FAKE: {
+                constants.HV_MIGRATION_MODE: constants.HT_MIGRATION_NONLIVE
+            }
+        },
+        "mocked_os_2": {} # This is the one that needs to be removed.
+    }
+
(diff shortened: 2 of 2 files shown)


lib/cmdlib/cluster/__init__.py
```diff
@@ -1367,6 +1367,12 @@ def CheckPrereq(self):
             else:
               self.new_os_hvp[os_name][hv_name].update(hv_dict)
 
+      # Cleanup any OS that has an empty hypervisor parameter list, as we don't
+      # need them in the cluster config anymore.
+      for os_name, hvs in self.new_os_hvp.items():
+        if not hvs:
+          self.new_os_hvp.pop(os_name, None)
+
```
vbachraty: Does this mean in order to remove a OS, we need to remove associated hv params? Could you please add a note in the respective man page?

RafaelMarinheiro: If this is the case, would it break any workflow?

lib/cmdlib/cluster/__init__.py
```diff
@@ -1367,6 +1367,12 @@ def CheckPrereq(self):
             else:
               self.new_os_hvp[os_name][hv_name].update(hv_dict)
 
+      # Cleanup any OS that has an empty hypervisor parameter list, as we don't
+      # need them in the cluster config anymore.
+      for os_name, hvs in self.new_os_hvp.items():
```
RafaelMarinheiro: Are there any tests for this?

lib/cmdlib/cluster/__init__.py
Morgawr: It should not break any workflow because so far it was not possible to clean this file from stale/old OS images that were kept track of in the cluster config data. This change is just making sure that in case there are any, we properly get rid of them.

I'll be looking into the documentation to see if there are any places that I need to update about this change, but I couldn't find any on a cursory look. I don't think we document removing OSs or OS hypervisor parameters anywhere...

lib/cmdlib/cluster/__init__.py
Morgawr: I have added a test, PTAL :)

lib/cmdlib/cluster/__init__.py
Morgawr: I have looked further into this. There is currently no accepted way to "remove" an OS image/definition from within ganeti. We keep OS images at /srv/ganeti/os/ and those can be installed/removed from different workflows (deb packages, custom admin scripts, manually, etc). Then we also have os hypervisor parameters which are recorded in the cluster config data. This patch is taking care of cleaning up the latter (not the former) under the assumption that if you have no hypervisors or hypervisor-specific data for a specific OS image, then the OS image should not be present in the config data in the first place (and we have encountered problems already on dangling OS images from config data from images that we had removed previously)

Morgawr:
PTAL

PR #41, opened 2017-04-27
Add gnt-network mode ext

atta:
this patch allow the user to use the mode=ext within ganeti.
So if a vm is starting, the script "/usr/share/ganeti/extnetwork/<Provider>/up" given by the gnt-network-tag "ext_provider" is triggerd.

We use this to create a vxlan on the node wehn ever it is needed, but a lot of other things are imaginationable.


atta:
i have signed the CLA some time ago but it did not show up
Google Individual CLA Ansgar Jazdzewski Dec 03, 2012 21:57 PST

Morgawr:
Should I be closing this as duplicate of #42 or are you still working on this? If you're planning to submit a future PR with everything implemented, I'll take a look at that and close this one as duplicate of #42, otherwise I can independently review and merge this while waiting for the next one.

iustin:
Ping, what's the status on this? And repeating the question on the disk io patch, should this rather be against master?

atta:
hi can we bing this back up for ganeti 3.0? i will update the PR

apoikos:
@atta, sure!

rbott:
@atta could you try to re-implement this against the current master?

Also it needs documentation (it should be mentioned as an option in the instance parameter list but I think it also deserves some example code. Currently there is some network-related content in the [installation tutorial](https://docs.ganeti.org/docs/ganeti/3.0/html/install.html#id23). Maybe we should in general cover all network modes in the administrator's guide, but that is for another PR to address :-)

rbott:
I will close this now, as the 'ext' mode has been implemented in #1919

PR #42, opened 2017-05-09
Stable 2.16 iops

atta:
add parameter to limit bps and iops for instance-disks

### DIFF
--- a/lib/config/__init__.py
+++ b/lib/config/__init__.py
@@ -1011,6 +1011,8 @@ class ConfigWriter(object):
           link = "route:%s" % nic_link
         elif nic_mode == constants.NIC_MODE_OVS:
           link = "ovs:%s" % nic_link
+        elif nic_mode == constants.NIC_MODE_EXT:
+          link = "ext:%s" % nic_link
         else:
           raise errors.ProgrammerError("NIC mode '%s' not handled" % nic_mode)
 
--- a/lib/config/__init__.py
+++ b/lib/config/__init__.py
@@ -1011,8 +1011,6 @@ class ConfigWriter(object):
           link = "route:%s" % nic_link
         elif nic_mode == constants.NIC_MODE_OVS:
           link = "ovs:%s" % nic_link
-        elif nic_mode == constants.NIC_MODE_EXT:
-          link = "ext:%s" % nic_link
         else:
           raise errors.ProgrammerError("NIC mode '%s' not handled" % nic_mode)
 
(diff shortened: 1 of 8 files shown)


Morgawr:
It looks like there's a double commit that is being picked up by this pull request (probably related to #41 ). This seems to be also conflicting with the CLA. Would you mind cleaning up the branch to make it easier to merge and test for us? 

On a cursory look code looks fine, so if we can get that fixed it should be good.
Also, if you want the CLA bot to not give you issues, you can add your other email account (the spreadshirt one) to your github profile as a secondary email address. That should take care of the CLA problems. Otherwise, please change the patch to point to your gmail/googlemail address instead.

Morgawr:
If you get this fixed, this PR will get merged and override #41, otherwise I can independently review and merge #41 first and then work on the diff for this one. However I think just focusing on a single one (this one) is faster for both of us :)

Morgawr:
Hi, sorry for the slow update on this. I finally took some further time to look into this patch. I can't get it to merge locally to test the changes because the patch has a series of commits that interfere with each other. I think the entire branch has become a bit dirty with the successive commits, could I ask you to clean up the branch (or maybe open a brand new branch) and re-format the commits so you have a clean series of modifications with no intermediate cleanups?

The best way to do this is to apply whatever changes you need to do and then squash them into a single commit with a rebase. This makes it easier to test and review the changes.

atta:
It is OK,

i will create a new request based on a clean patch

yiannist:
Hey everybody,

I know that this is closed but I found some time to cleanup and test this patch and I thought that I should report back. @atta, thanks for your work at this!

The [patch](https://gist.github.com/yiannist/c951f440b4e935319e9dd2653ff52ba9) looks and works great. I tested it on _our_ ganeti (i.e. custom snf-ganeti 2.16) on a testing cluster. I only have two comments:

- The patch should basically go on the `master` branch as `stable-2.16` is already in an RC state (still...).
- The current implementation passes the I/O limits on the spawn command. This approach requires an instance reboot for the changes to be applied. QEMU has the `block_set_io_throttle` command to set those values from a monitor:
```
(qemu) info block
info block
disk-f4d63ea1-8a01-4e3d (#block156): /var/run/ganeti/instance-disks/fixed-iops.me:0 (raw)
    Cache mode:       writeback, direct
    I/O throttling:   bps=0 bps_rd=0 bps_wr=0 bps_max=0 bps_rd_max=0 bps_wr_max=0 iops=80 iops_rd=0 iops_wr=0 iops_max=8 iops_rd_max=0 iops_wr_max=0 iops_size=0 group=disk-f4d63ea1-8a01-4e3d

ide1-cd0: [not inserted]
    Removable device: not locked, tray closed

floppy0: [not inserted]
    Removable device: not locked, tray closed

sd0: [not inserted]
    Removable device: not locked, tray closed
(qemu) help block_set_io_throttle
help block_set_io_throttle
block_set_io_throttle device bps bps_rd bps_wr iops iops_rd iops_wr -- change I/O throttle limits for a block drive
```
It would be very convenient to update the patch to support this also!

PR #44, opened 2017-05-26
Adding links in readme to community resources

chandru9279:
I spent a lot of time installing haskell and cabal dependencies for building ganeti from sources in Centos 6.7 using Install instructions of Ganeti docs

When I discovered RPMs were already available - I realized could have saved a lot of time by using it. 

Also the debootstrap link pointing to code.google.com is broken. Since gnt-cluster verify warns that is required if we have guest debian OS I used the debootstrap rpm from the community site, which works well.

### DIFF
--- a/README
+++ b/README
@@ -1,8 +1,0 @@
-Ganeti 2.17
-===========
-
-For installation instructions, read the INSTALL and the doc/install.rst
-files.
-
-For a brief introduction, read the ganeti(7) manpage and the other pages
-it suggests.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,22 @@
+Ganeti 2.17
+===========
+
+For installation instructions, read the INSTALL and the doc/install.rst
+files.
+
+For a brief introduction, read the ganeti(7) manpage and the other pages
+it suggests.
+
+
+## Community Resources
+
+#### RPM packages
+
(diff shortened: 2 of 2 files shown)


chandru9279:
I signed it!

Morgawr:
Thanks for spotting this, unfortunately after we migrated from code.google.com to github some stuff is still in pending transition and some links may be broken. I appreciate the effort of adding clarity to our documentation.

Ganeti has been generally a "debian" project in the way that we don't really do any testing on centos or use any RPMs officially, so I'm not entirely sure about having direct links to external RPM/centos sites in our readme. Maybe we should point the user to www.ganeti.org website and have those linked there instead as part of community resources?

Morgawr:
Also I'm not sure why the CLA googlebot is complaining again after I updated the branch for the pull request, but we can solve that later.

iustin:
@Morgawr - I think this should be fine to go in the readme, as long as a new paragraph would be added saying explicitly that these (the community resources) are unsupported. What do you think?

atta:
The links are still OK but we will not release ganeti 2.17!
So if we can make shure to have ganeti 3.0.0 rpm-packages at this resources i would be more than happy to add it.

rbott:
I will close this in favor of #1609 1609

PR #45, opened 2017-05-31
kvm: Make use of the -D option

yiannist:
Use a dedicated directory for the QEMU logs (i.e.
/var/log/ganeti/kvm). Along with the -daemonize option
pass the -D option in the KVM command line. QEMU instead
of redirecting stdout/stderr to /dev/null will use the
specified file (i.e. /var/log/ganeti/kvm/<instance name>.log).

Signed-off-by: Dimitris Aragiorgis <dimitris.aragiorgis@gmail.com>
Signed-off-by: Yiannis Tsiouris <tsiou@grnet.gr>


yiannist:
I've signed it.

Morgawr:
Looks like the CLA googlebot is complaining, Yannis I think you need to confirm that you're okay with merging this Pull request.

yiannist:
@Morgawr, I checked with @dimara and we're both OK. Thanks for the review!

PR #55, opened 2017-06-19
hv_xen: generate correct type for paravirtualized nic

rkojedzinszky:
With debian stretch (ganeti 2.15 and xen 4.8) the configuration file generated by ganeti is not correct, type=vif should be specified instead of type=paravirtualized when using paravirtual nics. This is the same with latest ganeti.


rkojedzinszky:
I signed it!

PR #1263, opened 2017-08-02
Move postcopy migration logic to cmdlib

calumcalder:
Exposes hypervisor postcopy functions via RPC, and implements logic to execute them in the MigrateInstance tasklet


lib/cmdlib/instance_migration.py
```diff
@@ -916,6 +917,10 @@ def _ExecMigration(self):
 
     self.feedback_fn("* starting memory transfer")
     last_feedback = time.time()
+
+    migration_caps = \
+      self.instance.hvparams.get(constants.HV_KVM_MIGRATION_CAPS, None)
```
Morgawr: If you return [] instead of None in the get() function when we can't find the KVM migration caps, you don't need to test if the list is empty in the following postcopy_enabled = ... statement.

You can just do

migration_caps = self.instance.hvparams.get(constants.HV_KVM_MIGRATION_CAPS, [])
postcopy_enabled = 'postcopy-ram' in migration_caps

if it were None, it would except obviously so we prevent that here.

This has the advantage that postcopy_enabled will always be a boolean, otherwise in the situation of an empty list it would end up being [], which tests for false but is a different type and I don't like having the type of a variable change based on the context.

lib/server/noded.py
```diff
@@ -710,6 +710,15 @@ def perspective_instance_migrate(params):
     return backend.MigrateInstance(cluster_name, instance, target, live)
 
   @staticmethod
+  def perspective_instance_start_postcopy(params):
+    """ Moves a migrating instance from precopy to postcopy mode
```
Morgawr: Nit: To keep a somewhat consistent docstring, use either "move" or "switch" both here and in the rpc_defs.py file.

lib/cmdlib/instance_migration.py
calumcalder: Good call. hvparams[HV_KVM_MIGRATION_CAPS] is a ':' delimited string, so I'll have it default to an empty string, which should serve the same purpose.

PR #1274, opened 2017-11-23
Feature/limit disk io

atta:
Allow to control the amount of I/O that a Instance can perform


iustin:
So, without looking at the code itself - adding a new feature doesn't seem like the proper thing to do on a stable branch. Would you mind sending this against master instead?

atta:
since we wait so long to get 2.16 released we had to start to move things forward and add features to 2.16 we needed.

i'am fine with adding the feature to 2.17 but please keep ganeti going

saschalucas:
@atta, would you be so kind to PR against master, so that we can merge it for 3.1? The code looks good to me. Spelling and wording in the man page can be improved.

rbott (approved):
LGTM and has passed the QA suite (although there are no specific tests for it of course - but it also does not break anything unrelated).

rbott:
This should be rebased on the current master once #1667 has been merged. It will remove the `-drive` parameter and hence render this patch unusable. From the [QMP Reference Docs](https://qemu.readthedocs.io/en/latest/interop/qemu-qmp-ref.html) it looks like you need to add a "throttle filter driver" in between the blockdev and the disk driver. 

Please also be aware that this needs to be done for regular instance start (`_GenerateKVMBlockDevicesOptions()` in `hv_kvm/__init__.py`) and also for disk hotplugging (`HotAddDisk()` in `hv_kvm/monitor.py`).

atta:
thanks will have a look

PR #1277, opened 2018-01-12
Fix several outdated test files and lint errors

Morgawr:
This commit fixes some issues with distcheck scripts and python linter
problems.

### DIFF
--- a/src/Ganeti/Config.hs
+++ b/src/Ganeti/Config.hs
@@ -185,7 +185,9 @@ getNodeInstances cfg nname =
         sec_insts :: [Instance]
         sec_insts = [inst |
           (inst, disks) <- inst_disks,
-          s_uuid <- mapMaybe (\d -> (instPrimaryNode inst) >>= (computeDiskSecondaryNode d)) disks,
+          s_uuid <- mapMaybe (\d ->
+                              (instPrimaryNode inst) >>=
+                              (computeDiskSecondaryNode d)) disks,
           s_uuid == nname]
     in (pri_inst, sec_insts)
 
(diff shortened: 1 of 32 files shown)


src/Ganeti/Config.hs
```diff
@@ -185,7 +185,9 @@ getNodeInstances cfg nname =
         sec_insts :: [Instance]
         sec_insts = [inst |
           (inst, disks) <- inst_disks,
-          s_uuid <- mapMaybe (\d -> (instPrimaryNode inst) >>= (computeDiskSecondaryNode d)) disks,
+          s_uuid <- mapMaybe (\d ->
+                              (instPrimaryNode inst) >>=
```
iustin: Since you touch this code anyway: I think the parantheses here are not needed, but if lint doesn't complain…

I.e. I think simply

```
mapMaybe (\d -> 
  instPrimaryNode inst >>=
  computeSecondaryNode d) disks
````

would be OK.

File: test/data/htools/hail-alloc-secondary.json [hunk 2]
```diff
@@ -294,5 +294,6 @@
   "request": {
     "name": "instance-to-be-converted-to-drbd",
     "type": "allocate-secondary"
-  }
+  },
+"enabled_hypervisors": ["xen-pvm"]
```
iustin: I think this should also be indented by 2 spaces.

Morgawr: Done.

src/Ganeti/Config.hs
Morgawr: Done.

Morgawr:
The tests were outdated because htools tests did not update the 'enabled_hypervisors' flag, which is required I will try to write a more exhaustive commit message when I merge-squash this PR into the 2.16 branch. Thanks for the review and PTAL.

PR #1278, opened 2018-01-23
Checkpoint Ganeti stable 2.16-rc2

Morgawr:
This is a new release of Ganeti stable 2.16, release candidate 2. 

If this release has no reported outstanding big issues, it will be promoted to stable 2.16 officially.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,111 @@ News
 ====
 
 
+Version 2.16.0 rc2
+------------------
+
+*(Released Mon, 29 Jan 2018)*
+
+Changes since 2.16.0 rc1
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+New features:
+- Do not prompt when force-setting a node online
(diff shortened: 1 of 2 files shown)


iustin:
The NEWS delta seems a bit too verbose; e.g. "Fix hlint errors" doesn't seem to have much value to the end-user (and many others). Would it be possible to trim down the entries to just bug fixes and, if any, new features?

Should I suggest which ones to drop here in the comment? I don't seem to have rights to edit the pull request.

Morgawr:
I agree with the verbosity of the NEWS file. I have change it to reflect a more structured output, merged a few lines and removed some unnecessary ones. I also updated the date. 

Iustin what do you think now?

iustin:
Thanks, this looks much better; LGTM.

Morgawr:
I have updated the NEWS file with the current date (I had some issues with some tests which took more time than I expected). I'll be merging this.

PR #1279, opened 2018-01-30
Typos in gnt-instance man page

regnauld:


man/gnt-instance.rst
```diff
@@ -2084,9 +2084,9 @@ If ``--ignore-ipolicy`` is given any instance policy violations occuring
 during this operation are ignored.
 
 If the ``--cleanup`` option is passed, the operation changes from
-performin a failover to attempting recovery from a failed previous failover.
+performing a failover to attempting recovery from a failed previous failover.
 In this mode, Ganeti checks if the instance runs on the correct node (and
-updates its configuration if not) and ensures the instances' disks
```
kasimon: I think this is correct, it's "instances" (plural) plus an additional s abbreviated to an »'« due to the grammatic case. See https://www.englishclub.com/esl-articles/possessive-apostrophe.htm

iustin: IMHO the fix is correct. `gnt-instance failover` acts on single instances, furthermore the text says "and updates its configuration" and not "their configuration".

PR #1283, opened 2018-02-04
Add SkipTest in hv_kvm psutil-specific test cases

Morgawr:


Morgawr:
It would be cleaner, however we do the psutil check on hv_kvm/__init__.py with a bit of standalone logic to report/bubble up the error (psutil_err). I did not want to replicate the logic in the tests as well, and this seemed the most clean way to deal with this idiosyncrasy in the code.

iustin:
Ah right, then LGTM as is. I guess the old comment ("replace with unittest.skip") was not really applicable then.

PR #1285, opened 2018-02-06
Update Ganeti master branch to build on Ubuntu 18.04

Morgawr:
This series of commits makes it possible to build Ganeti on the latest Ubuntu (18.04). 
There is very little functional changes, mostly just dependency fixes for GHC8 and updated Haskell libraries.

### DIFF
--- a/src/Ganeti/Confd/Client.hs
+++ b/src/Ganeti/Confd/Client.hs
@@ -40,9 +40,11 @@ module Ganeti.Confd.Client
 import Control.Concurrent
 import Control.Exception (bracket)
 import Control.Monad
+import Data.ByteString.Char8 (pack, unpack)
 import Data.List
 import Data.Maybe
 import qualified Network.Socket as S
+import qualified Network.Socket.ByteString as BS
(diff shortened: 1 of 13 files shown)


src/Ganeti/Confd/Client.hs
```diff
@@ -40,9 +40,11 @@ module Ganeti.Confd.Client
 import Control.Concurrent
 import Control.Exception (bracket)
 import Control.Monad
+import Data.ByteString.Char8 (pack, unpack)
```
iustin: Uh, most of the code here is dealing network I/O. In this context, my first thought is that `Data.ByteString.Char8` is not appropriate, and that instead plain `Data.ByteString` should be used. But maybe I'm missing some thing?

iustin:
@Morgawr friendly ping. Could we get maybe the rest of the patches (outside Client.hs and the bytestring change) with proper authorship?

apoikos:
Closing this in favor of #1331. Thanks @Morgawr for your work!

PR #1290, opened 2018-02-28
utils.livelock: use portable struct flock type

apoikos:
From `fcntl(2)`:
```
struct flock {
    ...
    short l_type;    /* Type of lock: F_RDLCK,
                        F_WRLCK, F_UNLCK */
    short l_whence;  /* How to interpret l_start:
                        SEEK_SET, SEEK_CUR, SEEK_END */
    off_t l_start;   /* Starting offset for lock */
    off_t l_len;     /* Number of bytes to lock */
    pid_t l_pid;     /* PID of process blocking our lock
                        (set by F_GETLK and F_OFD_GETLK) */
    ...
};
```
On 64-bit systems, `off_t` is always 64 bits long (`long`/`long long`). On
32-bit systems however, depending on whether large file support is
enabled or not, it may be 64 bits (`long long`) or 32 bits (`long`)
long.

The code in `LiveLock.__init__` would always assume `off_t` to be `long`,
breaking on 32-bit systems with LFS support. Fix this by picking the
correct type to use depending on the existence of `os.O_LARGEFILE`.

Note that LFS is enabled almost universally these days and it would be
safe to just use `long long` unconditionally, but it doesn't harm to
make the actual check.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1995,6 +1995,7 @@ python_tests = \
 	test/py/ganeti.utils.io_unittest-runasroot.py \
 	test/py/ganeti.utils.io_unittest.py \
 	test/py/ganeti.utils.log_unittest.py \
+	test/py/ganeti.utils.livelock_unittest.py \
 	test/py/ganeti.utils.lvm_unittest.py \
 	test/py/ganeti.utils.mlock_unittest.py \
 	test/py/ganeti.utils.nodesetup_unittest.py \
--- a/lib/utils/livelock.py
+++ b/lib/utils/livelock.py
@@ -72,8 +72,16 @@ class LiveLock(object):
     name = "%s_%d" % (name, int(time.time()))
     fname = os.path.join(pathutils.LIVELOCK_DIR, name)
     self.lockfile = open(fname, 'w')
+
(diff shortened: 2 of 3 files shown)


apoikos:
Well, it wouldn't be much to add a test. However, note that the issue at hand was architecture-specific and it wouldn't be covered by the unit test (unless run on every build). Instead, we found out because of failing autopkgtests (which run the QA suite) on Ubuntu.

iustin:
Correct, and I'm not concerned about that particular architecture - I'm rather worried to not break on amd64, hence I'd like to see a test that this still (trivially) works.

And my reason for concern:
```
$ uname -m
x86_64
$ python -c 'import os; print os.O_LARGEFILE'
0
```

Which means on O_LARGEFILE is defined (but 0) even on amd64, which makes me believe this will change behaviour when it shouldn't. Hence, a test would be good :)

apoikos:
It will, yes, but on amd64 `long` == `long long` (at least for the foreseeable future). I'll write a test though, seems pretty straightfordward. And I'll also run the Python unit tests as autopkgtests in the Debian package.

PR #1294, opened 2018-03-05
ovf_unittest: use a temporary output directory

apoikos:
Since this is a temporary working directory for the duration of the
test, use a directory under /tmp. There are a number of reasons why this
is desirable, e.g. to be able to run the unittests with a read-only
source tree, as in the case of Debian autopkgtests.


atta:
nice find

PR #1295, opened 2018-03-05
qa.qa_instance_utils: handle SSConf read race

apoikos:
As WConfd updates and distributes SSConf asynchronously, it may take
some time until config updates hit all nodes. This race has been
triggering sporadic Debian automatic as-installed package test failures.

Rather than check the SSConf instance list once, we use `retry.Retry`
until either the instance list contains the desired instance name, or a
timeout is reached.

### DIFF
--- a/qa/qa_instance_utils.py
+++ b/qa/qa_instance_utils.py
@@ -36,6 +36,8 @@ from ganeti import utils
 from ganeti import constants
 from ganeti import pathutils
 
+from ganeti.utils import retry
+
 import qa_config
 import qa_error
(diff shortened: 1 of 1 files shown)


apoikos:
@iustin: thanks! I thought about checking the serial as well, but it's not being distributed as an ssconf variable. Sleeping and/or retrying is obviously not ideal, but still it will close the race that appears to be in the orders of tens of milliseconds.

iustin:
Yep, agreed. Thanks for fixing!

PR #1296, opened 2018-03-15
fixes for drbd >= 8.4: add drbdsetup new-device and new-mirror for ne…

vladimir-ipatov:
fixes for drbd >= 8.4: add drbdsetup new-device and new-mirror for network part assembling of drbd (needed for diskless states of drbd), relocate syncer parameters setting to local part (in 8.4 syncer require disk-config)

Drbd >= 8.4 comes with changed assemble logic.

In drbd 8.3 there is two parts can be initializated apart.
Local part - drbdsetup attach - attaches lvm volume to drbd minor.

Network part - drbdsetup net - create network connection for drbd minor
Also in network part present syncer parameters set for minor

In drbd 8.4 there is more complicated init.
In normal state (newly created drbd or assemble healphy drbd) old code (adapted to 8.4) working ok.
But in other state (e.g. if drbd diskless in one node - no lvm lv on it) old code can't assemble drbd. There is two reasons:
1) first of all we must initialise resources:
drbdsetup new-resource 
drbdsetup new-minor
in diskless case local part initializing is skipped, so it attempts to init network part, but fails with "no resource"
So I added that commands in network part too (it can be called multiple times without errors even if resource already exist).

2) in 8.4 syncer parameters no longer relates to network part, it relates to local disk parameters, so init without local part fails for this reason too.
I moved syncer init to local part after disk attach. Also I must to add one second pause before syncer init - in some cases disk attach doesn't finish in time.

This patch is a part of SkyCover Infrastructure project - Debian based distro with xen and ganeti virtualization. Overview and download: http://sci.skycover.co

### DIFF
--- a/lib/storage/drbd.py
+++ b/lib/storage/drbd.py
@@ -398,6 +398,19 @@ class DRBD8Dev(base.BlockDev):
       if result.failed:
         base.ThrowError("drbd%d: can't attach local disk: %s",
                         minor, result.output)
+   
+    # syncer init only for drbd => 8.4 - in 8.4 it must be set after local, not in net
+    info = DRBD8.GetProcInfo()
+    version = info.GetVersion()
+    if version["k_minor"] >= 4:
+      sync_errors = self._SetMinorSyncParams(minor, self.params)
+      # try second time because disk config for drbd resource may be still uninitialized
+      if sync_errors:
+        time.sleep(1)
+        sync_errors = self._SetMinorSyncParams(minor, self.params)
+        if sync_errors:
+          base.ThrowError("drbd%d: can't set the synchronization parameters: %s" %
+                          (minor, utils.CommaJoin(sync_errors)))
 
   def _AssembleNet(self, minor, net_info, dual_pri=False, hmac=None,
                    secret=None):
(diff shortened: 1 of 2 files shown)


mweinelt:
When you write >= 8.4, does that also mean >= 9.0?

vladimir-ipatov:
Did not tested with 9.0.
Tested with debian wheezy (backports) - linux-image-3.16.0-5-amd64 and drbd userspace utils 8.9.2 and with debian stretch -  linux-image-4.9.0-5-amd64 and userspace 8.9.10

If there is no changes in syntax and assemble logic in 9.0 since 8.4, it will work with 9.0. But I doubd it.

lib/storage/drbd.py
```diff
@@ -398,6 +398,19 @@ def _AssembleLocal(self, minor, backend, meta, size):
...
                         minor, result.output)
+   
+    # syncer init only for drbd => 8.4 - in 8.4 it must be set after local, not in net
+    info = DRBD8.GetProcInfo()
+    version = info.GetVersion()
+    if version["k_minor"] >= 4:
+      sync_errors = self._SetMinorSyncParams(minor, self.params)
+      # try second time because disk config for drbd resource may be still uninitialized
+      if sync_errors:
+        time.sleep(1)
```
apoikos: It's best to avoid using `time.sleep()` for cases like this. `utils.Retry` with an appropriate timeout works much better.

lib/storage/drbd.py
```diff
@@ -398,6 +398,19 @@ def _AssembleLocal(self, minor, backend, meta, size):
       if result.failed:
         base.ThrowError("drbd%d: can't attach local disk: %s",
                         minor, result.output)
+   
+    # syncer init only for drbd => 8.4 - in 8.4 it must be set after local, not in net
+    info = DRBD8.GetProcInfo()
+    version = info.GetVersion()
+    if version["k_minor"] >= 4:
```
apoikos: I think it's best to abstract this check into a separate function and use it in both places. Something like:

```python
@staticmethod
def _NeedsLocalSyncerParams():
  info = DRBD8.GetProcInfo()
  version = info.GetVersion()
  return version["k_minor"] >= 4
```

Then you could write:

```python
def _AssembleLocal(self, minor, backend, meta, size):
  …
  if self._NeedsLocalSyncerParams():
    sync_errors = …
…

def _AssembleNet(self, minor, net_info, dual_pri=False, hmac=None,
                   secret=None):
  …
  if not self._NeedsLocalSyncerParams():
    sync_errors = …
…

This leads to cleaner code and keeps the version check logic in one place.

File: lib/storage/drbd.py [hunk 3]
```diff
@@ -575,6 +596,7 @@ def _SetMinorSyncParams(self, minor, params):
       logging.error(msg)
       return [msg]
 
+
```
apoikos: I guess this blank line was introduced by accident.

lib/storage/drbd_cmdgen.py
```diff
@@ -345,8 +345,14 @@ def GenLocalInitCmds(self, minor, data_dev, meta_dev, size_mb, params):
 
     return cmds
 
-  def GenNetInitCmd(self, minor, family, lhost, lport, rhost, rport, protocol,
+  def GenNetInitCmds(self, minor, family, lhost, lport, rhost, rport, protocol,
                     dual_pri, hmac, secret, params):
+    cmds = []
+
+    cmds.append(["drbdsetup", "new-resource", self._GetResource(minor)])
```
apoikos: Do these commands also work for 8.3?

mbakke:
@vladimir-ipatov will you be able to finish this before Ganeti 3.0?

vladimir-ipatov:
Hello!
Since I'm not a programmer, requested changes will be difficult to do.
Also there will be a lots of tests after making this changes.
This code version is working well for couple years on some dozens of installation of SkyCover server mirror (http://skycover.co).
In SkyCover server mirror we using forked ganeti 2.10.

So, I propose to include my patch "as is" or make this changes by own. I can help with testing.

lib/storage/drbd_cmdgen.py
mbakke: These commands are inside the `DRBD84CmdGenerator` class, so not relevant for 8.3.

rbott:
I will close this PR in favor of #1496 which picks up the work by @vladimir-ipatov already done here.

PR #1297, opened 2018-03-15
fix for ganeti user_shutdown design for xen.

vladimir-ipatov:
if user_shutdown disabled, set domain poweroff behaviour to destroy, if enabled, set to preserve.

In Ganeti 2.11 new feature was implemented - user shutdown detection. But it was implemented partially. In cluster config exists variable enable user shudown (false by default) that actually does nothing.
Also there is config variable for KVM hypervisor, so KVM users cat enable and disable it.
Bit for Xen new behavior (on_powerof=preserve instead of destroy) was hardcoded, so Xen users can't disable it.

I implemented similar config variable for xen-pvm and xen-hvm hypervisors, which disable new behavior by default, but can be enabled cluster wide or for certain instance.

This patch is a part of SkyCover Infrastructure project - Debian based distro with xen and ganeti virtualization. Overview and download: http://sci.skycover.co


rbott:
Hi @vladimir-ipatov, is this still a valid request? Sorry for the _long_ delay. If it is still valid, I have several comments:

* when you introduce new parameters, please add up/downgrade code to the [cfgupgrade](https://github.com/ganeti/ganeti/blob/master/tools/cfgupgrade) script (e.g. remove the new variable from config data during downgrade)
* also please document the parameter change in the [man page](https://github.com/ganeti/ganeti/blob/master/man/gnt-instance.rst)

Cheers, Rudi

rbott:
I will close this PR due to inactivity. Please feel free to reopen rebased against the current master.

PR #1308, opened 2018-09-26
Improve gnt-node evacuate man page

### LINKED ISSUE #552: gnt-node evacuate does not work as documented in the man page
Originally reported of Google Code with ID 496.

```
The man page says

This command will move instances away from the given node. If
``--primary-only`` is given, only primary instances are evacuated, with
``--secondary-only`` only secondaries. If neither is given, all
instances are evacuated. It works only for instances having a drbd disk
template.

However, the implementation (lib/cmdlib/node.py) literally reads as
follows.

    else:
      # All instances
      assert self.op.mode == constants.NODE_EVAC_ALL
      inst_fn = _GetNodeInstances
      # TODO: In 2.6, change the iallocator interface to take an evacuation mode
      # per instance
      raise errors.OpPrereqError("Due to an issue with the iallocator"
                                 " interface it is not possible to evacuate"
                                 " all instances at once; specify explicitly"
                                 " whether to evacuate primary or secondary"
                                 " instances",
                                 errors.ECODE_INVAL)

This TODO for 2.6 should be fixed eventually, even though it probably won't make it into 2.6 any more...

In the short term, we should at least add the warning to the man page as well.
```
Originally added on 2013-06-20 14:37:01 +0000 UTC.

iustin:
The man page still claims that the `-n` mode can evacuate both primary and secondaries, or that it can evacuate primaries at all. Looking at current code on the master branch, only iallocator can evacuate
primaries (probably because with shared file, new primary can be anywhere so it's not fixed), and nothing can evacuate both.

So let's improve the code and fix #552.

### DIFF
--- a/doc/design-ifdown.rst
+++ b/doc/design-ifdown.rst
@@ -7,38 +7,38 @@ Design for adding ifdown script to KVM
 
 .. contents:: :depth: 4
 
-This is a design document about adding support for an ifdown script responsible
-for deconfiguring network devices and cleanup changes made by the ifup script. The
-first implementation will target KVM but it could be ported to Xen as well
-especially when hotplug gets implemented.
+This is a design document about adding support for an ifdown script
+responsible for deconfiguring network devices and cleanup changes made
+by the ifup script. The first implementation will target KVM but it
+could be ported to Xen as well especially when hotplug gets implemented.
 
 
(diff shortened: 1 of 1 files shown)


iustin:
@apoikos - this is rebased onto new master.

PR #1324, opened 2019-02-07
Fix openvswitch port setup when no untagged VLANs are used

gedia:
OpenVSwitch port setup fails when a port is only configured to carry
tagged VLANs, due to the VLAN's configuration string being malformed.
This patch fixes these cases.

Signed-off-by: George Diamantopoulos <gdiamantopoulos@modulus.gr>

### DIFF
--- a/tools/net-common.in
+++ b/tools/net-common.in
@@ -89,6 +89,7 @@ function setup_ovs {
     # Set up trunk port
     # From gnt-instance man page vlan should be :VLAN_ID[:VLAN_ID2..]
     TRUNKS=${VLAN#.*:}  # remove any access info
+    TRUNKS=${TRUNKS#:}  # remove leading ':', if still present
     [ -n "$TRUNKS" ] && ovs-vsctl set port $INTERFACE trunks=${TRUNKS//:/,}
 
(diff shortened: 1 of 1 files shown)


gedia:
I signed it!

gedia:
I signed it!

apoikos:
@gedia, thanks for the PR!

Since we're lacking tests for net-common, could you give an example of a `vlan` string that triggers this bug?

gedia:
Sure:

`nics: "0:link=vswitch0,mode=openvswitch,vlan=:120:1200"`

We use RAPI to create instances, I'm not sure if this also manifests from the command line, although I expect it does.

apoikos:
Thanks! Would you mind including this in the commit message, so that it's clearly visible what this patch tries to achieve? Other than that, LGTM.

apoikos:
@gedia: If you have some time to address the comment, I'd like to merge this in for v2.16.1 (probably early next week). Thanks!

gedia:
Sorry, it's been a hectic past few days.

I've just force-pushed the patch with an amended commit message.

Issue #1268, opened 2017-09-14
Ganeti virtio-net multiqueue support is (was?) broken


yiannist:
There has been a discussion recently about the multiqueue support in Ganeti's [mailing list](https://groups.google.com/forum/#!topic/ganeti/8_LI7SQqPdg).

**TL;DR:** This feature is broken because, throughout Ganeti's code-base, it is assumed that there is an one-to-one relationship between taps and VM NICs.

Multiqueue support was introduced in 47eca2c618107b09c8474539963fb05f0c692003. Since then, QEMU had some changes and thus the following patch is needed (thanks Andrew!):
```diff
diff --git a/lib/hypervisor/hv_kvm/__init__.py b/lib/hypervisor/hv_kvm/__init__.py
index 9fdae5383..62d7ab9e5 100644
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -569,7 +569,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
   _QMP_RE = re.compile(r"^-qmp\s", re.M)
   _SPICE_RE = re.compile(r"^-spice\s", re.M)
   _VHOST_RE = re.compile(r"^-net\s.*,vhost=on|off", re.M)
-  _VIRTIO_NET_QUEUES_RE = re.compile(r"^-net\s.*,fds=x:y:...:z", re.M)
+  _VIRTIO_NET_QUEUES_RE = re.compile(r"^-net(dev)?\s.*,fds=x:y:...:z", re.M)
   _ENABLE_KVM_RE = re.compile(r"^-enable-kvm\s", re.M)
   _DISABLE_KVM_RE = re.compile(r"^-disable-kvm\s", re.M)
   _NETDEV_RE = re.compile(r"^-netdev\s", re.M)
@@ -1835,9 +1835,9 @@ class KVMHypervisor(hv_base.BaseHypervisor):
           # Check for multiqueue virtio-net support.
           if self._VIRTIO_NET_QUEUES_RE.search(kvmhelp):
             # As advised at http://www.linux-kvm.org/page/Multiqueue formula
-            # for calculating vector size is: vectors=2*N+1 where N is the
+            # for calculating vector size is: vectors=2*N+2 where N is the
             # number of queues (HV_VIRTIO_NET_QUEUES).
-            nic_extra_str = ",mq=on,vectors=%d" % (2 * virtio_net_queues + 1)
[...]

PR #1326, opened 2019-02-07
Fix virtio-net multiqueue support (#1268)

gedia:
As reported in #1268 virtio-net multiqueue support is broken.
This patch implements the solutions proposed therein.

Signed-off-by: George Diamantopoulos <gdiamantopoulos@modulus.gr>

### DIFF
--- a/lib/hypervisor/hv_kvm/netdev.py
+++ b/lib/hypervisor/hv_kvm/netdev.py
@@ -189,6 +189,12 @@ def OpenTap(name="", features=None):
 
     tapfds.append(tapfd)
 
+    if name == "":
+      # Set the tap device name after the first iteration of the loop going over
+      # the number of net queues, if it is not already set. If we don't do this,
+      # a new tap device will be created for each queue.
+      name = struct.unpack("16sh", res)[0].strip("\x00")
+
   # Get the interface name from the ioctl
   ifname = struct.unpack("16sh", res)[0].strip("\x00")
 
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -565,7 +565,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
   _QMP_RE = re.compile(r"^-qmp\s", re.M)
   _SPICE_RE = re.compile(r"^-spice\s", re.M)
   _VHOST_RE = re.compile(r"^-net\s.*,vhost=on|off", re.M)
-  _VIRTIO_NET_QUEUES_RE = re.compile(r"^-net\s.*,fds=x:y:...:z", re.M)
+  _VIRTIO_NET_QUEUES_RE = re.compile(r"^-net(dev)?\s.*,fds=x:y:...:z", re.M)
   _ENABLE_KVM_RE = re.compile(r"^-enable-kvm\s", re.M)
   _DISABLE_KVM_RE = re.compile(r"^-disable-kvm\s", re.M)
   _NETDEV_RE = re.compile(r"^-netdev\s", re.M)
@@ -1738,9 +1738,9 @@ class KVMHypervisor(hv_base.BaseHypervisor):
           # Check for multiqueue virtio-net support.
           if self._VIRTIO_NET_QUEUES_RE.search(kvmhelp):
             # As advised at http://www.linux-kvm.org/page/Multiqueue formula
-            # for calculating vector size is: vectors=2*N+1 where N is the
+            # for calculating vector size is: vectors=2*N+2 where N is the
             # number of queues (HV_VIRTIO_NET_QUEUES).
-            nic_extra_str = ",mq=on,vectors=%d" % (2 * virtio_net_queues + 1)
+            nic_extra_str = ",mq=on,vectors=%d" % (2 * virtio_net_queues + 2)
             update_features["mq"] = (True, virtio_net_queues)
           else:
             raise errors.HypervisorError("virtio_net_queues is configured"


atta:
i just build ganeti with the patch in our staging environment, i will test it in the nexet days.

nice work, Thanks

gedia:
I signed it!

gedia:
I signed it!

lib/hypervisor/hv_kvm/netdev.py
```diff
@@ -189,6 +189,9 @@ def OpenTap(name="", features=None):
 
     tapfds.append(tapfd)
 
+    if ( _ == 0 and name == "" ):
```
apoikos: Since we are now checking the value of `_`, it should get a proper variable name such as `q_idx`. `_` is typically used for throw-away variables whose value we don't care about.

Then again, you probably don't care at all if this is the first iteration, just checking if `name` is blank should be enough.

Also the parentheses here are not necessary :)

apoikos:
@gedia: if you have some time to address the comments, I'd like to merge this in time for v2.16.1 (ideally until early next week). Thanks!

gedia:
@apoikos: I've force-pushed an update. Let me know if more is needed.

lib/hypervisor/hv_kvm/netdev.py
apoikos: I think `if name == ""` is sufficient here.

gedia:
Sorry, I had made the changes but forgot to do 'git add' before 'git commit --amend'. Should be OK now.

apoikos:
Thanks for your contributions @gedia!

regnauld:
Shouldn't https://github.com/ganeti/ganeti/blob/4278cbc9592d1d57d3f221dc5c0ad339b337bb8a/lib/hypervisor/hv_kvm/monitor.py#L518 also be updated to be 2N+2 for the vector ?

(And a typo just above that) - https://github.com/ganeti/ganeti/blob/4278cbc9592d1d57d3f221dc5c0ad339b337bb8a/lib/hypervisor/hv_kvm/monitor.py#L513 - sould -> Should

saschalucas:
> https://github.com/ganeti/ganeti/blob/4278cbc9592d1d57d3f221dc5c0ad339b337bb8a/lib/hypervisor/hv_kvm/monitor.py#L518
> also be updated to be 2N+2 for the vector ?

That is a nice finding. While the [qemu command line assembly](https://github.com/ganeti/ganeti/blob/4278cbc9592d1d57d3f221dc5c0ad339b337bb8a/lib/hypervisor/hv_kvm/__init__.py#L1844) is correct, the hotplug code is not. Sadly the logic is somehow duplicated there.

> (And a typo just above that) -
> 
> https://github.com/ganeti/ganeti/blob/4278cbc9592d1d57d3f221dc5c0ad339b337bb8a/lib/hypervisor/hv_kvm/monitor.py#L513
> - sould -> Should

For both, please create a simple PR, maybe the edit function in github would be enough.

gedia:
Nice catch, although if I'm not mistaken QEMU monitor code is being retired according to #1542

regnauld:
Done, if anything the Debian folks might pick it up :D

regnauld:
Hmm, I think there's another issue (I'm trying to get it working on a 2.16.0-rc2 setup), and I noticed in hv_kvm/__init__.py:

  _VHOST_RE = re.compile(r"^-net\s.*,vhost=on|off", re.M)

1. on|off should be on\|off
2. even corrected, this doesn't match on kvm --help output. I tried on 2.11 and on 4.2.1 - in both versions, the closest would be:
...
         [,br=bridge][,helper=helper][,sndbuf=nbytes][,vnet_hdr=on|off][,vhost=on|off]
...

saschalucas:
> _VHOST_RE = re.compile(r"^-net\s.*,vhost=on|off", re.M)
> 
>     1. on|off should be on\|off

Seems true. According to the [docs](https://docs.python.org/3/library/re.html) `|` is a special character and needs to be escaped.

>     2. even corrected, this doesn't match on kvm --help output. 

Seems also true. With qemu-4.2.1 there is no `^-net\s` having an option `vhost`. Just `^-netdev\s` or more precise `^-netdev tap`.

The existing RE seems useless, because it matches many `off`, like demonstrated with this python2/3 script:

```python
import os
stream = os.popen('kvm --help')
kvmhelp = stream.read()
import re
_VHOST_RE = re.compile(r"^-net\s.*,vhost=on|off", re.M)
print(_VHOST_RE.findall(kvmhelp))
```
will print `['off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off', 'off']`

> I ended up getting it working by using ' _VHOST_RE = re.compile(r"[,vhost=on|off]$", re.M) ' as the RE for checking for vhost support.

That is probably a better solution... where the multiline flag `re.M` can be removed, too.

The problem with the current RE is, that it tries to match on multiline without specifying the `re.S` flag, so that `.` also matches on line break. A multiline RE that matches for me is `_VHOST_RE = re.compile(r"^-netdev tap.*,vhost=on\|off", re.M|re.S)`. The above script will print `['-netdev tap,id=str[,fd=h][,fds=x:y:...:z][,ifname=name][,script=file][,downscript=dfile]\n         [,br=bridge][,helper=helper][,sndbuf=nbytes][,vnet_hdr=on|off][,vhost=on|off'`. Which seems like intended.... and also works down to qemu-2.11, which should be fine for `master`. Feel free to PR for that, maybe with some "story telling" commit message.

Issue #473, opened 2017-06-23
Allow storing kvm machines on hugepages


Ganeti-Issues-Migrator:
Originally reported of Google Code with ID 417.

```
When you have hugepages enabled on nodes, the hugepages are subtracted from 
MemFree.  This makes ganeti think that nodes do not have enough free memory to create instances.

Example (64G ram node)
# free
             total       used       free     shared    buffers     cached
Mem:      65960476   64905924    1054552          0       7400     173216
-/+ buffers/cache:   64725308    1235168
Swap:            0          0          0

# cat /proc/meminfo 
MemTotal:       65960476 kB
MemFree:         1054420 kB
Buffers:            7400 kB
Cached:           173216 kB
SwapCached:            0 kB
Active:            61580 kB
Inactive:         142780 kB
Active(anon):      15424 kB
Inactive(anon):    18612 kB
Active(file):      46156 kB
Inactive(file):   124168 kB
Unevictable:       11228 kB
Mlocked:           11228 kB
SwapTotal:             0 kB
SwapFree:              0 kB
Dirty:                12 kB
Writeback:             0 kB
AnonPages:         35084 kB
Mapped:            11760 kB
Shmem:               428 kB
Slab:              45012 kB
SReclaimable:      17180 kB
SUnreclaim:        27832 kB
KernelStack:        1032 kB
PageTables:         1864 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:      998668 kB
Committed_AS:      68928 kB
VmallocTotal:   34359738367 kB
VmallocUsed:      382136 kB
VmallocChunk:   34325270528 kB
HardwareCorrupted:     0 kB
AnonHugePages:         0 kB
HugePages_Total:   31232
HugePages_Free:    31232
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
DirectMap4k:       53824 kB
DirectMap2M:     5171200 kB
DirectMap1G:    61865984 kB

# gnt-instance startup TEST
Waiting for job 1266 for TEST ...
Job 1266 for TEST has failed: Failure: prerequisites not met for this operation:
error type: insufficient_resources, error details:
Not enough memory on node NODE for starting instance TEST: needed 2048 MiB, available 1194 MiB

# gnt-node list
Node              DTotal DFree MTotal MNode MFree Pinst Sinst
[...]

PR #1327, opened 2019-02-08
Preliminary support for persistent hugepages under KVM (#473)

gedia:
This pull request is an early, I-don't-really-know-what-I'm-doing attempt at allowing KVM instances to use persistent hugepages as memory backend.

My intention for these patches was to check the mem_path hypervisor parameter for each instance, and allow their creation or startup if there are enough unreserved hugepages available to accomodate them. Previously, such actions would fail because only free memory would be taken into account for resource accounting when creating instances. But if persistent hugepages are enabled (via kernel boot command parameter or sysfs, for example), their capacity, even if they are unused, is not included in free memory reporting by the kernel.

With this patch, free memory and free hugepages are calculated separately, and resource accounting is performed accordingly (based on the mem_path parameter being set or not).

Free hugepages are now also reported with 'gnt-node list'.

This pull request is mostly meant as a request for comments from the community and developers involed in this project. Some (successful) preliminary testing has been done, but by no means should this work be considered as stable or suitable for use of any kind at this point.

### DIFF
--- a/lib/backend.py
+++ b/lib/backend.py
@@ -703,6 +703,7 @@ def _GetHvInfo(name, hvparams, get_hv_fn=hypervisor.GetHypervisor):
     - vg_free is the free size of the volume group in MiB
     - memory_dom0 is the memory allocated for domain0 in MiB
     - memory_free is the currently available (free) ram in MiB
+    - hugepages_free is the currently available (unreserved) hugepages capacity in MiB
     - memory_total is the total number of ram in MiB
     - hv_version: the hypervisor version, if available
 
--- a/lib/client/gnt_node.py
+++ b/lib/client/gnt_node.py
@@ -58,7 +58,7 @@ from ganeti.confd import client as confd_client
 _LIST_DEF_FIELDS = [
   "name", "dtotal", "dfree",
   "mtotal", "mnode", "mfree",
-  "pinst_cnt", "sinst_cnt",
+  "hpfree", "pinst_cnt", "sinst_cnt",
   ]
 
 
--- a/lib/cmdlib/instance_migration.py
+++ b/lib/cmdlib/instance_migration.py
@@ -38,6 +38,7 @@ from ganeti import errors
 from ganeti import locking
 from ganeti import hypervisor
 from ganeti.masterd import iallocator
+from ganeti import objects
 from ganeti import utils
 from ganeti.cmdlib.base import LogicalUnit, Tasklet
 from ganeti.cmdlib.common import ExpandInstanceUuidAndName, \
@@ -413,11 +414,12 @@ class TLMigrateInstance(Tasklet):
     if (not self.cleanup and
          (not self.failover or
            self.instance.admin_state == constants.ADMINST_UP)):
+      hvfull = objects.FillDict(cluster.hvparams.get(self.instance.hypervisor, {}), cluster.FillHV(self.instance))
       self.tgt_free_mem = CheckNodeFreeMemory(
           self.lu, target_node_uuid,
           "migrating instance %s" % self.instance.name,
           i_be[constants.BE_MINMEM], self.instance.hypervisor,
-          self.cfg.GetClusterInfo().hvparams[self.instance.hypervisor])
+          hvfull)
     else:
       self.lu.LogInfo("Not checking memory on the secondary node as"
                       " instance will not be started")
--- a/lib/cmdlib/instance_operation.py
+++ b/lib/cmdlib/instance_operation.py
@@ -129,6 +129,8 @@ class LUInstanceStartup(LogicalUnit):
       hv_type.CheckParameterSyntax(filled_hvp)
       CheckHVParams(self, self.cfg.GetInstanceNodes(self.instance.uuid),
                     self.instance.hypervisor, filled_hvp)
+    else:
+      filled_hvp = cluster.FillHV(self.instance)
 
     CheckInstanceState(self, self.instance, INSTANCE_ONLINE)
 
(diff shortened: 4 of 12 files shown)


apoikos:
Thanks for the PR @gedia!

There was a discussion [a while back](https://groups.google.com/forum/#!searchin/ganeti-devel/hugetlbfs%7Csort:date/ganeti-devel/-ty3L5QLioA/FrY5kSmT06AJ) about hugepages which led to a [design doc](https://github.com/ganeti/ganeti/tree/master/doc/design-hugepages-support.rst) that we currently have in the repository. Perhaps it could be of some guidance, but I feel much of it can be discussed anew.

The problem with (preallocated) hugepage support is that it partitions the system memory in two, creating a reserved area that is unavailable to the rest of the system. This is similar to what Xen does when limiting the dom0 memory size, with the difference that under Xen instances are expected to use **just** that reserved memory pool. Now, Ganeti's memory management model is arguably a bit Xen-centric, so it should be able to deal with hugepages if it actually knew about them (as your patch does).

However, problems show up when you try to use *both*, instances backed by hugepages and instances backed by normal pages on the same node. Think about the kinds of restrictions this places on the cluster level, for instance with Ganeti wanting to ensure N+1 redundancy by default,  the hugepage pool size (which is not controlled by Ganeti itself) adds another restriction to the problem; for every hugepage-backed DRBD instance, you would want to have guaranteed hugepages free on the secondary as well to facilitate failovers, and dynamically changing the number of hugepages available on the node is not always possible.

The proposed *memory pools* model in the design doc aims at solving these issues, by introducing a fair bit of complexity (that also requires changes in the iallocator protocol). Personally, I would try a different approach and make hugepage support opportunistic: treat `mem_path` as a soft preference for hugepages and fall back to using standard pages if not enough free hugepages are available. In practice this can be done by QEMU automatically by removing `-mem-prealloc` from the KVM command line. The only other change needed would be to add the free hugepage size to the free memory to avoid startup check issues. This is not a 100% correct solution, but it's simple enough to make sense and it would make hugepages usable tomorrow, for those who want to use them, and not break anything for those who haven't configure `mem_path`.

Note that this solution leaves it up to the administrator to decide how they want to utilize hugepage support: they can either configure all of the node's memory as hugepages so that all instances are backed by them, or configure part of the memory as hugepages and let instances race to use them. The latter of course would result in non-deterministic performance, but so do transparent hugepages which are already [enabled by default since QEMU 1.3.0](https://github.com/qemu/qemu/blob/v1.3.0/exec.c#L2594) :-).

gedia:
Thank you for the insights @apoikos !

I was aware of the design doc (http://docs.ganeti.org/ganeti/2.16/html/design-hugepages-support.html), and in part that's what inspired me to calculate hugapages capacity separately (although only mfree/hp is implemented in this patch).

Unfortunately I am not familiar with ganeti's codebase enough (yet) to pursue a more comprehensive take on hugepages support. My primary motivation behind this was that hugepages is the only way that I know of to ensure memory locality for big VMs that scale across more than one NUMA node. What I mean is, with this patch, I can now do this for an instance on a host node where cores 0:4:8:12:16 are located on physical NUMA node 0 and cores 1:5:9:13:17 on physical NUMA node 1:
```
Hypervisor parameters:

- cpu_mask: 0:4:8:12:16:1:5:9:13:17
- cpu_sockets: 2
- cpu_cores: 5
- cpu_threads: 1
- kvm_extra: -object memory-backend-file,id=ram-node0,size=12288M,mem-path=/dev/hugepages,share=off,merge=off,dump=off,prealloc=off,host-nodes=0,policy=bind -numa node,nodeid=0,cpus=0-4,memdev=ram-node0 -object memory-backend-file,id=ram-node1,size=12288M,mem-path=/dev/hugepages,share=off,merge=off,dump=off,prealloc=off,host-nodes=1,policy=bind -numa node,nodeid=1,cpus=5-9,memdev=ram-node1

Back-end parameters:

- maxmem: 24576
- vcpus: 10
- mem_path: /dev/hugepages

```

This should make a big difference under certain workloads. In our experience, on older debian versions at least, such setups could end up being painfully slow. I'm not sure if resource allocation and thread scheduling is more efficient with newer kernels/qemu versions, but I tend to have a preference for more deterministic configurations. By exposing a virtual NUMA configuration to the guest that matches the physical NUMA architecture, I expect the guest kernel to make optimal decisions on process/thread scheduling and memory allocation.

In our deployment logic for testing of this feature we have disabled hugepages for VMs that have a secondary node configured (i.e. we disable mem_path even if it's set in inventory definitions). Not only would this complicate things further in the ways you described (in that one must reserve memory on the secondary node as well), but I'm also not sure if live migration is handled correctly with HP instances (I came across a couple of patches on qemu-devel discussing this, and it appears some significant fixes went into 2.9 which is one version ahead of what stretch currently ships in stable).

Of course this is an administrator-level decision and this patch does not deal with the complexities of having an instance set up in high-availability mode on a secondary node in any way. Additionally, there's no support for Xen whatsoever.

I'll try and revisit this if I have the time at some point. Any contribution is more than welcome.

apoikos:
@gedia: it is more convenient to rebase your changes on top of the upstream branches (using e.g. `git pull -r` or `git rebase origin/stable-2.16`) while you're working on a PR and use a forced push, rather than merge the updated upstream branches into your working branch. It will make reviewing the PR easier, it will allow you to amend your commits before submitting them (using `git rebase` across merges is not trivial) and will also result in a cleaner git history.

gedia:
Sorry, I abused github's GUI a little to help dissect an issue I was investigating, I forgot I had a PR in progress. I'll fix this and revisit this PR shortly.

PR #1329, opened 2019-02-15
Initiate the 3.0 development cycle

apoikos:
Mark master as 2.18.0~alpha1.

Note that while `2.17.0~beta1` was released, we decided to skip 2.17.0
altogether. Bumping to 2.18 will help avoid confusion and will also
allow `cfgupgrade` to revert any config changes for people using
`2.17.0~beta1`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,8 +1,8 @@
 # Configure script for Ganeti
-m4_define([gnt_version_major], [2])
-m4_define([gnt_version_minor], [16])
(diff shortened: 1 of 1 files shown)


iustin:
Since automatic upgrades and rollbacks won't work anyway, would it make more sense to call it 3.0? I liked Morg's suggestion in the thread.

apoikos:
Switched to `3.0.0~alpha1`

PR #1330, opened 2019-02-15
A few cherrypicks from attic/master

iustin:
There are a lot of commits, and probably many actual bug fixes that will be forgotten in _attic/master_. This is just a very small drop in the bucket…


iustin:
The cherry-picks from attic/master do not change the CLA status, so I'm going to override the labels.

iustin:
Ah, that's a good idea. I'll split this into two then.

iustin:
I've dropped the first two and sent them as #1332, will submit now.

PR #1332, opened 2019-02-16
Cherrypicks for stable-2.16

iustin:
Per discussion in #1330, let's have these in stable-2.16, since they're trivial.


iustin:
Argh, sent against wrong branch, corrected. CLA set to "yes" as these are cherry-picks from committed code.

PR #1333, opened 2019-02-17
Make Sphinx compatible to version 1.7/1.8

YSelfTool:
Sphinx 1.7 changed to output format of `sphinx-build --version`. `configure.ac` uses a regex to check that the sphinx major version is at least one. The first commit changes this regex to match both formats.
Alternatively, we could drop the regex completely, because realistically nobody will try to build the docs with sphinx < 1.0, which was release in 2010.

Also, `sphinx.util.compat.Directive` was deprecated and removed, it should be replaced by `docutils.parsers.rst.Directive`.

With these two commits, I can build (`make`) ganeti and the docs with sphinx 1.8.4

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -526,13 +526,13 @@ else
   # Sphinx exits with code 1 when it prints its usage
   sphinxver=`{ $SPHINX --version 2>&1 || :; } | head -n 3`
 
-  if ! echo "$sphinxver" | grep -q -w -e '^Sphinx' -e '^Usage:'; then
+  if ! echo "$sphinxver" | grep -q -w -i -e '^Sphinx' -e '^Usage:'; then
     AC_MSG_ERROR([Unable to determine Sphinx version])
 
   # Note: Character classes ([...]) need to be double quoted due to autoconf
   # using m4
   elif ! echo "$sphinxver" | grep -q -E \
-       '^Sphinx([[[:space:]]]+|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
+       '^(Sphinx|sphinx-build[[1-9]]?)([[[:space:]]]+|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
     AC_MSG_ERROR([Sphinx 1.0 or higher is required])
   fi
 fi
--- a/lib/build/sphinx_ext.py
+++ b/lib/build/sphinx_ext.py
@@ -42,13 +42,12 @@ import docutils.statemachine
 import docutils.nodes
 import docutils.utils
 import docutils.parsers.rst
+from docutils.parsers.rst import Directive
 
 import sphinx.errors
 import sphinx.util.compat
 import sphinx.roles
 import sphinx.addnodes
-
-s_compat = sphinx.util.compat
 
 orig_manpage_role = None
 
@@ -213,7 +212,7 @@ def _BuildOpcodeResult(op_id):
   return "``%s``" % result_fn
 
 
-class OpcodeParams(s_compat.Directive):
+class OpcodeParams(Directive):
   """Custom directive for opcode parameters.
 
   See also <http://docutils.sourceforge.net/docs/howto/rst-directives.html>.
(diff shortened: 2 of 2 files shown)


lib/build/sphinx_ext.py
```diff
@@ -213,7 +211,7 @@ def _BuildOpcodeResult(op_id):
   return "``%s``" % result_fn
 
 
-class OpcodeParams(s_compat.Directive):
+class OpcodeParams(docutils.parsers.rst.Directive):
```
apoikos: I think it's more readable to just once `from docutils.parsers.rst import Directive` and then use plain `Directive` for all classes.

configure.ac
```diff
@@ -526,13 +526,10 @@ else
   # Sphinx exits with code 1 when it prints its usage
   sphinxver=`{ $SPHINX --version 2>&1 || :; } | head -n 3`
 
-  if ! echo "$sphinxver" | grep -q -w -e '^Sphinx' -e '^Usage:'; then
```
apoikos: This check does not need to go, it just needs a `-i` in grep's invocation.

configure.ac
```diff
@@ -526,13 +526,10 @@ else
...
-    AC_MSG_ERROR([Unable to determine Sphinx version])
-
   # Note: Character classes ([...]) need to be double quoted due to autoconf
   # using m4
-  elif ! echo "$sphinxver" | grep -q -E \
-       '^Sphinx([[[:space:]]]+|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
+  if ! echo "$sphinxver" | grep -q -E \
+       '^([[[:space:]]]+|Sphinx|sphinx-build|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
```
apoikos: The new version of the regex will match a single digit, optionally preceded by spaces. Instead of removing `^Sphinx`, it should probably be `^(Sphinx|sphinx-build)`.

lib/build/sphinx_ext.py
YSelfTool: The file did already import docutils.parsers.rst and use it like that, but I agree that this looks more readable. Is changed.

configure.ac
YSelfTool: Fixed.

configure.ac
YSelfTool: Fixed.

apoikos:
@YSelfTool, unfortunately changing the base branch is not enough, as the PR now wants to merge all changes in master to stable-2.16. You need to rebase your branch on top of stable-2.16:

```
git rebase --onto origin/stable-2.16 sphinx~2 sphinx
```

YSelfTool:
Thanks, that rebase worked.

configure.ac
```diff
@@ -526,13 +526,13 @@ else
...
 
-  if ! echo "$sphinxver" | grep -q -w -e '^Sphinx' -e '^Usage:'; then
+  if ! echo "$sphinxver" | grep -q -w -i -e '^Sphinx' -e '^Usage:'; then
     AC_MSG_ERROR([Unable to determine Sphinx version])
 
   # Note: Character classes ([...]) need to be double quoted due to autoconf
   # using m4
   elif ! echo "$sphinxver" | grep -q -E \
-       '^Sphinx([[[:space:]]]+|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
+       '^(Sphinx|sphinx-build[[1-9]]?) ([[[:space:]]]+|\(sphinx-build[[1-9]]?\)|v)*[[1-9]]\>'; then
```
apoikos: I'm not sure we need `[[1-9]]`here, also you've added a space after the first group that might break things.

YSelfTool: I do need the `[[1-9]]`, it maches the executable name, which on my machine is `sphinx-build2`. I've dropped the space, it should be matched by the following `[[[:space:]]]`.

apoikos: Ah, I see, thanks!

PR #1334, opened 2019-02-18
daemon-util: also match processes by name when using s-s-d

### LINKED ISSUE #1322: daemon-util fails to stop daemons using start-stop-daemon from dpkg 1.19.4
As of dpkg 1.19.4, start-stop-daemon will refuse to target processes identified by PID file only, when the PID file is not owned by root. This is done on security grounds, to prevent a regular user from killing random processes on the system using a user-writable PID file.

This breaks Ganeti's process handling when run with user separation enabled (as is the case for the Debian package). Patch can be found [here](https://salsa.debian.org/ganeti-team/ganeti/blob/debian/2.16.0-4/debian/patches/daemon-util-also-match-processes-by-name.patch).

apoikos:
As of dpkg 1.19.4, start-stop-daemon will refuse to match a running process by pidfile only, when the pidfile is not owned by root. It does so on security grounds, as a user-controlled pidfile could be used to
kill any process on the system.

When running with user separation enabled, most Ganeti daemons write their PID files as regular users, which means that process control with s-s-d is currently broken. Fix this by also matching processes by name,in addition to the PID file.

Fixes #1322.

### DIFF
--- a/daemons/daemon-util.in
+++ b/daemons/daemon-util.in
@@ -262,7 +262,7 @@ check() {
     fi
   elif type -p start-stop-daemon >/dev/null; then
     start-stop-daemon --stop --signal 0 --quiet \
-      --pidfile $pidfile
+      --pidfile $pidfile --name "$name"
   else
     _ignore_error status \
       -p $pidfile \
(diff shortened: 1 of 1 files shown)


apoikos:
@iustin, it can be told to explicitly match the user, but doesn't do that by default.

iustin:
> @iustin, it can be told to explicitly match the user, but doesn't do that by default.

I mean, I would have kept the previous behaviour (allow just PID matching) if the pidfile and process user match. Probably has race conditions though, so I understand the rationale for safety.

Issue #1339, opened 2019-02-19
Ceph/RBD support for ceph mimic


atta:
Ceph has changed the json-ouptut of rbd showmapped so we need to cover it for bdev.py

* http://docs.ceph.com/docs/master/releases/mimic/ (The rbd CLI’s showmapped JSON and XML output has changed.)

```
diff --git a/lib/storage/bdev.py b/lib/storage/bdev.py
index c7d5f3b6d..f866951f2 100644
--- a/lib/storage/bdev.py
+++ b/lib/storage/bdev.py
@@ -1064,8 +1064,12 @@ class RADOSBlockDevice(base.BlockDev):
     except ValueError, err:
       base.ThrowError("Unable to parse JSON data: %s" % err)

+    # since ceph mimic the json output changed from dict to list
+    if type(devices) is dict:
+      devices = devices.values()
+
     rbd_dev = None
-    for d in devices.values(): # pylint: disable=E1103
+    for d in devices: # pylint: disable=E1103
       try:
         name = d["name"]
       except KeyError:

````

apoikos:
Thanks! Just a quick note:

Avoid using `type(devices) is dict` if possible because it might break at some point, e.g. if somebody decides to use a *subclass* of dict for `devices`. Use `isinstance(devices, dict)` instead, which works correctly with inheritance. 

Also, would you mind opening a PR so that we can merge this directly?

atta:
i opened a merge request

https://github.com/ganeti/ganeti/pull/1340

apoikos:
Merged #1340

PR #1340, opened 2019-02-19
fix rbd showmapped for ceph mimic

atta:


atta:
this ist the patch for: https://github.com/ganeti/ganeti/issues/1339

apoikos:
@atta thanks! Can you please update the commit message to make it a bit more descriptive. It's good to be able to see why things are done without having to open a GitHub issue in a browser :)

PR #1342, opened 2019-02-19
QEMU 3.1 compatibility

### LINKED ISSUE #1338: QEMU 3.1 compatibility
Debian Buster will be shipping with QEMU 3.1. Currently KVM instances fail to start because we're using `-balloon virtio`, which was removed in 3.1:

```
$ qemu-system-x86_64 -balloon virtio
qemu-system-x86_64: -balloon: invalid option
$ qemu-system-x86_64 --version
QEMU emulator version 3.1.0 (Debian 1:3.1+dfsg-2+b1)
Copyright (c) 2003-2018 Fabrice Bellard and the QEMU Project developers
```

We need to make sure Ganeti works with QEMU 3.1, if possible without using any deprecated features.

apoikos:
A number of options have been dropped in QEMU 3.1 and as a result KVM does not work out of the box. Fix these by converting to the new syntax. Note that for users using VNC TLS, this means a minimum QEMU version of 2.5 (released in Dec 2015).

This fixes #1338

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1289,7 +1289,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
         "%s,id=scsi" % hvp[constants.HV_KVM_SCSI_CONTROLLER_TYPE]
         ])
 
-    kvm_cmd.extend(["-balloon", "virtio"])
+    kvm_cmd.extend(["-device", "virtio-balloon"])
     kvm_cmd.extend(["-daemonize"])
     if not instance.hvparams[constants.HV_ACPI]:
       kvm_cmd.extend(["-no-acpi"])
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1547,7 +1547,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
         kvm_cmd.extend(["-nographic"])
 
     if hvp[constants.HV_USE_LOCALTIME]:
-      kvm_cmd.extend(["-localtime"])
+      kvm_cmd.extend(["-rtc", "base=localtime"])
 
     if hvp[constants.HV_KVM_USE_CHROOT]:
       kvm_cmd.extend(["-chroot", self._InstanceChrootDir(instance.name)])
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -621,7 +621,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
   # accept the output even on failure.
   _KVMOPTS_CMDS = {
     _KVMOPT_HELP: (["--help"], False),
-    _KVMOPT_MLIST: (["-M", "?"], False),
+    _KVMOPT_MLIST: (["-machine", "?"], False),
     _KVMOPT_DEVICELIST: (["-device", "?"], True),
   }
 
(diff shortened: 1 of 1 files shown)


apoikos:
I have successfully tested this PR with QEMU 3.1 and all possible option combinations. Also, live-migrations between QEMU 2.8 and QEMU 3.1 seem to work fine.

This PR also demonstrates the need for a better framework to manage the QEMU command line. String concatenation/parsing is obviously too error-prone.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -381,6 +381,61 @@ def update_hvinfo(dev, dev_type):
     # We have a (Disk, link, uri) tuple
     update_hvinfo(disk_entry[0], constants.HOTPLUG_TARGET_DISK)
 
+  # Handle KVM command line argument changes
+  try:
+    idx = kvm_cmd.index("-localtime")
+  except ValueError:
+    pass
+  else:
+    kvm_cmd = kvm_cmd[:idx] + ["-rtc", "base=localtime"] + kvm_cmd[idx+1:]
```
iustin: This block seems too complicated. Is there a specific need to keep argument ordering, or would the following work?

```
try:
  kvm_cmd.remove("-localtime")
  kvm_cmd += ["-rtc", "base=localtime"]
except ValueError:
  pass

File: lib/hypervisor/hv_kvm/__init__.py [hunk 2]
```diff
@@ -381,6 +381,61 @@ def update_hvinfo(dev, dev_type):
...
+    idx = kvm_cmd.index("-balloon")
+  except ValueError:
+    pass
+  else:
+    balloon_args = kvm_cmd[idx+1].split(",")[1:]
+    balloon_str = "virtio-balloon"
+    if balloon_args:
+      balloon_str += ",%s" % ",".join(balloon_args)
+
+    kvm_cmd = kvm_cmd[:idx] + ["-device", balloon_str] + kvm_cmd[idx+2:]
```
iustin: OK, here I understand why, but the tracking of indices… ouch.

iustin:
> This PR also demonstrates the need for a better framework to manage the QEMU command line. String concatenation/parsing is obviously too error-prone.

Yes, this is ugly, but for the last commit, the original data is a raw string, so no way around it, even if for backwards compatibility. 3.0+ could change to store in a JSON file, and fallback to parsing, but not earlier…

lib/hypervisor/hv_kvm/__init__.py
apoikos: Yes, that would have been much simpler actually, thanks!

lib/hypervisor/hv_kvm/__init__.py
apoikos: I think I can figure out something more readable, let me try...

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -381,6 +381,62 @@ def update_hvinfo(dev, dev_type):
...
+
+      vnc_args.append(arg)
+
+    if tls_obj is not None:
+      vnc_cmd = ["-vnc", ",".join(vnc_args)]
+      tls_obj_cmd = ["-object",
+                     "%s,%s" % (tls_obj, ",".join(tls_obj_args))]
+
+      # Replace the original vnc argument with the new ones
+      kvm_cmd[idx:idx+2] = tls_obj_cmd + vnc_cmd
```
apoikos: I *think* slice assignments work better. The rule here is list[idx:idx+n] will replace `n` elements of the list beginning with `idx` with the RHS.

iustin: Yes, this is better, but for the simpler cases, I'd remove the `else` branch and just do the asignments in the try block itself. Any way, looks good.

PR #1343, opened 2019-02-20
Fix ganeti-kvm-poweroff with systemd

### LINKED ISSUE #1288: ganeti-kvm-poweroff does not work in debian 9
Hi,
ganeti-kvm-poweroff script does not play with debian 9/systemd; after installation with

    /usr/sbin/update-rc.d ganeti-kvm-poweroff defaults

it is not fired on node shutdown. Changing https://github.com/ganeti/ganeti/blob/master/doc/examples/ganeti-kvm-poweroff.initd.in#L8 like this

    -# Default-Start:
    +# Default-Start: 2 3 4 5

works for us.

Regards,
Pawel

apoikos:
The ganeti-kvm-poweroff script is lacking Default-Start runlevels. While
this is okay with sysvinit, it doesn't work with systemd because the
latter keeps state of whether a service was started before stopping it.

Fix this by populating Default-Start on the LSB header. Note that start
is noop already, so this is harmless.

Thanks to @pboguslawski.

Fixes #1288.

### DIFF
--- a/doc/examples/ganeti-kvm-poweroff.initd.in
+++ b/doc/examples/ganeti-kvm-poweroff.initd.in
@@ -5,7 +5,7 @@
 # Provides:          ganeti-kvm-poweroff
 # Required-Start:
 # Required-Stop:     drbd qemu-kvm $local_fs
-# Default-Start:
+# Default-Start: 2 3 4 5
 # Default-Stop: 0 1 6
 # Short-Description: Poweroff Ganeti KVM instances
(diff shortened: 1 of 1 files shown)


apoikos:
> Hmm, there are some concerning things about this scripts itself, but the change LG.

I agree. To begin with, it should probably be using hypervisor code and not accessing stuff directly. Then again, at least someone finds it useful, so it doesn't harm to fix it.

Issue #1318, opened 2019-01-21
prop_jobFiltering has insufficient coverage


apoikos:
The `prop_jobFiltering` property produces insufficient coverage with "modern" (>= 2.8) versions of QuickCheck. At heart, I think the issue lies with the fact that we take a completely randomized jobqueue state and a completely random set of filters and we expect the filter predicates to match *something* on the jobqueue and also produce a uniform distribution of *actions*. The only reason there are matches at all, is because QuickCheck will always generate a match-all filter predicate as part of the basic coverage.

I think a better approach would be to start with the randomized jobqueue state and - given some probabilities - either generate a meaningful rule that matches at least one job - or generate a completely randomized rule.

I've played around a bit with a custom generator, but there's still work to be done. In the meantime I suggest disabling the test, as it doesn't offer much in its current form.

iustin:
After looking at this test for half an hour, I'm tempted to say it should be completely removed. It tries to test the whole of job filtering design *at once*, which makes the test heavy (and slow, and not as paralelisable as it could be), plus it's hard to understand and modify. Sigh.

PR #1344, opened 2019-02-20
Make the test suites runnable on Travis again

### LINKED ISSUE #1258: Travis-CI and Haskell version/dependencies are broken and fail to build tests
According to https://travis-ci.org/ganeti/ganeti in the latest build for 2.16, tests are broken on what I assume is an incorrect match of haskell/ghc versions and cabal invocations.

apoikos:
To make the tests runnable on Travis again, we basically need to address two issues:

 - The travis-provided environments are too dated to matter, so we need to use docker with a clean Debian Stretch environment instead.
 - Handle the lack of IPv6 support in Travis' docker executor, as this breaks some tests.

Additionally, I'm disabling the `prop_jobFiltering` Haskell test as per #1318. Note that the Python tests are still failing because we haven't addressed `cfgupgrade` from 2.16 to 3.0, but that's legitimate.

This resolves #1258 for the time being and is good for a start. Down the road we'll need a more versatile and faster CI system to also run the QA suite and perform matrix testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - docker
 
 install:
-  - docker build -t ganeti --file Dockerfile.stretch .
+  - docker build -t ganeti --build-arg uid=$(id -u) --file Dockerfile.stretch .
 
 script:
   - docker run -v $(pwd):/build ganeti sh -c "./autogen.sh && ./configure --enable-haskell-tests && make"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,88 +1,12 @@
-language: haskell
+dist: xenial
 
-sudo: false
-
-addons:
-  apt:
-    packages: &core_packages
-      - bridge-utils
-      - fping
-      - iproute
-      - iputils-arping
-      - lvm2
-      - m4
-      - make
-      - ndisc6
-      - openssl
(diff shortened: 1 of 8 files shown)


iustin:
Very good work, thanks a lot!

Question: does it even make sense to support socat < 1.7.3? Old-stable has 1.7.2, but also backprots for 1.7.3, so we could say - 1.7.3 is required. Maybe for 3.0?

Not sure I understand the run tests as non-root and how that works under Docker TBH…

apoikos:
> Question: does it even make sense to support socat < 1.7.3? Old-stable has 1.7.2, but also backprots for 1.7.3, so we could say - 1.7.3 is required. Maybe for 3.0?

I think we can drop it altogether. We should do a dependency cleanup sprint at some point :)
 
> Not sure I understand the run tests as non-root and how that works under Docker TBH…
You need to create a user inside the container. However, because I'm using a bindmount from the host to preserve artifacts across runs (otherwise build + tests would have to be a single execution), the uid must match between host and guest for the directory to be writable.

iustin:
@apoikos - what do you think about cherry-picking this for stable-2.16, so that we do have a valid CI build for it too?

PR #1345, opened 2019-02-20
Configure cfgupgrade for 3.x

apoikos:
Assume that we will be upgrading from 2.16 (which will probably be 99.9% of the cases). A separate PR will deal with upgrading from 2.17 and reverting the MaintD changes to the config.


apoikos:
… and the tests pass :100:

iustin:
> … and the tests pass 100

Nice! I totally agree upgrading from 2.16 will be the case, and that supporting it is the only reasonable way forward.

PR #1347, opened 2019-02-22
NEWS: updates for stable-2.16

apoikos:
Update information on merged changes and add contributor attributions
where available.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,9 @@ This is a bugfix and compatibility release.
 Important changes
 ~~~~~~~~~~~~~~~~~
 
+Updated X.509 certificate signing algorithm
++++++++++++++++++++++++++++++++++++++++++++
+
 Ganeti now uses the SHA-256 digest algorithm to sign all generated
 X.509 certificates used to secure the RPC communications between nodes.
 Previously, Ganeti was using SHA-1 which is seen as weak (but not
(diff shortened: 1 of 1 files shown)


iustin:
Looks good, thanks. Question: do we you want to organise the cherry-picking? I guess you have a list and go through it, I started to do the same but I wonder if sharing the list (google sheets?) would make more sense. It's a lot of commits that were "lost" with the master reset…

apoikos:
@iustin: it wouldn't be a bad idea actually. We could also use a wiki page here I guess.

iustin:
On 2019-02-22 05:18:11, Apollon Oikonomopoulos wrote:
> @iustin: it wouldn't be a bad idea actually. We could also use a wiki page here I guess.

Let's see how it works. It's only 409 commits, so a wiki should be
doable but at the limit.

See https://github.com/ganeti/ganeti/wiki/AtticMasterCherrypicks.

apoikos:
@iustin: i'd add a `--no-merges` to that `git log` command, which would probably save 40-50 commits.

iustin:
On 2019-02-22 17:12:08, Apollon Oikonomopoulos wrote:
> @iustin: i'd add a `--no-merges` to that `git log` command, which
> would probably save 40-50 commits.

Good idea. 41 commits, to be precise.

I know you did some cherry-picks, can you please update the one you did?
I don't expect we'll regenerate the page anymore now.

PR #1348, opened 2019-02-22
A few more cherry-picks for stable 2.16

iustin:
I think all three can go into stable-2.16, even the function rename, since it's more likely we'll do more stable-2.16→master merges than stable-2.15→2.16.


iustin:
All cherry-picks, updating CLA status.

PR #1349, opened 2019-02-22
Two more cherry-picks for stable 2.16

iustin:
Starting to run out of "obvious" things that can be ported over…

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1347,6 +1347,10 @@ class KVMHypervisor(hv_base.BaseHypervisor):
 
     kvm_cmd.extend(["-device", "virtio-balloon"])
     kvm_cmd.extend(["-daemonize"])
+    # logfile for qemu
+    qemu_logfile = utils.PathJoin(pathutils.LOG_KVM_DIR,
+                                  "%s.log" % instance.name)
+    kvm_cmd.extend(["-D", qemu_logfile])
     if not instance.hvparams[constants.HV_ACPI]:
       kvm_cmd.extend(["-no-acpi"])
     if instance.hvparams[constants.HV_REBOOT_BEHAVIOR] == \
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1987,7 +1987,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     # during instance startup anyway, and to avoid problems when soft
     # rebooting the instance.
     cpu_pinning = False
-    if up_hvp.get(constants.HV_CPU_MASK, None):
+    if up_hvp.get(constants.HV_CPU_MASK, None) \
+        and up_hvp[constants.HV_CPU_MASK] != constants.CPU_PINNING_ALL:
       cpu_pinning = True
 
     if security_model == constants.HT_SM_POOL:
--- a/lib/pathutils.py
+++ b/lib/pathutils.py
@@ -164,6 +164,8 @@ LOG_OS_DIR = LOG_DIR + "/os"
 LOG_ES_DIR = LOG_DIR + "/extstorage"
 #: Directory for storing Xen config files after failed instance starts
 LOG_XEN_DIR = LOG_DIR + "/xen"
+# Directory to store the output of kvm instances
+LOG_KVM_DIR = LOG_DIR + "/kvm"
 
 # Job queue paths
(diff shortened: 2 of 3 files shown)


iustin:
Just cherry-picks, setting CLA.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1987,7 +1987,8 @@ def _ExecuteKVMRuntime(self, instance, kvm_runtime, kvmhelp, incoming=None):
     # during instance startup anyway, and to avoid problems when soft
     # rebooting the instance.
     cpu_pinning = False
-    if up_hvp.get(constants.HV_CPU_MASK, None):
+    if up_hvp.get(constants.HV_CPU_MASK, None) \
+        and up_hvp[constants.HV_CPU_MASK] != constants.CPU_PINNING_ALL:
```
apoikos: To be completely honest, I don't understand why this is necessary and neither the commit message nor the original PR (#1261) help me much. Why is just unsetting `cpu_mask` not enough? @calumcalder, could you please help me out a bit here?

calumcalder: Hi @apoikos, as far as I remember we were having issues with dependencies on psutil, which is used by ganeti to set the CPU affinity. This issue was cropping up on configs which had the CPU mask set to CPU_PINNING_ALL, and I narrowed down the issue to CPU pinning code paths still being hit.

I don't remember exactly why unsetting cpu_mask wasn't possible or didn't solve the issue, but I do remember doing some investigation around that (that'll teach me to write better commit messages!)

I can see an argument for having both a PINNING_DISABLED and PINNING_ALL option to be more explicit, and re-enabling the PINNING_ALL option for explicitly setting the CPU affinity to 'any CPU', but the only use case I can see is if for some reason your hypervisor spawns instances with affinities set, and you want to override that, which is not something I've come across.

iustin:
On 2019-02-22 14:03:37, Apollon Oikonomopoulos wrote:
> apoikos commented on this pull request.
> 
> 
> 
> > @@ -213,6 +213,7 @@ def GetPaths():
>      (mond_log, FILE, 0600, getent.mond_uid, getent.masterd_gid, False),
>      (pathutils.LOG_OS_DIR, DIR, 0750, getent.noded_uid, getent.daemons_gid),
>      (pathutils.LOG_XEN_DIR, DIR, 0750, getent.noded_uid, getent.daemons_gid),
> +    (pathutils.LOG_KVM_DIR, DIR, 0750, getent.noded_uid, getent.daemons_gid),
> 
> It's not an issue with this specific patch, but really these
> directories should probably be group `adm` instead of gnt-daemons.

Good point, but that probably should go into new release, not point one.

> > @@ -1347,6 +1347,10 @@ def _GenerateKVMRuntime(self, instance, block_devices, startup_paused,
>  
>      kvm_cmd.extend(["-device", "virtio-balloon"])
>      kvm_cmd.extend(["-daemonize"])
> +    # logfile for qemu
> +    qemu_logfile = utils.PathJoin(pathutils.LOG_KVM_DIR,
> +                                  "%s.log" % instance.name)
> +    kvm_cmd.extend(["-D", qemu_logfile])
> 
> This is neither a bugfix, nor a compatibility fix, but it's useful
> nevertheless so let's pretend it should have landed in stable-2.16 in
> the first place.

Exactly my thought :)

> 
> > @@ -1987,7 +1987,8 @@ def _ExecuteKVMRuntime(self, instance, kvm_runtime, kvmhelp, incoming=None):
>      # during instance startup anyway, and to avoid problems when soft
>      # rebooting the instance.
>      cpu_pinning = False
> -    if up_hvp.get(constants.HV_CPU_MASK, None):
> +    if up_hvp.get(constants.HV_CPU_MASK, None) \
> +        and up_hvp[constants.HV_CPU_MASK] != constants.CPU_PINNING_ALL:
> 
> To be completely honest, I don't understand why this is necessary and
> neither the commit message nor the original PR (#1261) help me much.
> Why is just unsetting `cpu_mask` not enough? @calumcalder, could you
> please help me out a bit here?

The code already does the dance of using cpu_pinning boolean variable.
Changing to unset the HV_CPU_MASK would require changing this code path,
versus just adding one more conditional.

So to me this is the simplest fix for the problem, which looks good.

lib/hypervisor/hv_kvm/__init__.py
iustin: Replying here as well as email reply doesn't put the comments in the right place:

The code already does the dance of using cpu_pinning boolean variable. Changing to unset the HV_CPU_MASK would require changing multiple code paths, versus just adding one more conditional. So to me this is the simplest fix for the problem, which looks good.

PR #1355, opened 2019-02-28
TagSet changes for better test coverage

iustin:
I'm sending this to stable-2.16 since it's a regression in tests, and would be good to get back good coverage in preparation for 2.16.1 (and actually non-failing tests). The changes are really trivial and non-invasive, so I think this has low risk of breaking things.

The only (Haskell) test failing is jobFiltering, in getting proper coverage, which was failing before as well.

### DIFF
--- a/src/Ganeti/WConfd/Ssconf.hs
+++ b/src/Ganeti/WConfd/Ssconf.hs
@@ -88,7 +88,7 @@ mkSSConfHvparams cluster = map (id &&& hvparams) [minBound..maxBound]
 mkSSConf :: ConfigData -> SSConf
 mkSSConf cdata = SSConf . M.fromList $
     [ (SSClusterName, return $ clusterClusterName cluster)
-    , (SSClusterTags, toList $ tagsOf cluster)
+    , (SSClusterTags, toList . unTagSet $ tagsOf cluster)
     , (SSFileStorageDir, return $ clusterFileStorageDir cluster)
     , (SSSharedFileStorageDir, return $ clusterSharedFileStorageDir cluster)
     , (SSGlusterStorageDir, return $ clusterGlusterStorageDir cluster)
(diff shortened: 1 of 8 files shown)


src/Ganeti/WConfd/Ssconf.hs
```diff
@@ -88,7 +88,7 @@ mkSSConfHvparams cluster = map (id &&& hvparams) [minBound..maxBound]
 mkSSConf :: ConfigData -> SSConf
 mkSSConf cdata = SSConf . M.fromList $
     [ (SSClusterName, return $ clusterClusterName cluster)
-    , (SSClusterTags, toList $ tagsOf cluster)
+    , (SSClusterTags, toList . unTagSet $ tagsOf cluster)
```
apoikos: This is more of an exercise for my Haskell, but I guess we could define toList for TagSet:

```haskell
toList :: TagSet -> [String]
toList = Set.toList . unTagSet
```
This will keep the Set API and make the TagSet type opaque by not exposing the `unTagSet` accessor.

iustin: Yes, but two things:

1. toList defined here for this small type, as a separate function, with one use, seems overboard. It would be better to define something like a Foldable instance, and use the toList from that class. However, even that's over doing it, because in reality:

2. The real issue is that this patch is not the real fix. We don't need to hide the structure of the TagSet, because the structure itself is not the problem; instead, we should define a proper type for a Tag itself, and make that opaque, which would allow validation all the time on what is a proper tag, and would allow TagSet to remain just a plain Set, and the Arbitrary (default) Set instance would be enough as long as Tag has a good Arbitrary instance itself.

But for the proper fix, the impact could be much wider. I don't know if (somehow) invalid tags have sneaked in cluster configurations (in productions), which means we shouldn't actually reject invalid tags at load time, etc.

So I thought to apply the smaller fix, even though it's not the _proper_ fix. What do you think?

apoikos:
@iustin, I'm fine with it! As I said I was mostly thinking out loud trying to understand good Haskell practices. Thanks for the insight, LGTM!

PR #1357, opened 2019-03-01
Relax python-psutil version check

gedia:
CPU pinning on KVM would fail with psutil versions greater than 2.x.x due to there being a check in ganeti explicitly rejecting any version of psutil other than 2.x.x.
Pinning functionality has been tested and is working with python-psutil versions 5.0.1 (shipping with debian stretch) and 5.5.1 (debian buster soft freeze), so it should be safe to relax this restriction to allow versions 2 through 5 inclusive.

However, further discussion in the pull request resulted in dropping entirely the upper bound, as @apoikos mentioned that "psutil does not follow semantic versioning rules".

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -47,10 +47,7 @@ try:
   import psutil   # pylint: disable=F0401
   if psutil.version_info < (2, 0, 0):
     # The psutil version seems too old, we ignore it
-    psutil_err = "too old (2.x.x needed, %s found)" % psutil.__version__
-    psutil = None
-  elif psutil.version_info >= (3,):
-    psutil_err = "too new (2.x.x needed, %s found)" % psutil.__version__
+    psutil_err = "too old (2.x.x or newer needed, %s found)" % psutil.__version__
     psutil = None
   else:
     psutil_err = "<no error>"


lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -47,10 +47,10 @@
   import psutil   # pylint: disable=F0401
   if psutil.version_info < (2, 0, 0):
     # The psutil version seems too old, we ignore it
-    psutil_err = "too old (2.x.x needed, %s found)" % psutil.__version__
+    psutil_err = "too old (2.x.x or newer needed, %s found)" % psutil.__version__
     psutil = None
-  elif psutil.version_info >= (3,):
-    psutil_err = "too new (2.x.x needed, %s found)" % psutil.__version__
+  elif psutil.version_info >= (6,):
```
apoikos: AFAICT, psutil does not follow semantic versioning rules (see [changelog](https://github.com/giampaolo/psutil/blob/master/HISTORY.rst)), so I think it's more futureproof to remove the upper version bound, than not do so :)

iustin: I wonder if we can relax/change this so that we check if psutil understands what we want to do, rather than base check on version?

gedia: @iustin Sorry, I force-pushed an update before reading your comment here.

I have no idea how this could be implemented, but I could look into it at some point if you think it makes sense. I don't think I'll be able to before 2.16.1 goes GA though...

iustin: @gedia - oh sure, I didn't mean for this patch, but as general concept. The way it is right now looks good.

PR #1361, opened 2019-03-03
utils.process: use soft NOFILE limit in CloseFDs

apoikos:
CloseFDs is used by utils.RunCmd and friends to make sure we don't leak
any file descriptors to child processes. The way CloseFDs operates, is
by attempting to close all FDs up to the RLIMIT_NOFILE *hard* limit,
or up to SC_OPEN_MAX if the former is set to infinity.

SC_OPEN_MAX is internally set by glibc to the *soft* RLIMIT_NOFILE
limit, usually around 1024. The *hard* RLIMIT_NOFILE limit has
traditionally been set around 65k - still manageable to iterate over.
However, recent kernel and systemd versions bumped tha hard limit to 1M,
which takes a considerable amount of time to handle:

  timeit.timeit("RunCmd('/bin/true', noclose_fds=[1])",
                setup="from ganeti.utils import RunCmd",
                number=10)
  18.81471085548401

Note that CloseFDs is only used when the noclose_fds is set, otherwise
we rely on subprocess.Popen()'s built-in logic which merely closes FDs 3
through SC_OPEN_MAX.

Now, should we care about the *hard* limit at all? The answer is
probably no. From getrlimit(2):

  Each resource has an associated soft and  hard  limit,  as defined by
  the rlimit structure:

      struct rlimit {
          rlim_t rlim_cur;  /* Soft limit */
          rlim_t rlim_max;  /* Hard limit (ceiling for rlim_cur) */
      };

  The  soft limit is the value that the kernel enforces for the
  corresponding resource. The hard limit acts as a ceiling for the soft
  limit: an unprivileged process may set only its soft limit to a value
  in the range from 0 up to the hard limit, and (irreversibly) lower its
  hard limit. A privileged process (under Linux: one with the
  CAP_SYS_RESOURCE capability in the initial user namespace) may
  make arbitrary changes to either limit value.

So what we should care about is the *soft* limit instead, which we
already do through SC_OPEN_MAX. Removing the getrlimit logic completely
yields a massive speedup:

  timeit.timeit("RunCmd('/bin/true', noclose_fds=[1])",
                setup="from ganeti.utils import RunCmd",
                number=10)
  0.0466561317444

This reduces the runtime of the ganeti.utils.process tests
significantly.

### DIFF
--- a/lib/utils/process.py
+++ b/lib/utils/process.py
@@ -1093,12 +1093,8 @@ def CloseFDs(noclose_fds=None):
   else:
     MAXFD = 1024
 
-  maxfd = resource.getrlimit(resource.RLIMIT_NOFILE)[1]
-  if maxfd == resource.RLIM_INFINITY:
-    maxfd = MAXFD
-
   # Iterate through and close all file descriptors (except the standard ones)
-  for fd in range(3, maxfd):
+  for fd in range(3, MAXFD):
     if noclose_fds and fd in noclose_fds:
       continue
     utils_wrapper.CloseFdNoError(fd)


iustin:
So, this is per se a very good find, thanks a lot! We should get this in indeed.

But question: looking at how the glibc code implements SC_OPEN_MAX, it seems that especially when this is set to infinity (-1), we (before and after this patch), just decide 1024 as random (?) number.

I could have sworn I've see discussion a few years ago on exporting "max-open-fd" or "close-to-max-open-fd", but I can't find it, so maybe I remember wrongly or confuse things.

So, LGTM, but this still seems somewhat fragile (but it's what Popen does as well, so meh).

apoikos:
> But question: looking at how the glibc code implements SC_OPEN_MAX, it seems that especially when this is set to infinity (-1), we (before and after this patch), just decide 1024 as random (?) number.

It's more of a historical value than random. But really, if the *hard* limit is set to infinity, you can't expect to have to close 4 billion file descriptors or so :)

> So, LGTM, but this still seems somewhat fragile (but it's what Popen does as well, so meh).

I think the tradeoff is not bad. `noclose_fds` is only used in a couple of places and this PR just makes it consistent with the more regular codepath that goes through plain `Popen`. On the flip side, it speeds up test execution greatly.

PR #1365, opened 2019-04-02
Merge stable-2.16 to master

apoikos:
This is just an upmerge of stable-2.16 to master. Only ca3dbb5 is relevant.


iustin:
Note: added `cla:yes` due to it being a merge.

PR #1367, opened 2019-04-05
Some forward-compatibility with Python 3

apoikos:
This changeset fixes all low-hanging fruits regarding Python 3 compatibility, by adapting to the syntax changes in a backwards-compatible way. With these changes we still need to run on Python 2, but the code is syntactically correct for Python 3 as well and the Python 3 interpreter will happily *parse* (but not execute) the code. This of course does not mean that we're ready for Python 3, there's still a lot to do.

This is suitable for merging into master; upcoming changes will break Python 2 compatibility, as we will need to adapt to Python 3 APIs, so subsequent work will be done on a topic branch.

**Note**: for the sake of sanity, you are advised to review the changes commit-by-commit.


iustin:
All good, a few questions:

- in visit_BinOp, should it be either left _or_ right float, not _and_?
- should the from __future__ import print be committed _before_ the switch to print as function?

And sorry for the delay.

apoikos:
@iustin: under Python 2, the result of a division is always a float if at least one of the two operands is a float, so we only care about cases where none of the operands is a float:

```python
Python 2.7.16rc1 (default, Feb 18 2019, 11:05:09) 
[GCC 8.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 1 / 2.
0.5
>>> 1. / 2
0.5
>>> 1 / 2
0
```

As for the import statement, once you import the print function, then you **must** use it and print cannot be a statement anymore. So, the only way to do the change atomically would be to squash both commits into one.

iustin:
> @iustin: under Python 2, the result of a division is always a float if at least one of the two operands is a float, so we only care about cases where none of the operands is a float:
> 
> ```python
> Python 2.7.16rc1 (default, Feb 18 2019, 11:05:09) 
> [GCC 8.2.0] on linux2
> Type "help", "copyright", "credits" or "license" for more information.
> >>> 1 / 2.
> 0.5
> >>> 1. / 2
> 0.5
> >>> 1 / 2
> 0
> ```

Yes, exactly. Hence why… oh. I read the "and" wrongly, it checks if the _current_ operand is both a Num and a Float.

For some reason, I read is as: `if left is Float and right is Float: break`, which is very different than what the code actually does. Sorry for the noise :)T

> As for the import statement, once you import the print function, then you **must** use it and print cannot be a statement anymore. So, the only way to do the change atomically would be to squash both commits into one.

Ah, got it. Never mind then, good as is.

iustin:
Ping on merging?

PR #1391, opened 2019-06-19
Support build in separate directory from RO source dir.

paurkedal:
AM_MAINTAINER_MODE is added to allow disabling changes to the source
directory.  This allows, e.g. mounting it RO in a Docker container.


paurkedal:
I signed it!

paurkedal:
I signed it!

configure.ac
```diff
@@ -17,6 +17,7 @@ AM_INIT_AUTOMAKE([1.9 foreign tar-ustar -Wall -Wno-portability]
                              *1.11*);;
                              *) echo serial-tests;;
                              esac]))
+AM_MAINTAINER_MODE([enable])
```
apoikos: Note to self: [as it turns out](https://blogs.gnome.org/desrt/2011/09/08/am_maintainer_mode-is-not-cool/), maintainer mode is enabled by default, and `AM_MAINTAINER_MODE([enable])` actually allows us to **disable** it using `--disable-maintainer-mode`. So this does not change the default semantics of ./configure.

paurkedal: Yes, that was the intension. I agree with the linked document (as long as the generated files are not committed to Git), but I needed a way to avoid accidental reconfiguration in Docker RO mount.

PR #1394, opened 2019-06-22
remove call to brctl and advance to iproute2

saschalucas:
This is the only place, where brctl is called. To get rid of the dependency
of bridge-utils substitute it by a proper iproute2 call.


atta:
this will be also covered in https://github.com/ganeti/ganeti/pull/1395

apoikos:
Merged, thanks!

PR #1402, opened 2019-11-21
Build using GHC 8.6/network 2.8

apoikos:
Add MonadFail instances for a couple of types to allow building against GHC 8.6, see [here](https://gitlab.haskell.org/ghc/ghc/wikis/migration/8.6#monadfaildesugaring-by-default) and [here](https://prime.haskell.org/wiki/Libraries/Proposals/MonadFail#Adaptingoldcode).

Also relax the dependency on network, since Ganeti builds fine with network 2.8.


apoikos:
Note that ganeti can still build with GHC 8.0 and later, as indicated by https://buildbot.ganeti.dev/#/builders/2/builds/9

PR #1405, opened 2019-11-29
Port Ganeti.Hash to cryptonite

apoikos:
Crypto has not been updated since 2012, and it currently seems to bebroken with GHC 8.6; `hmac_sha1 [] []` gets caught in an endless loop accumulating over an array. Rather than try to fix it, I think it's a good idea to port the HMAC implementation to cryptonite, which is thede-facto standard for crypto in Haskell these days.

Re-implement computeMac using cryptonite, while keeping the high-levelAPI intact.


apoikos:
@iliastsi kindly pointed out that this is actually due to Crypto [assuming accumArray to be lazy](https://gitlab.haskell.org/ghc/ghc/issues/16406). Consensus is that it's rather unlikely to be fixed in Crypto.

dannyman:
Apollon, thank you for pushing the ball forward on ganeti maintenance.

apoikos:
@dannyman you're welcome :)

PR #1406, opened 2019-11-29
QuickCheck 2.12 compatibility

apoikos:
QuickCheck 2.12 has some breaking changes, in particular it swapped the
order of the two first arguments of `cover' and changed the percentage
from Int to Double.

For the first change, adapt the code to the new order and provide a
compatibility function for older QuickCheck versions. When time comes to
drop support for pre-2.12 QuickCheck versions, we can simply drop the
compatibility function and unhide the original.

For the latter change, just add a fromIntegral conversion which
shouldn't harm anyway.

### DIFF
--- a/cabal/ganeti.template.cabal
+++ b/cabal/ganeti.template.cabal
@@ -85,7 +85,7 @@ library
   if flag(htest)
     build-depends:
         HUnit                         >= 1.2.4.2    && < 1.7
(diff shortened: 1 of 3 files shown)


apoikos:
@iustin thanks! However, it looks like the type change from Int to Double breaks older versions, so this needs a bit more work (ideas welcome).

PR #1409, opened 2019-12-04
Specify the entire sshd-restart-command at build-time

rbott:
An older commit moved the RPL_SSH_INITD_SCRIPT command from direct execution into a variable named sshd_init_script (4d482b06778da6ae6af734cb6583be041a7f669f). If the command contains spaces (e.g. for parameters), the variable will only hold everything up to the first space.

I have added quotes to fix this behaviour.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,14 +136,14 @@ AC_ARG_WITH([haskell-flags],
   [hextra_configure=""])
 AC_SUBST(HEXTRA_CONFIGURE, $hextra_configure)
 
-# --with-ssh-initscript=...
-AC_ARG_WITH([ssh-initscript],
-  [AS_HELP_STRING([--with-ssh-initscript=SCRIPT],
-    [SSH init script to use (default is /etc/init.d/ssh)]
+# --with-sshd-restart-command=...
+AC_ARG_WITH([sshd-restart-command],
+  [AS_HELP_STRING([--with-sshd-restart-command=SCRIPT],
+    [SSH restart command to use (default is /usr/sbin/service ssh restart)]
   )],
-  [ssh_initd_script="$withval"],
-  [ssh_initd_script="/etc/init.d/ssh"])
-AC_SUBST(SSH_INITD_SCRIPT, $ssh_initd_script)
+  [sshd_restart_command="$withval"],
+  [sshd_restart_command="/usr/sbin/service ssh restart"])
+AC_SUBST(SSHD_RESTART_COMMAND, $sshd_restart_command)
 
 # --with-export-dir=...
 AC_ARG_WITH([export-dir],
--- a/Makefile.am
+++ b/Makefile.am
@@ -2492,7 +2492,7 @@ $(REPLACE_VARS_SED): $(SHELL_ENV_INIT) Makefile stamp-directories
 	  echo 's#@''CUSTOM_XEN_INITRD@#$(XEN_INITRD)#g'; \
 	  echo 's#@''CUSTOM_IALLOCATOR_SEARCH_PATH@#$(IALLOCATOR_SEARCH_PATH)#g'; \
 	  echo 's#@''CUSTOM_EXPORT_DIR@#$(EXPORT_DIR)#g'; \
-	  echo 's#@''RPL_SSH_INITD_SCRIPT@#$(SSH_INITD_SCRIPT)#g'; \
+	  echo 's#@''RPL_SSHD_RESTART_COMMAND@#$(SSHD_RESTART_COMMAND)#g'; \
 	  echo 's#@''PKGLIBDIR@#$(libdir)/ganeti#g'; \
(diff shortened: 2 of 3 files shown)


rbott:
I probably should have added some more information, that would have made it clearer. I found the bug while building Debian packages from the current master branch. Currently, `debian/rules` configures Ganeti with this flag:
```
--with-ssh-initscript="/usr/sbin/invoke-rc.d ssh"
```

This ends up in the `daemon-utils` script just fine but in the end it gets executed as  `/usr/sbin/invoke-rc.d restart`.

About systemd: it depends on the system you are working on. Debian does not yet use systemd to manage ganeti (it still ships with an init script). Hence there are no unit files for the ganeti daemons and no `ganeti.target`, which the function `use_systemctl()` (https://github.com/ganeti/ganeti/blob/master/daemons/daemon-util.in#L172) checks for. Looking at the function and its other usages it seems like its purpose was to find out if systemd is present _and_ manages ganeti and has been reused in `reload_ssh_keys()` in a slightly wrong context. 

I will extend this PR so `reload_ssh_keys()` will properly detect systemd (even it does not yet manage ganeti natively).

rbott:
Thinking about this a bit more, I think this needs a slightly bigger rework and should probably be solved outside of this PR.

```
  local ssh_initd_script="@RPL_SSH_INITD_SCRIPT@"

  if use_systemctl; then
    systemctl restart ${ssh_initd_script##*/}.service
  else
    $ssh_initd_script restart
  fi
```
Currently the code tries to derive just the _name_ of the SSH server service from a custom command provided at compile time which might work or not.

We should probably ditch the parameter `--with-ssh-initscript`/`@RPL_SSH_INITD_SCRIPT@` altogether and ask the user to provide the name of the ssh service instead during `configure` time.
The `daemon-util` script should contain its own logic to figure out which init system is present and how to interact with it.
But it is actually harder to figure out whether the SSH service is known to the init system as 'ssh', 'sshd', 'openssh-server' etc. - so that should default to 'ssh' and could be optionally overwritten.

Any opinions on that approach?

iustin:
Yes, your'e right that `use_systemctl` was mistakenly borrowed.

Way back, the proper thing to pass to `--with-ssh-initscript` was `/etc/init.d/ssh` - per the name even. And that would work fine with both systemd and sysvinit, even today, with the current code. Probably Debian switch to invoke-rc.d before `use_systemctl` was introduced here.

So I'm not sure which way to fix:

- just adding quotes still leaves systemctl use here broken
- Debian switching back to script would still work (I think)

To your proposal: just passing the name of the unit is not enough, since the daemon-util would still need to decide which system to use to restart. I think this all should get replaced with a `--with-sshd-restart-command`, set at configure time by the distribution. Most likely this will be `systemctl restart` nowadays anyway, but better to get Ganeti out of the guessing game?

What do you think?

ge-fa:
Just a quick note: Isn't `service restart ...` init system agnostic, and if so, would be the right choice?

rbott:
@ge-fa I am not very familiar with non-debian-based distributions so I am not sure if that statement is true. But I tend to agree :-) Maybe this should be a sane default here so it would already cover most distributions out there.

@iustin I will rework this PR later today to have a more general solution as you suggested. That seems like the best way to solve it.

iustin:
@ge-fa: Probably yes, but if we are reworking this anyway, let's give the option to use whatever? Hard-coding "service ... restart" and only allowing the service name override seems not a real improvement.

ge-fa:
Yeah, I agree, makes sense.

rbott:
I have reworked this PR to reflect our discussion. `configure` now allows to specify the entire command for restarting the SSH daemon, but defaults to `/usr/sbin/service ssh restart`. This should provide a sane default for most distributions but leaves the option to customize the command as well as the name of the ssh-service itself.

configure.ac
```diff
@@ -136,14 +136,14 @@ AC_ARG_WITH([haskell-flags],
...
+AC_ARG_WITH([sshd-restart-command],
+  [AS_HELP_STRING([--with-sshd-restart-command=SCRIPT],
+    [SSH restart command to use (default is /usr/sbin/service ssh restart)]
   )],
-  [ssh_initd_script="$withval"],
-  [ssh_initd_script="/etc/init.d/ssh"])
-AC_SUBST(SSH_INITD_SCRIPT, $ssh_initd_script)
+  [sshd_restart_command="$withval"],
+  [sshd_restart_command="/usr/sbin/service ssh restart"])
+AC_SUBST(SSHD_RESTART_COMMAND, $sshd_restart_command)
```
iustin: Does this need to be quoted, or does it work with whitespace? Don't remember anymore how m4 macros work.

rbott: I am not very familiar with m4 macros. But the rest of the file looks exactly the same and it worked in my builds/tests :-)

iustin: That's good enough for me :)

apoikos:
Merged, thanks!

PR #1413, opened 2019-12-11
LuxiD: wait for a job's process to exit before archiving it

### LINKED ISSUE #1266: ganeti 2.15.2: master_candidate job queue management
during a recent masterfailover our new cluster master kept losing its `ganeti-luxid` process, dying with errors like "openFile: resource exhausted (Too many open files)"

while investigating, it was noticed that there were a lot of old "job-[[:digit:]]+" files, stretching back quite some time, in the new master's `/var/lib/ganeti/queue`.  after some time and a number of restarts, luxid eventually stabilized.  our guess is that `ganeti-cleaner`, once this node was a master and the cleaner could run, eventually knocked down the queue backlog to the point that luxid was able to startup and process the remainder.

looking around our clusters, i've found that the master_candidates do indeed have a tendency to accumulate huge numbers of jobs in `/var/lib/ganeti/queue`.  10s of thousands in some cases, in others even throwing "ls: Argument list too long" errors.

it was suggested that this problem is related to #1149.

possible workarounds are to regularly failover the master to the candidates to keep the job
directories clean, or run some scripts with cron job that regularly clean the queue.

apoikos:
There's a race condition in ArchiveJob that may result in stale jobfiles remaining on master candidates. The race is due to the fact that ArchiveJob checks whether a job has finished relying solely on the job status, while the actual job process might be still alive and replicating the finalized job file to the master candidates - which is the last thing it does before exiting.

LuxiD is usually faster than Python and so the archival event will be replicated to some master candidates before the original jobfile reaches them through the Python replication loop. As a result, master candidates end up with an earlier version of the job file in the archive and the finalized job file in their regular queue.

Note that master candidates do not perform their own archival and only rely on archival events replicated from the master. Because the affected jobs have already been archived on the master, there will be no future attempts to archive them again and they will remain on each affected master candidate until the mc gets promoted to master.

This bug affects solely the OP_GROUP_VERIFY_DISKS jobs submitted by the watcher, which the watcher archives after finishing. The problem is especially prominent on large clusters, where job replication may take a non-trivial amount of time, making the race more probable. The result is the accumulation of hundreds of jobs per day, which LuxiD will eventually try to load when promoted to master, making its startup time and resource usage skyrocket to the point where the master failover operation may fail.

Close this race by making sure the job's process is indeed dead before performing any archival operation. We wait up to 30 seconds for the job to exit after having finalized its job status, otherwise abort the whole operation and let the job be bulk-archived by the watcher in the future. This timeout seems a bit high, but it's there to allow for the delays caused by some master candidates being offline.

While at it, we move the status check before the process check, to avoid waiting for processes of jobs not in a finalized stage. Moving the check outside the queue lock is in order, since we are not manipulating the queue at this point.

This fixes #1266.

### DIFF
--- a/src/Ganeti/JQueue.hs
+++ b/src/Ganeti/JQueue.hs
@@ -73,6 +73,7 @@ module Ganeti.JQueue
     , cancelJob
     , tellJobPriority
     , notifyJob
+    , waitUntilJobExited
     , queueDirPermissions
     , archiveJobs
     -- re-export
@@ -88,6 +89,7 @@ import Control.Concurrent (forkIO, threadDelay)
 import Control.Exception
 import Control.Lens (over)
 import Control.Monad
+import Control.Monad.Fix
 import Control.Monad.IO.Class
 import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Maybe
@@ -561,6 +563,34 @@ isQueuedJobDead ownlivelock =
   maybe (return False) (liftIO . isDead)
   . mfilter (/= ownlivelock)
   . qjLivelock
+
+-- | Waits for a job's process to exit
+waitUntilJobExited :: Livelock -- ^ LuxiD's own livelock
+                   -> QueuedJob -- ^ the job to wait for
+                   -> Int -- ^ timeout in milliseconds
+                   -> ResultG (Bool, String)
+waitUntilJobExited ownlivelock job tmout = do
+  let sleepDelay = 100 :: Int -- ms
+      jName = ("Job " ++) . show . fromJobId . qjId $ job
+  logDebug $ "Waiting for " ++ jName ++ " to exit"
+
+  start <- liftIO getClockTime
+  let tmoutPicosec :: Integer
+      tmoutPicosec = fromIntegral $ tmout * 1000 * 1000 * 1000
+      wait = noTimeDiff { tdPicosec = tmoutPicosec }
+      deadline = addToClockTime wait start
+
+  liftIO . fix $ \loop -> do
+    -- fail if the job is in the startup phase or has no livelock
+    dead <- maybe (fail $ jName ++ " has not yet started up")
+      (liftIO . isDead) . mfilter (/= ownlivelock) . qjLivelock $ job
+    curtime <- getClockTime
+    let elapsed = timeDiffToString $ System.Time.diffClockTimes
+                  curtime start
+    case dead of
+      True -> return (True, jName ++ " process exited after " ++ elapsed)
+      _ | curtime < deadline -> threadDelay (sleepDelay * 1000) >> loop
+      _ -> fail $ jName ++ " still running after " ++ elapsed
 
 -- | Waits for a job ordered to cancel to react, and returns whether it was
 -- canceled, and a user-intended description of the reason.
--- a/src/Ganeti/Query/Server.hs
+++ b/src/Ganeti/Query/Server.hs
@@ -507,7 +507,7 @@ handleCall _ qstat  cfg (CancelJob jid kill) = do
       return result
     Bad s -> return . Ok . showJSON $ (False, s)
 
-handleCall qlock _ cfg (ArchiveJob jid) =
+handleCall qlock qstat cfg (ArchiveJob jid) =
   -- By adding a layer of MaybeT, we can prematurely end a computation
   -- using 'mzero' or other 'MonadPlus' primitive and return 'Ok False'.
   runResultT . liftM (showJSON . fromMaybe False) . runMaybeT $ do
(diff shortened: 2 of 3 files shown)


apoikos:
I haven't field-tested this yet, but I'm opening a PR because I would like some validation of the general concept. Also, I'm slowly beginning to find my way through Ganeti's Monad stack, so I'd appreciate a thorough review there :)

apoikos:
Used  2 * (RPC connect timeout) + 1 instead of a random 30-second timeout.

apoikos:
This updated version works as expected, raising a GenericError when we hit the timeout:

```
2019-12-12 13:38:26,267722822000 EET: ganeti-luxid pid=188356/ThreadId 127 WARNING Failed to execute request ArchiveJob (JobId {fromJobId = 3097}): GenericError "Job 3097 still running after 11000 ms"
2019-12-12 13:38:26,272116350000 EET: ganeti-luxid pid=188356/ThreadId 127 DEBUG sendMsg: 0ms
2019-12-12 13:38:26,277527757000 EET: ganeti-luxid pid=188356/ThreadId 127 DEBUG Received message (truncated): RecvConnClosed
```
Note that this bubbles up to the watcher as well, so we need to handle it there.

iustin:
You mentioned a thorough review - then it'll have to wait a few days, sorry.

apoikos:
@iustin, no problem, take your time. I may try to improve it in the meanwhile, so ignore the noise until you're ready :)

src/Ganeti/JQueue.hs
```diff
@@ -562,6 +563,33 @@ isQueuedJobDead ownlivelock =
...
 
+-- | Waits for a job's process to exit
+waitUntilJobExited :: Livelock -- ^ LuxiD's own livelock
+                   -> JobId -- ^ the job to wait for
+                   -> Int -- ^ timeout in milliseconds
+                   -> ResultG (Bool, String)
+waitUntilJobExited luxiLivelock jid tmout = do
+  let sleepDelay = 100 :: Int -- ms
+      tries = tmout `div` sleepDelay
+      jName = ("Job " ++) . show. fromJobId $ jid
```
iustin: Missing space `show.` (for consistency).

src/Ganeti/JQueue.hs
```diff
@@ -562,6 +563,33 @@ isQueuedJobDead ownlivelock =
...
+  logDebug $ "Waiting for " ++ jName ++ " to exit"
+  result <- runMaybeT . msum . flip map [0..tries :: Int] $ \tryNo -> do
```
iustin: tryNo is confusing here (as opposite of tryYes) - tryNum?

src/Ganeti/JQueue.hs
```diff
@@ -562,6 +563,33 @@ isQueuedJobDead ownlivelock =
...
+    -- if we're retrying, sleep for some time
+    when (tryNo > 0) . liftIO . threadDelay $ sleepDelay * 1000
+
+    qDir <- liftIO queueDir
+    (job, _) <- lift . mkResultT $ loadJobFromDisk qDir True jid
```
iustin: If I understand correctly, you're loading and parsing the job from disk every 100ms. My (weak) understanding is that a job's livelock cannot change once it has its own livelock, in which case it would be better to a) load the job once, and then b) pass the same job to `isQueueJobDead`, or directly to `isDead`?

src/Ganeti/Query/Server.hs
```diff
@@ -507,18 +507,24 @@ handleCall _ qstat  cfg (CancelJob jid kill) = do
...
-handleCall qlock _ cfg (ArchiveJob jid) =
+handleCall qlock qstat cfg (ArchiveJob jid) =
   -- By adding a layer of MaybeT, we can prematurely end a computation
   -- using 'mzero' or other 'MonadPlus' primitive and return 'Ok False'.
   runResultT . liftM (showJSON . fromMaybe False) . runMaybeT $ do
     qDir <- liftIO queueDir
     let mcs = Config.getMasterCandidates cfg
         live = liveJobFile qDir jid
         archive = archivedJobFile qDir jid
+        exitTimeout = (2 * C.rpcConnectTimeout + 1) * 1000 -- milliseconds
```
iustin: Can you explain (in a code comment) why this formula?

src/Ganeti/JQueue.hs
apoikos: Yes, I didn't like loading the job every 100ms either. I'll refactor the loop a bit, thanks!

apoikos:
I have now tested this on a real cluster and indeed it addresses the issue of unarchived jobs on master candidates.

iustin:
Ah, forgot about this. Will try to review in the next couple of days.

A bit "too much: activity on Ganeti lately :)

apoikos:
@iustin: friendly ping :)

PR #1414, opened 2019-12-12
Remove leftover calls to "python" interpreter, enable pylint3/python3-coverage

rbott:
I have tried building on a pure Python 3 Debian system and found some references to /usr/bin/python (which does not exist in this case). These should be fixed now. This allows the build and the python tests to pass.

I will also open a separate issue and list all critical errors pylint3 has found.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2777,7 +2777,7 @@ hlint: $(HS_BUILT_SRCS) src/lint-hints.hs
 	@test -n "$(HLINT)" || { echo 'hlint' not found during configure; exit 1; }
 	@rm -f doc/hs-lint.html
 	if tty -s; then C="-c"; else C=""; fi; \
-	$(HLINT) --utf8 --report=doc/hs-lint.html --cross $$C \
+	$(HLINT) --report=doc/hs-lint.html --cross $$C \
 	  --hint src/lint-hints \
 	  --cpp-file=$(HASKELL_PACKAGE_VERSIONS_FILE) \
 	  $(filter-out $(HLINT_EXCLUDES),$(HS_LIBTEST_SRCS) $(HS_PROG_SRCS))
--- a/Makefile.am
+++ b/Makefile.am
@@ -3030,7 +3030,7 @@ apps/%.hs.stamp: Makefile
 # Builds the cabal file
 ganeti.cabal: cabal/ganeti.template.cabal Makefile cabal/cabal-from-modules.py $(CABAL_EXECUTABLES_APPS_STAMPS)
 	@echo $(subst /,.,$(patsubst %.hs,%,$(patsubst test/hs/%,%,$(patsubst src/%,%,$(HS_SRCS))))) \
-	  | python $(abs_top_srcdir)/cabal/cabal-from-modules.py $(abs_top_srcdir)/cabal/ganeti.template.cabal > $@
+	  | python3 $(abs_top_srcdir)/cabal/cabal-from-modules.py $(abs_top_srcdir)/cabal/ganeti.template.cabal > $@
 
 	for p in $(CABAL_EXECUTABLES); do \
 	  echo                                   >> $@; \
--- a/autotools/ac_python_module.m4
+++ b/autotools/ac_python_module.m4
@@ -26,7 +26,7 @@
 AC_DEFUN([AC_PYTHON_MODULE],[
     if test -z $PYTHON;
     then
-        PYTHON="python"
+        PYTHON="python3"
     fi
     PYTHON_NAME=`basename $PYTHON`
     AC_MSG_CHECKING($PYTHON_NAME module: $1)
--- a/configure.ac
+++ b/configure.ac
@@ -571,9 +571,9 @@ then
                             documentation rebuild not possible]))
 fi
 
-# Check for pylint
-AC_ARG_VAR(PYLINT, [pylint path])
-AC_PATH_PROG(PYLINT, [pylint], [])
+# Check for pylint3
+AC_ARG_VAR(PYLINT, [pylint3 path])
+AC_PATH_PROG(PYLINT, [pylint3], [])
 if test -z "$PYLINT"
 then
   AC_MSG_WARN([pylint not found, checking code will not be possible])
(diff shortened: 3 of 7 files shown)


rbott:
On a sidenode, I have been able to build Debian packages (using the debian folder & pipline found [here](https://salsa.debian.org/rbott-guest/ganeti-build/).

The packages get build, however
* pylint3 (which is not part of the Debian build right now) complains about some rather critical errors (see #1415)
* the runtime packages depend on python-fdsend (this seems to be involved in KVM NIC hotplugging after a quick search in the repository) which does _not_ exist for Python 3 on Debian Buster. I will investigate this a bit more and see if python-fdsend is deprecated/not available for Python 3, if this is only related to Debian and open an issue if we need to do anything about this from the Ganeti end
* at least neither the build itself nor tests seem to pick up that python-fdsend is missing

Makefile.am
```diff
@@ -2777,7 +2777,7 @@ hlint: $(HS_BUILT_SRCS) src/lint-hints.hs
 	@test -n "$(HLINT)" || { echo 'hlint' not found during configure; exit 1; }
 	@rm -f doc/hs-lint.html
 	if tty -s; then C="-c"; else C=""; fi; \
-	$(HLINT) --utf8 --report=doc/hs-lint.html --cross $$C \
```
apoikos: Why is this change necessary?

apoikos:
> On a sidenode, I have been able to build Debian packages (using the debian folder & pipline found [here](https://salsa.debian.org/rbott-guest/ganeti-build/)). While writing this, the packages have been successfully installed by the CI/CD pipeline and a test cluster has been initialized, but the QA suite does not work. I will investigate this and extend this PR with a fix.

Good work! I'm interested in the QA errors. IIRC, you're running against a real cluster, so you probably get more coverage than I did with the vCluster.

> Some more findings:
> 
>     * pylint3 (which is not part of the Debian build right now) complains about some rather critical errors (see #1415)
> 
>     * the runtime packages depend on python-fdsend (this seems to be involved in KVM NIC hotplugging after a quick search in the repository) which does _not_ exist for Python 3 on Debian Buster. I will investigate this a bit more and see if python-fdsend is deprecated/not available for Python 3, if this is only related to Debian and open an issue if we need to do anything about this from the Ganeti end

You should be able to drop the dependency on fdsend, I re-implemented all of it inside ganeti using pure Python (see 493c901bd23c9a891a3426b0468a4db462058af3 and 822b9e0c0319fa723a374c8e43436d28e246321c)

>     * at least neither the build itself nor tests seem to pick up that python-fdsend is missing

apoikos:
> On a sidenode, I have been able to build Debian packages (using the debian folder & pipline found [here](https://salsa.debian.org/rbott-guest/ganeti-build/)). While writing this, the packages have been successfully installed by the CI/CD pipeline and a test cluster has been initialized, but the QA suite does not work. I will investigate this and extend this PR with a fix.

I had a quick look, the error is
```
  File "./ganeti-qa.py", line 43, in <module>
    from qa import colors
ModuleNotFoundError: No module named 'qa'
```
This is because of the changes in ee2bcec93576a80e761205da91cdea89bce636bf together with the fact that `ganeti-qa.py` resides in the `qa/` directory. If you set `PYTHONPATH` to the root of the ganeti source in a built tree, it works. If you need to run the QA suite from a source path that has not been built, you need to set `PYTHONPATH` to `/usr/share/ganeti/default:/path/to/ganeti/source` (not tested).

Makefile.am
rbott: Oh, that slipped into this PR and was not supposed to be here. hlint does not know the `--utf8` parameter (any more?).

However, I will revert this commit as it has nothing to todo with _this_ PR.

apoikos:
@rbott: no need to revert the commit, I was just asking :).

LGTM, but I need to ask a couple of things before merging:
 - Please undo the revert of the hlint commit (in case you need help, you can do so using `git rebase -i origin/master` and by deleting the respective line). It doesn't make sense to open another PR just for this :-)
 - We usually want commits to be signed-off-by, which is done by passing the `-s` flag to `git commit`.
 - Regarding the `hlint` commit, please state in the commit message in which version of hlint the `--utf8` switch was removed, as this is useful information.

rbott:
Thanks @apoikos, extending the PYTHONPATH did the trick. Now we end up somewhere else:
```
<<<< 2019-12-13 09:01:19.125708 start [TestClusterCopyfile] gnt-cluster copyfile
---- FAILED [TestClusterCopyfile] gnt-cluster copyfile: a bytes-like object is required, not 'str'
>>>> 2019-12-13 09:01:19.127310 time=0:00:00.001602 [TestClusterCopyfile] gnt-cluster copyfile
---- BLOCK FAILED [RunClusterTests] Runs tests related to gnt-cluster: a bytes-like object is required, not 'str'
^^^^ BLOCK 2019-12-13 09:01:19.127436 time=0:00:46.385969 [RunClusterTests] Runs tests related to gnt-cluster
  File "./ganeti-qa.py", line 1135, in <module>
    main()
  File "/usr/share/ganeti/default/ganeti/rapi/client.py", line 278, in wrapper
    return fn(*args, **kwargs)
  File "./ganeti-qa.py", line 1130, in main
    RunQa()
  File "./ganeti-qa.py", line 957, in RunQa
    RunTestBlock(RunClusterTests)
  File "./ganeti-qa.py", line 169, in RunTestBlock
    return fn(*args, **kwargs)
  File "./ganeti-qa.py", line 275, in RunClusterTests
    RunTestIf(test, fn)
  File "./ganeti-qa.py", line 150, in RunTestIf
    RunTest(fn, *args, **kwargs)
  File "./ganeti-qa.py", line 112, in RunTest
    retval = fn(*args, **kwargs)
  File "/usr/share/ganeti/testsuite/qa/qa_cluster.py", line 1532, in TestClusterCopyfile
    f.write(uniqueid)
  File "/usr/lib/python3.7/tempfile.py", line 620, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```
But getting the QA suite up and running should probably not be discussed here :-)

Regarding the other issues:
* I will try to undo the revert and also try to figure out more details about the missing `--utf8` parameter
* from this day on, I will `-s` my git commits :-)

Makefile.am
rbott: This has been removed in hlint 2.0:
```
2.0, released 2017-04-06
    [...]
    Remove the --encoding/--utf8 options, always use UTF8
    [...]
```

iustin:
Beware that, as far as I know, pylint3 doesn't exist anymore in unstable. It only existed until pylint (source) 2.2.2-1, afterwards the standard `pylint` package is the python version 3 one, but it doesn't provide a pylint3 alias.

IMHO the pylint maintainers didn't do a good transition here, but that's beside the point.

apoikos:
@iustin you are right, I missed that. Apparently there is only one `pylint` "binary" and it only really supports Python 3 in 2.4. We need to revert parts of 2bd8f04ce0393902620045d2dc676e7d04350747, as this is broken in unstable.

Issue #1415, opened 2019-12-12
Python 3 migration: critical issues found by pylint3


rbott:
#1414 allows to run `make pylint` after the Python 3 migration again (at least when the package `pylint3` is installed).

I have attached the entire pylint3 output to this issue but will also include all *critical* errors in this post:
* ganeti/opcodes_base.py:149:11: E1121: Too many positional arguments for classmethod call (too-many-function-args)
* ganeti/opcodes_base.py:253:22: E1101: Instance of 'BaseOpCode' has no 'OP_ID' member (no-member)
* ganeti/opcodes_base.py:260:24: E1101: Instance of 'BaseOpCode' has no 'OP_ID' member (no-member)
* ganeti/opcodes_base.py:263:36: E1101: Instance of 'BaseOpCode' has no 'OP_ID' member (no-member)
* ganeti/errors.py:445:6: E0633: Attempting to unpack a non-sequence (unpacking-non-sequence)
* ganeti/opcodes.py:160:22: E1102: field_formatter is not callable (not-callable)
* ganeti/tools/burnin.py:1086:14: E0602: Undefined variable 'reduce' (undefined-variable)
* ganeti/jqueue/exec.py:72:4: E1101: Instance of 'closing' has no 'Call' member (no-member)
* ganeti/jqueue/exec.py:74:20: E1101: Instance of 'closing' has no 'Call' member (no-member)
* ganeti/client/gnt_cluster.py:2236:5: E0602: Undefined variable 'IsBelowDir' (undefined-variable)
* ganeti/hypervisor/hv_xen.py:1072:4: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
* ganeti/hypervisor/hv_lxc.py:96:11: E1101: Instance of 'LXCVersion' has no 'original_string' member (no-member)
* ganeti/utils/__init__.py:990:9: E0602: Undefined variable 'array' (undefined-variable)
* ganeti/utils/__init__.py:994:6: E0602: Undefined variable 'array' (undefined-variable)
* ganeti/utils/__init__.py:998:6: E0602: Undefined variable 'array' (undefined-variable)
* ganeti/utils/retry.py:54:32: E1136: Value 'self.args' is unsubscriptable (unsubscriptable-object)
* ganeti/utils/retry.py:55:12: E1136: Value 'self.args' is unsubscriptable (unsubscriptable-object)
[...]

PR #1416, opened 2019-12-13
Fix issues uncovered by pylint3

apoikos:
This fixes some of the issues uncovered in #1415.

### DIFF
--- a/autotools/check-header
+++ b/autotools/check-header
@@ -138,7 +138,7 @@ def Main():
     lines = zip(itertools.count(1), content.splitlines())
 
     try:
-      _CheckHeader(compat.partial(lines.pop, 0))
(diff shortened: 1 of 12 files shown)


iustin:
Are you sure all these overrides are still present with 2.4? pylint3 in Debian is old, as I've commented on the other patch.

Or put another way, I wonder why so many false positives.

apoikos:
> Are you sure all these overrides are still present with 2.4? pylint3 in Debian is old, as I've commented on the other patch.
> 
> Or put another way, I wonder why so many false positives.

I was testing with 2.2.2, I'll test again with 2.4 and see how it goes.

apoikos:
Yes, 2.4 seems to produce the same false positives as 2.2.

PR #1422, opened 2019-12-16
RAPI client: fix Python 3 compatibility

apoikos:
Under Python 3, PycURL writes bytes to the write function, rather than
strings. This is currently broken because we allocate a StringIO buffer
that can't handle bytes. Rather than do version guesswork, we add a shim
class that behaves like a Reader and a Writer, but uses BytesIO or
StringIO under the hood based on the first write's type.

Note that this only matters for PycURL, simplejson has no problem
loading data from binary file-like objects.

### DIFF
--- a/lib/rapi/client.py
+++ b/lib/rapi/client.py
@@ -55,11 +55,7 @@ except ImportError:
 import pycurl
 import simplejson
 
-try:
-  from io import StringIO
-except ImportError:
-  from io import StringIO
-
+from io import StringIO, BytesIO
 
 GANETI_RAPI_PORT = 5080
 GANETI_RAPI_VERSION = 2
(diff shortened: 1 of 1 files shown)


rbott:
Our basic problem seems to be solved (this is one of several RAPI tests from the QA suite which works now flawlessly):
```
<<<< 2019-12-17 17:36:55.785364 start [TestRapiNodeGroups] Test several node group operations using RAPI
POST /2/groups
GET /2/jobs/140
POST /2/groups
GET /2/jobs/141
PUT /2/groups/group1/modify
GET /2/jobs/142
PUT /2/groups/group2/rename
GET /2/jobs/143
Testing get/modify symmetry of /2/groups/group3 and /2/groups/group3/modify
GET successful at /2/groups/group3
PUT successful at /2/groups/group3/modify
DELETE /2/groups/group1
GET /2/jobs/145
DELETE /2/groups/group3
GET /2/jobs/146
---- PASSED [TestRapiNodeGroups] Test several node group operations using RAPI
>>>> 2019-12-17 17:37:02.387442 time=0:00:06.602078 [TestRapiNodeGroups] Test several node group operations using RAPI
```

The QA run still fails but that is (as far as I can tell) due to the issue fixed in 6a4cb4125 (which is not currently visible to this PR). Details from test run (including the full logs in the job artifacts) can be found here: https://salsa.debian.org/rbott-guest/ganeti-build/-/jobs/460820

PR #1425, opened 2019-12-21
getNodeInstances: count each instance at most once

### LINKED ISSUE #1399: gnt-node list shows wrong Sinst count (sinst_cnt)
# Version info.
# gnt-cluster version
Software version: 2.16.1
Internode protocol: 2160000
Configuration format: 2160000
OS api version: 20
Export interface: 0
VCS version: (ganeti) version 2.16.1-9

The command "gnt-node list" shows the wrong "Sinst" count.

Running the following command shows the instances (names) multiple times, once for each disk/drbd device the instance has (in this case 3 disks per instance).

# gnt-node list --no-headers -o sinst_list testnode2 | tr ',' '\n'
instance1
instance1
instance1
instance2
instance2
instance2
instance3
instance3
instance3
instance4
instance4
instance4

It seems like the data is received from luxid at this line https://github.com/ganeti/ganeti/blob/0dfa4d212ae2fc216db0249d5f6e0b4aa22ac059/lib/cli.py#L2035 and i do not have any haskell skills. ¯\_(ツ)_/¯

apoikos:
Commit 9825767a ("Tune getNodeInstances DRBD secondary computation")
refactored the way we compute secondary instances for a node, by
iterating over an instance's disk and extracting the secondary node.
Unfortunately, this introduced a bug where each instance is counted as
many times as its disk count.

We want to consider each instance only once, so just check if any of the
instance disks have the node in question as secondary.

This fixes #1399.

### DIFF
--- a/src/Ganeti/Config.hs
+++ b/src/Ganeti/Config.hs
@@ -89,7 +89,7 @@ import Control.Monad.State
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.UTF8 as UTF8
 import qualified Data.Foldable as F
-import Data.List (foldl', nub)
+import Data.List (foldl', nub, any)
 import Data.Maybe (fromMaybe, mapMaybe)
 import Data.Monoid
 import qualified Data.Map as M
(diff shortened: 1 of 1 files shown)


apoikos:
Note, I've verified this works in a test cluster. Unfortunately there are no tests for Ganeti.Config and I currently don't have time to implement them.

Issue #1417, opened 2019-12-13
Python 3 Migration: Fix various documentation issues and provide up-to-date Dockerfile


rbott:
The current documention files contain various pointers to outdated Python 2 stuff which needs to fixed for Ganeti 3.0.

Furthermore, there is only the older Dockerfile.stretch which is of no use for current tests/builds so we need an updated version.

apoikos:
The Dockerfile is irrelevant and can be safely removed.

rbott:
This has been addressed in #1428 and #1438 

Hence, we can close this issue for now.

PR #1428, opened 2019-12-22
Documentation update

rbott:
This PR partially addresses #1417 and also updates other aspects of the documentation (e.g. external links to dependencies).

Unfortunately I do not have access to yum based systems like Fedora or RHEL so I can not update the examples for these systems (yet).


rbott:
As per @iustin's review I dropped the commit which updated the repository's main README to show Ganeti Version 3.0. Unfortunately that also took away his comment. Sorry for that :-)

Issue #1300, opened 2018-05-09
shutdown-timeout


ganwell:
It is not explicitly documented that shutdown timeout is in minutes (only that default is 2 minutes):

```gnt-instance reboot --shutdown-timeout=10 -t hard vm``` will this set a shutdown timeout of 10min?

ge-fa:
feature: docs

saschalucas:
fixed in #1429

PR #1429, opened 2020-01-01
document the unit of cli option shutdown-timeout

saschalucas:
Clarify the unit of measurement for the cli option --timeout respectively
--shutdown-timeout in gnt-instance(8), gnt-backup(8) and their command
help output. This fixes issue #1300.

Signed-off-by: Sascha Lucas <sascha_lucas@web.de>


lib/cli_opts.py
```diff
@@ -1204,7 +1204,7 @@ class CliOption(Option):
 
 TIMEOUT_OPT = cli_option("--timeout", dest="timeout", type="int",
                          default=constants.DEFAULT_SHUTDOWN_TIMEOUT,
-                         help="Maximum time to wait")
+                         help="Maximum number of minutes to wait")
```
iustin: Not sure "number of minutes" sounds idiomatic English.

Maybe reword into "How much time (in minutes) to wait"? Or something like this. Same below.

PR #1431, opened 2020-01-20
Fix typo in gnt-instance man page.

volans-:
Fix a single char typo that I noticed while reading the man page.


volans-:
@googlebot I signed it!

PR #1433, opened 2020-02-02
Support Live Migration with QEMU 4.2+ (use QMP for migration parameters)

rbott:
This PR implements setting the QEMU/KVM live migration parameters through the QMP function `migrate-set-parameters`. The old way is now deprecated (see [QEMU 4.2 changelog](https://wiki.qemu.org/ChangeLog/4.2#New_deprecated_options_and_features)

This change has been tested against Debian Bullseye (which contains QEMU 4.2) and also works on Debian Buster (which contains QEMU 3.1).

Parameters that could be set through the QMP function are listed here:
https://github.com/qemu/qemu/blob/28db64fce555a03b4ca256d5b6f4290abdfbd9e8/qapi/migration.json#L734-L826

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2555,13 +2555,14 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     if not live_migration:
       self._CallMonitorCommand(instance_name, "stop")
 
-    migrate_command = ("migrate_set_speed %dm" %
-                       instance.hvparams[constants.HV_MIGRATION_BANDWIDTH])
-    self._CallMonitorCommand(instance_name, migrate_command)
-
-    migrate_command = ("migrate_set_downtime %dms" %
-                       instance.hvparams[constants.HV_MIGRATION_DOWNTIME])
-    self._CallMonitorCommand(instance_name, migrate_command)
+    qmp = QmpConnection(self._InstanceQmpMonitor(instance_name))
+    qmp.connect()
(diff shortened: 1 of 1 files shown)


rbott:
@apoikos or @iustin could you please take a look at this? If you need more information, I'm happy to help out.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2555,13 +2555,14 @@ def MigrateInstance(self, cluster_name, instance, target, live_migration):
     if not live_migration:
       self._CallMonitorCommand(instance_name, "stop")
```
apoikos, line 2556: How about switching the whole migration logic to QMP, while at it?

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2555,13 +2555,14 @@ def MigrateInstance(self, cluster_name, instance, target, live_migration):
...
 
-    migrate_command = ("migrate_set_speed %dm" %
-                       instance.hvparams[constants.HV_MIGRATION_BANDWIDTH])
-    self._CallMonitorCommand(instance_name, migrate_command)
-
-    migrate_command = ("migrate_set_downtime %dms" %
-                       instance.hvparams[constants.HV_MIGRATION_DOWNTIME])
-    self._CallMonitorCommand(instance_name, migrate_command)
+    qmp = QmpConnection(self._InstanceQmpMonitor(instance_name))
```
apoikos, line 2558: Minor, but there's the `@_with_qmp` decorator that makes a QMP connection available at `self.qmp`. This will be especially handy if you decide to switch everything over to QMP.

rbott:
I have reworked this PR completely and therefore created a new one, #1445. I will close this PR here.

Issue #1423, opened 2019-12-17
gnt-cluster upgrade does not support major version upgrades


apoikos:
As it stands, gnt-cluster upgrade will abort when asked to upgrade to a different major version:

https://github.com/ganeti/ganeti/blob/9d1c128b3135ea0aa17397bc7e1b8d56b542dde5/lib/client/gnt_cluster.py#L2434-L2437

https://github.com/ganeti/ganeti/blob/9d1c128b3135ea0aa17397bc7e1b8d56b542dde5/lib/utils/version.py#L109-L130

This means that `gnt-cluster upgrade --to=3.0` will not work out of the box, and we have to think how we'll address this.

rbott:
Just for clarification: this needs to be most likely addressed in a new 2.16 release as the 2.x code will refuse to upgrade to a new major version. I originally assumed you only need to alter the code in 3.x to allow upgrades _from_ 2.x

However, @apoikos pointed me to the problem that after an installation of ganeti 3, the cluster still runs ganeti 2.x code and the upgrade routines of ganeti 2.x do the relevant steps and change all symlinks to the 3.x binaries/paths at the end.

So we either need support in ganeti 2.15/2.16 for this or we need to ship a tool with ganeti 3 which works around this limitation (sort of 2to3 "ganeti edition")

atta:
somthing like,

MAJOR_UPGRADE_VERSION = (2, 15, 0) ???

https://github.com/ganeti/ganeti/blob/a50e5bb77e20acc9f627ace36cd2c79dade46d7e/lib/utils/version.py#L124

rbott:
This will be / has been addressed in #1449 and #1450

saschalucas:
seems to be solved now: PRs #1449 for 2.16 and #1451 for 3.0

PR #1444, opened 2020-04-29
allow up-/downgrades between 2.16 and 3.0

saschalucas:
This will allow up- and downgrades between version 2.16 and 3.0.
Hence the minimum Python version needed is 3.6, Ganeti-3.0 will only
be available for Debian Buster/Ubuntu Bionic or later. Both ship
with Ganeti-2.16. Apparently no other versions are needed.

Probably not the best program flow and hard coded exceptional versions.

Limited testing, by running UpgradeRange() standalone (extra python script) and feeding different target/current versions (no real upgrade test done).

This is against master, for allowing downgrade from upcoming 3.0. Needs also merging into stable-2.16 for a new 2.16 release, allowing the upgrade.

this addresses issue #1423

### DIFF
--- a/lib/utils/version.py
+++ b/lib/utils/version.py
@@ -122,7 +122,16 @@ def UpgradeRange(target, current=CURRENT_VERSION):
     return "automatic upgrades only supported from 2.10 onwards"
 
   if target[0] != current[0]:
-    return "different major versions"
+    # allow major upgrade from 2.16 to 3.0
+    if current[0:2] == (2,16) and target[0:2] == (3,0):
+      return None
+    # allow major downgrade from 3.0 to 2.16
+    if current[0:2] == (3,0) and target[0:2] == (2,16):
+      return None
+
+    # forbid any other major version up-/downgrades
+    return "major version up- or downgrades are only supported between " \
+      "2.16 and 3.0"
 
   if target[1] < current[1] - 1:
     return "can only downgrade one minor version at a time"
--- a/src/Ganeti/Config.hs
+++ b/src/Ganeti/Config.hs
@@ -88,7 +88,7 @@ import Control.Monad.State
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.UTF8 as UTF8
 import qualified Data.Foldable as F
-import Data.List (foldl', nub)
+import Data.List (foldl', nub, any)
 import Data.Maybe (fromMaybe, mapMaybe)
 import Data.Monoid
 import qualified Data.Map as M
@@ -184,10 +184,8 @@ getNodeInstances cfg nname =
         sec_insts :: [Instance]
         sec_insts = [inst |
           (inst, disks) <- inst_disks,
-          s_uuid <- mapMaybe (\d ->
-                              instPrimaryNode inst >>=
-                              computeDiskSecondaryNode d) disks,
-          s_uuid == nname]
+          any ((==) nname) $ mapMaybe (\d -> instPrimaryNode inst >>=
+                                       computeDiskSecondaryNode d) disks]
     in (pri_inst, sec_insts)
 
 -- | Computes the role of a node.
--- a/test/py/ganeti.utils.version_unittest.py
+++ b/test/py/ganeti.utils.version_unittest.py
@@ -56,15 +56,21 @@ class UpgradeRangeTest(unittest.TestCase):
         self.assertEqual(version.UpgradeRange((2,11,3), current=(2,12,99)),
                           None)
         self.assertEqual(version.UpgradeRange((3,0,0), current=(2,12,0)),
-                          "different major versions")
+                          "major version up- or downgrades are only supported " \
+                          "between 2.16 and 3.0")
         self.assertEqual(version.UpgradeRange((2,12,0), current=(3,0,0)),
-                          "different major versions")
+                          "major version up- or downgrades are only supported " \
+                          "between 2.16 and 3.0")
         self.assertEqual(version.UpgradeRange((2,10,0), current=(2,12,0)),
                           "can only downgrade one minor version at a time")
         self.assertEqual(version.UpgradeRange((2,9,0), current=(2,10,0)),
                           "automatic upgrades only supported from 2.10 onwards")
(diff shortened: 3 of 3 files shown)


saschalucas:
hm... python tests fails:

```text
Running ganeti.utils.version_unittest.py
.....F
======================================================================
FAIL: testUpgradeRange (__main__.UpgradeRangeTest)
----------------------------------------------------------------------
  File "test/py/ganeti.utils.version_unittest.py", line 59, in testUpgradeRange
    "different major versions")
AssertionError: 'major version up- or downgrades are only supported between 2.16 and 3.0' != 'different major versions'
- major version up- or downgrades are only supported between 2.16 and 3.0
+ different major versions
```

need to investigate further.

atta:
cool thanks for the work .. 
i think we can also allow upgrades from 2.15 but i'am not 100% shure about it

rbott:
So this means this PR needs to be applied to the master branch to allow downgrades from 3.0 installations back to 2.16 - but we also need this in a new 2.16.2 release to allow upgrades from 2.16 to 3.0 (and have distributions pick this up in current releases).

From a Ganeti point of view this seems to be the most feasible way, but it will make it harder for people to upgrade (because it depends on new package releases in stable distributions).

What do you think about adding a oneshot Python script to Ganeti 3.0 which will allow upgrading to 3.0 from current 2.16 versions (and possibly 2.15)? As there are no structural changes to the database etc. it should mostly be "replacing some symlinks", right?

saschalucas:
@rbott: for users not receiving new releases, the greatest flexibility would be, to point them to the relevant code line in lib/utils/version.py, so that they can help themselves? What I can image is some kind of helper script doing a live patch, with the help of the user who needs to point the script to lib/utils/version.py. This would cover all kind of installs (from source, build or packed without `--enable-symlinks` or with `--enable-versionfull`, lib or lib64 as libdir etc.).

ge-fa:
> From a Ganeti point of view this seems to be the most feasible way, but it will make it harder for people to upgrade (because it depends on new package releases in stable distributions).

Just a note regarding the Debian-side of things: It's quite possible to get an update into Debian buster to allow people to upgrade to do the upgrade dance from 2.16 to 3.0. The patch needs to be targeted and has to apply on top what's currently in Debian buster, but that should be easily doable.

rbott:
Does anyone have the time to actually test this on a real cluster? I might be able to test it during the weekend (on a nested KVM three node test cluster with Debian Buster). That will only cover DRBD backed clusters but honestly I do not see why the storage backend of the cluster should be an issue here.

iustin:
This looks good, but it would easier if it's submitted against 2.16 branch first, and then that branch merged to master. Doing it the other way around means a cherry-pick, rather than a merge.

Do you see problems doing it that way?

atta:
Hi Folks,

i thested the upgarde on my cluster and it also works.

```
root@node01:~# gnt-cluster upgrade --to=3.0                                                                                                                                                                                                                                                                                                                                                                                                                                                    
Verifying 3.0 present on all nodes                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Draining queue                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
Pausing the watcher for one hour.                                                                                                                                                                                                                                                                                                                                                                                                                                                              
Stopping daemons on master node.                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Stopping daemons everywhere.                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
Backing up configuration as /var/backups/ganeti1589097334.tar                                                                                                                                                                                                                                                                                                                                                                                                                                  
Switching to version 3.0 on all nodes                                                                                                                                                                                                                                                                                                                                                                                                                                                          
Upgrading configuration                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
Starting daemons everywhere.                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
Redistributing the configuration.                                                                                                                                                                                                                                                                                                                                                                                                                                                              
Restarting daemons everywhere.                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
Undraining the queue.                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
Running post-upgrade hooks                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
Unpausing the watcher.                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
Verifying cluster.                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
root@node01:~# gnt-cluster version                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Software version: 3.0.0~alpha1                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
Internode protocol: 3000000                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
Configuration format: 3000000                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
OS api version: 20                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Export interface: 0                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
VCS version: (ganeti) version 3.0.0-1~nightly20200510002326~09761f5 
```

atta:
> This looks good, but it would easier if it's submitted against 2.16 branch first, and then that branch merged to master. Doing it the other way around means a cherry-pick, rather than a merge.
> 
> Do you see problems doing it that way?
we have to povide a patch to ganeti 2.16 so you are right weh pushing it first to 2.16

atta:
> Does anyone have the time to actually test this on a real cluster? I might be able to test it during the weekend (on a nested KVM three node test cluster with Debian Buster). That will only cover DRBD backed clusters but honestly I do not see why the storage backend of the cluster should be an issue here.

my cluster is using shared-storage and it just works

rbott:
Oh wow, I entirely broke this PR by following the guides how to apply it to a different upstream branch :facepalm: 

I guess it would be easier to create a new PR against the stable-2.16 branch with just the two relevant commits in them. @saschalucas could you do that and create a new PR? We can reference this one and close it afterwards. Sorry for the mess :-(

saschalucas:
@iustin: thanks, will PR against stable-2.16

@rbott: git happens :-), will close this an PR again

new PR is #1449

PR #1445, opened 2020-05-03
KVM migration: use QMP instead of monitoring socket

rbott:
Starting with QEMU version 4.2 the socket commands migrate_set_downtime,
migrate_set_speed and migrate_set_cache_size have been deprecated and
superseeded by the QMP call migrate-set-parameters. This patch addresses
this change and also tries to eleminate all other calls to the monitoring
socket throughout the migration process.

For more information, see the changelog here:
https://wiki.qemu.org/ChangeLog/4.2#New_deprecated_options_and_features
The required parameters are available via QMP since QEMU 2.8

Signed-off-by: Rudolph Bott <bott@sipgate.de>

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -165,14 +165,19 @@ _MIGRATION_CAPS_DELIM = ":"
 
 def _with_qmp(fn):
   """Wrapper used on hotplug related methods"""
-  def wrapper(self, instance, *args, **kwargs):
+  def wrapper(self, *args, **kwargs):
     """Create a QmpConnection and run the wrapped method"""
     if not getattr(self, "qmp", None):
+      for arg in args:
+        if isinstance(arg, objects.Instance):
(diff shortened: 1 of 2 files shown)


rbott:
I had to apply some changes to the @_with_qmp decorator:

https://github.com/ganeti/ganeti/blob/c66db868aad54bda6a829545913775d098595b15/lib/hypervisor/hv_kvm/__init__.py#L170-L176

This is rather ugly, but the older decorator code assumes an instance object as first function parameter (see functions like VerifyHotplugSupport, HotAddDevice etc.). Unfortunately that does not apply to MigrateInstance:  https://github.com/ganeti/ganeti/blob/c66db868aad54bda6a829545913775d098595b15/lib/hypervisor/hv_kvm/__init__.py#L2532

Changing the signature of this exposed function is probably not a good idea, so I changed the decorator to find the first instance of a Ganeti Instance object within the function's arguments.

I am happy to incorporate better ways of handling this, just let me know!

PR #1449, opened 2020-05-10
allow up-/downgrades between 2.16 and 3.0

saschalucas:
This will allow up- and downgrades between version 2.16 and 3.0.
Hence the minimum Python version needed is 3.6, Ganeti-3.0 will only
be available for Debian Buster/Ubuntu Bionic or later. Both ship
with Ganeti-2.16. Apparently no other versions are needed.

Probably not the best program flow and hard coded exceptional versions.

Limited testing, by running UpgradeRange() standalone (extra python script) and feeding different target/current versions (no real upgrade test done).

This is against stable-2.16, for allowing upgrade to upcoming 3.0. Needs also merging into master for allowing downgrade back to 2.16.

this addresses issue #1423

### DIFF
--- a/lib/utils/version.py
+++ b/lib/utils/version.py
@@ -122,7 +122,16 @@ def UpgradeRange(target, current=CURRENT_VERSION):
     return "automatic upgrades only supported from 2.10 onwards"
 
   if target[0] != current[0]:
-    return "different major versions"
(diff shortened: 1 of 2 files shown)


rbott:
As already discussed and tested in #1444, this looks good!

PR #1450, opened 2020-05-10
allow downgrades from 3.0 to 2.16

rbott:
This PR cherry-picks the "allow upgrades from 2.16 to 3.0" from the stable-2.16 branch into master - and hence allows downgrades from 3.0 back to 2.16.

### DIFF
--- a/lib/utils/version.py
+++ b/lib/utils/version.py
@@ -122,7 +122,16 @@ def UpgradeRange(target, current=CURRENT_VERSION):
     return "automatic upgrades only supported from 2.10 onwards"
 
   if target[0] != current[0]:
-    return "different major versions"
+    # allow major upgrade from 2.16 to 3.0
+    if current[0:2] == (2,16) and target[0:2] == (3,0):
+      return None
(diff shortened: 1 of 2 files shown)


rbott:
Either my CLA has expired or the googlebot is suddenly confused that my github account knows two different email addresses. I'll check that.

saschalucas:
it's probably me, that had to:

@googlebot I consent

but @rbott: Don't really know, but Iustin said stable-2.16 can be merged into master, so that no cherry-pick is necessary?

saschalucas:
@googlebot I consent

saschalucas:
BTW: while test-merging stable-2.16 into master, it turns out, that commit a3441c7338404b0b8fb9564a6ba2ac8d71e9cd36 is not yet in master, which I think it should be?

the merge is also not fast-forward, so a merge commit is necessary.

iustin:
Yes, stable-X.Y → master should **always** be done via merge.

saschalucas:
closing this PR, because it is solved in #1451 (merge rather than cherry-pick). never the less thanks, to @rbott.

PR #1470, opened 2020-05-27
Add check-local to Github workflow

rbott:
This PR will add `make check-local` to the Github Actions workflow - hence add additional checks which uncover release-critical errors.

The NEWS file in the current master does not pass this test, but a fixed one is on its way with the `prepare-3.0` branch. Once that has been merged, I will rebase this PR.

Until then, reviews are welcome :-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      LC_ALL: C
     steps:
       - uses: actions/checkout@v1
 
@@ -23,6 +25,11 @@ jobs:
         with:
           args: make -j 2
 
+      - name: Run local checks
+        uses: docker://ganeti/ci:buster-py3
+        with:
+          args: make -j 2 check-local
+
       - name: Python tests
         uses: docker://ganeti/ci:buster-py3
         with:
@@ -32,3 +39,4 @@ jobs:
         uses: docker://ganeti/ci:buster-py3
         with:
           args: make -j 2 hs-tests
+
--- a/Makefile.am
+++ b/Makefile.am
@@ -2598,7 +2598,7 @@ ganeti:
 .PHONY: check-dirs
 check-dirs: $(GENERATED_FILES)
 	@set -e; \
-	find . -type d -not -name '.' -not -path './.*' | { \
+	find . -type d \( -name . -o -name .git -prune -o -name .github -prune -o -print \) | { \
 	  error=; \
 	  while read dir; do \
 	    case "$$dir" in \
--- a/Makefile.am
+++ b/Makefile.am
@@ -2598,7 +2598,7 @@ ganeti:
 .PHONY: check-dirs
 check-dirs: $(GENERATED_FILES)
 	@set -e; \
-	find . -type d -not -name '.' -not -path '*/\.*' | { \
+	find . -type d -not -name '.' -not -path './.*' | { \
 	  error=; \
 	  while read dir; do \
 	    case "$$dir" in \
--- a/Makefile.am
+++ b/Makefile.am
@@ -2598,7 +2598,7 @@ ganeti:
 .PHONY: check-dirs
 check-dirs: $(GENERATED_FILES)
 	@set -e; \
-	find . -type d \( -name . -o -name .git -prune -o -print \) | { \
+	find . -type d -not -name '.' -not -path '*/\.*' | { \
 	  error=; \
 	  while read dir; do \
 	    case "$$dir" in \
@@ -2657,7 +2657,8 @@ check-local: check-dirs check-news $(GENERATED_FILES)
 	    error=1; \
 	  fi; \
 	done; \
-	test -z "$$error"
+	test -z "$$error" ; \
+	echo "All checks passed successfully"
 
 .PHONY: hs-test-%
 hs-test-%: test/hs/htest


rbott:
hmmm, I force-pushed a change to the workflow file but for some reason github does not pick up this change :-(
This PR is now doomed to fail every time, as there are leftovers from the unit tests which confuse check-local.

rbott:
OK, now it fails as expected with the format of the NEWS file. That check should pass as soon as the `prepare-3.0` branch gets merged into master. We need to keep this open until then - but nevertheless please review :-)

Makefile.am
```diff
@@ -2598,7 +2598,7 @@ ganeti:
 .PHONY: check-dirs
 check-dirs: $(GENERATED_FILES)
 	@set -e; \
-	find . -type d \( -name . -o -name .git -prune -o -print \) | { \
+	find . -type d -not -name '.' -not -path '*/\.*' | { \
```
saschalucas, line 2601: Would it be sufficient to limit this to the root directory? i.e.
`find . -type d -not -name '.' -not -path './.*`
would be a more conservative approach. ATM I can not image what tool might clutter subdirs with .dot-folders? Just for curiosity: must `.` be escaped in `-path`?

.github/workflows/ci.yml
```diff
@@ -5,6 +5,8 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      LC_ALL: C
```
saschalucas, line 9: Would it make sense to add this to Makefile.am? This way CI and manual invocation `make check-local` can be covered? AFAIK the check-news target is the only one, that requires LC_ALL=C. There are other parts in Makefile.am, where LC_ALL is set accordingly (search for LC_ALL).

saschalucas:
currently CI fails: 
```bash
RELEASE=3.0.0~beta1 ./autotools/check-news < ./NEWS
Line 8: Unreleased version after current release 3.0.0 beta1
make: *** [Makefile:4863: check-news] Error 1
```
I assume this is expected and at the next release, incrementing the version in `configure.ac` is the last step we should do?

Makefile.am
apoikos, line 2601: @rbott, why is this change needed? The reason .git is ignored is because the build system knows how to filter it out from the final sources. If there's anything the build system does _not_ know how to handle, then I'd like it to fail in `check-local`.

apoikos:
> OK, now it fails as expected with the format of the NEWS file. That check should pass as soon as the `prepare-3.0` branch gets merged into master. We need to keep this open until then - but nevertheless please review :-)

The breakage is expected (although the message is misleading), because we already bumped the version in `configure.ac` without setting a release date in `NEWS`. Under normal circumstances, this should not happen.

Makefile.am
saschalucas: if you run `LC_ALL=C make check-local` it prints out:
```
Directory ./.github not listed in Makefile
Directory ./.github/workflows not listed in Makefile
make: *** [Makefile:4843: check-dirs] Error 1
```
The intention I assume, was to exclude `.dot-folders` at all (including .github). @apoikos: it sounds like this is the wrong way?

rbott: Thanks for the hin, that was probably a bit too much. `find . -type d -not -name '.' -not -path './.*'` works just as good and is easier to read :-) I will fix that!

saschalucas: While the `check-local` target is also intended to be run locally by the developer and therefore it sounds good to exclude all `.dot-folders`, here it will no longer check for `.hpc` directory, which is enumerate in: https://github.com/ganeti/ganeti/blob/72fcdea9139c22201d6b19cc60a0525c2a345fe4/Makefile.am#L270 To get `check-local` into the CI, I suggest a minimal invasive way by adding .github only to the excludes dirs:
```make
check-dirs: $(GENERATED_FILES)
        @set -e; \
        find . -type d \( -name . -o -name .git -o - name .github -prune -o -print \) | { \
``` 
Probably we can see later, how to exclude more things locally.

.github/workflows/ci.yml
saschalucas: looking deeper, it is the `check-news` target, that enforces the C locale: https://github.com/ganeti/ganeti/blob/72fcdea9139c22201d6b19cc60a0525c2a345fe4/autotools/check-news#L97-L101
Because `check-local` should run by CI and locally (includes `check-news`) I suggest to fix this in `Makefile.am`:
```make
check-news:
        LC_ALL=C RELEASE=$(PACKAGE_VERSION) $(CHECK_NEWS) < $(top_srcdir)/NEWS
```

rbott:
I am not entirely sure this PR / `make check-local` (or more specific: the versioning/NEWS part) will add or resolve more problems. So maybe we should drop the commit which adds it to the Github Actions workflow?

Makefile.am
rbott: Good catch, I'll update that accordingly!

rbott: I have updated this PR to only exclude .git and .github folders. This way we do not falsly exclude needed folders and also have a clean list of what is supposed to be excluded.

.github/workflows/ci.yml
rbott: I was wondering: could there be any other parts of the build process which actually rely on LC_ALL=C? Just setting it for a specific build step might break things at a later point.

Makefile.am
rbott: This has been fixed with the latest commit.

.github/workflows/ci.yml
iustin: I think it would be a good idea to make entire build work work an UTF-8 locale, even if C.UTF-8. Plain C seems so archaic :)

rbott: Sounds reasonable to me. But I would propose a separate PR for this. I suppose this should receive some through testing (especially regarding documentation). In any case I'd take a shot at it.

iustin: Oh, definitely didn't mean in this PR. Was just surprised seeing we rely on C...

rbott:
Pinging @saschalucas @iustin @apoikos: what is your current take on this PR?

saschalucas:
> Pinging @saschalucas @iustin @apoikos: what is your current take on this PR?

I'm ok with this PR, if the locale setting gets moved from `.github/workflows/ci.yml` to `Makefile.am`, so that `check-local` can be run manually, too. Later we should investigate why different parts of the code need different locales and move it to a single common locale.

rbott:
Interestingly, @apoikos implemented C (or rather: C.UTF-8) partially here in the Makefile: https://github.com/ganeti/ganeti/blob/4d8c16a0739c93400471023b3cf9adf73a037b8f/Makefile.am#L35
Related commit is 8563fdf57574e14d1a0da0727d11622dbc9f9145

The comment states `Use C.UTF-8 for tests/checks when available, fall back to en_US.UTF-8 else`, however the locale gets only set/used in the context of man pages and API docs. I would suggest to use this as a general setting for the whole build process. Does this seem right to everyone?

rbott:
I will close this PR for now (but take a look at the locale thing discussed above and possibly open a new PR for that).

Issue #1452, opened 2020-05-13
GHC show an error after compiles Exec.hs file.


rchatsiri:
I'm build source code from master branch. GHC compiler shows an error as below.
```
[GHC]: src/Ganeti/Query/Exec.o <- cabal_macros.h src/Ganeti/Query/Exec.hs src/Ganeti/UDSServer.hi src/Ganeti/UDSServer.o src/Ganeti/Types.hi src/Ganeti/Types.o src/Ganeti/Path.hi src/Ganeti/Path.o src/Ganeti/OpCodes.hi src/Ganeti/OpCodes.o src/Ganeti/Logging/WriterLog.hi src/Ganeti/Logging/WriterLog.o src/Ganeti/Logging.hi src/Ganeti/Logging.o src/Ganeti/JSON.hi src/Ganeti/JSON.o src/Ganeti/JQueue/Objects.hi src/Ganeti/JQueue/Objects.o src/Ganeti/BasicTypes.hi src/Ganeti/BasicTypes.o src/AutoConf.hi src/AutoConf.o

src/Ganeti/Query/Exec.hs:120:12: error:
    Variable not in scope:
      getPid :: ProcessHandle -> IO (Maybe ProcessID)
```
Additional information
* The Glorious Glasgow Haskell Compilation System, version 8.0.2.
* cabal-install version 1.24.0.2.

saschalucas:
According to the named versions, I assume this is an Ubuntu 18.04/Bionic build? If not please specify your Distribution and/or the source of your GHC environment (i.e. Stack or Haskell Platform). You may also attach the output of `cabal list --simple-output`

saschalucas:
in Ubuntu 18.04 the Haskell package `process` is version 1.4.3.0, but in commit 219b2ab926c8d2ec6157c6e87636d1b7ba031789 getPid was introduced in https://github.com/ganeti/ganeti/blob/fc0b28b91a92a58850676dd077aa9f89d7a81958/src/Ganeti/Query/Exec.hs#L119, which is only available since [process-1.6.3.0](https://hackage.haskell.org/package/process-1.6.3.0/changelog).

Therefore we need to update the Haskell requirements https://github.com/ganeti/ganeti/blob/fc0b28b91a92a58850676dd077aa9f89d7a81958/cabal/ganeti.template.cabal#L55.

ATM this would mean, that a build on stock Ubuntu-18.04 packages is not possible.

saschalucas:
[...]

PR #1474, opened 2020-05-31
Ubuntu 18.04 compatibility

saschalucas:
This PR addresses building Ganeti-3.0 from source on Ubuntu 18.04 (Bionic). Therefore two things need to be done:

### lower dependency of cryptonite to 0.23
Ubuntu 18.04 ships with Haskell library cryptonite-0.23, where Debain Buster ships 0.25. It is assumed, that there is no explicit dependency for the Buster version.

### support Haskell process before 1.6.3
Commit 219b2ab9 moves away from unreliable forkProcess to System.Process.createProcess and friends. The process library >=1.6.3 introduced the getPid function ([see here](https://hackage.haskell.org/package/process-1.6.3.0/changelog)), which is used in the mentioned commit: https://github.com/ganeti/ganeti/blob/bdeeccedd4cdeec5bee40145e71b876f427e1109/src/Ganeti/Query/Exec.hs#L119 On Ubuntu 18.04 there is only process-1.4.3 available. This PR introduces a wrapper function, that conditionally switches to an included implementation, if process is < 1.6.3

This will support Ganeti-3.0 builds from source on Ubuntu 18.04 (Bionic). ATM it is unknown if this PR would pass qa-suite on Ubuntu-18.04. At least it compiles and passes unit tests.

Because I'm a total Haskell moron, review should be done by an Haskell expert. I used copy&paste like methods (from [original code](https://github.com/haskell/process/pull/109/files)) + trail and error, not understanding much of the syntax/concepts.

Everybody is invited to discuss, whether to support build from source on Ubuntu 18.04 or not. It should be mentioned, that users already tried to build Ganeti-3.0 on Ubuntu 18.04 (see #1452).

### DIFF
--- a/src/Ganeti/Compat.hs
+++ b/src/Ganeti/Compat.hs
@@ -41,6 +41,7 @@ module Ganeti.Compat
   ( filePath'
   , maybeFilePath'
   , toInotifyPath
+  , getPid'
   ) where
 
 import qualified Data.ByteString.UTF8 as UTF8
@@ -49,6 +50,13 @@ import System.Posix.ByteString.FilePath (RawFilePath)
 import qualified System.INotify
 import qualified Text.JSON
 import qualified Control.Monad.Fail as Fail
+import System.Process.Internals
+import System.Posix.Types (CPid (..))
+#if MIN_VERSION_process(1,6,3)
+import System.Process (getPid)
+#else
+import Control.Concurrent.Lifted (readMVar)
+#endif
 
 -- | Wrappers converting ByteString filepaths to Strings and vice versa
 --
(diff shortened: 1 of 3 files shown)


src/Ganeti/Query/Exec.hs
```diff
@@ -88,6 +96,21 @@ rethrowAnnotateIOError :: String -> IO a -> IO a
 rethrowAnnotateIOError desc =
   modifyIOError (\e -> annotateIOError e desc Nothing Nothing)
 
+-- | Process 1.6.3. introduced the getPid function, for older versions
+-- provide an implemention here (https://github.com/haskell/process/pull/109)
+
+getPid' :: ProcessHandle -> IO (Maybe Pid)
+#if MIN_VERSION_process(1,6,3)
```
apoikos, line 103: I am a bit reluctant to add this to 3.0, but then again I don't know how many people will be building from source on 18.04. Let's say we go for it, then it should go into Ganeti.Compat instead, to keep the code in Ganeti.Query.Exec cleaner.

saschalucas: > Let's say we go for it, then it should go into Ganeti.Compat instead, to keep the code in Ganeti.Query.Exec cleaner.

I've tried to move the function definition into Ganeti.Compat. I hope it's not that far off.

saschalucas: > I am a bit reluctant to add this to 3.0, but then again I don't know how many people will be building from source on 18.04.

I don't know either. But this change seems small enough, to not support it. Do you see possible future troubles, that may arise from depending on ghc-8.0.2 (the bionic version)? Or could other old software break functionality, like qemu-2.11 (i.e. WRT # #1433)? Having a qa-suite for 18.04 would be fine.

src/Ganeti/Compat.hs
```diff
@@ -49,6 +50,12 @@ import System.Posix.ByteString.FilePath (RawFilePath)
 import qualified System.INotify
 import qualified Text.JSON
 import qualified Control.Monad.Fail as Fail
+import System.Process.Internals
+import System.Posix.Types (CPid (..))
+import Control.Concurrent.Lifted (readMVar)
```
apoikos, line 55: Technically this could be moved inside the `#if` clause.

PR #1479, opened 2020-06-10
Relax QuickCheck version restriction

mbakke:
Hi!  Long time user, first time contributor.  All tests pass with QuickCheck 2.13.2 which happens to be current on my distro, so bump the version range accordingly.

### DIFF
--- a/cabal/ganeti.template.cabal
+++ b/cabal/ganeti.template.cabal
@@ -85,7 +85,7 @@ library
   if flag(htest)
     build-depends:
         HUnit                         >= 1.2.4.2    && < 1.7
-      , QuickCheck                    >= 2.8        && < 2.13
+      , QuickCheck                    >= 2.8        && < 2.14
       , test-framework                >= 0.6        && < 0.9
       , test-framework-hunit          >= 0.2.7      && < 0.4
       , test-framework-quickcheck2    >= 0.2.12.1   && < 0.4


rbott:
I'll do some more testing on this and update this PR once finished. Just curious: what distro/version are you using?

mbakke:
@rbott I'm packaging Ganeti for GNU Guix, a rolling distribution; with the intent of getting rid of my remaining Debian servers :-)

There will likely be more PRs once I get around to actually deploying the thing.

iustin:
I've taken a quick look at https://github.com/nick8325/quickcheck/blob/master/changelog, and 2.13→2.14 delta is small, functionality-wise; if the code compiles and tests run, this is a no-brainer. I would LGTM this, but if @rbott already started, I'll let him approve.

saschalucas:
> I've taken a quick look at https://github.com/nick8325/quickcheck/blob/master/changelog, and 2.13→2.14 delta is small ...

As far as I understand this version syntax, this PR wants to set upper boundary to <2.14, which I understand as <=2.13? @iustin: do you suggest to expand it to <=2.14? Or should we better consult Changelog from 2.12→2.13?

iustin:
@saschalucas - ah, you are right, my bad. Changelog for 2.12→2.13 is a bit more complex, but nothing too much either. I think it's also fine.

Note: both 2.13 and 2.14 seem to bring quite a few good improvements…

PR #1489, opened 2020-06-26
fix various spelling/typos

saschalucas:
Just a quick effort by using a spell checker, trying to avoid differences in US/GB spelling.

### DIFF
--- a/man/gnt-instance.rst
+++ b/man/gnt-instance.rst
@@ -349,7 +349,7 @@ scsi\_controller\_type
 kvm\_pci\_reservations
     Valid for the KVM hypervisor.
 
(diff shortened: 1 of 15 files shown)


man/gnt-instance.rst
```diff
@@ -621,7 +621,7 @@ serial\_console
 
     Enabling serial console emulation also appends
     ``"console=ttyS0,<serial_speed>"`` to the end of ``kernel_args`` in KVM and
-    may infere with previous settings.
+    may infer with previous settings.
```
iustin, line 624: I think this was rather "interfere"; infer here doesn't make much sense.

PR #1496, opened 2020-07-02
Fixes for DRBD >= 8.4.

mbakke:
This is a cleaned-up version of #1296, taking the comments from @apoikos into account. I have tested this with drbd-utils 9.13.1 and kernel module version 8.4.11 (Linux 5.7.6).

More testing very welcome!

Cc @vladimir-ipatov

### DIFF
--- a/lib/storage/drbd.py
+++ b/lib/storage/drbd.py
@@ -315,6 +315,13 @@ class DRBD8Dev(base.BlockDev):
     """
     return self._show_info_cls.GetDevInfo(self._GetShowData(minor))
 
+  @staticmethod
+  def _NeedsLocalSyncerParams():
+    # For DRBD >= 8.4, syncer init must be done after local, not in net.
+    info = DRBD8.GetProcInfo()
+    version = info.GetVersion()
+    return version["k_minor"] >= 4
+
   def _MatchesLocal(self, info):
     """Test if our local config matches with an existing device.
 
@@ -397,6 +404,20 @@ class DRBD8Dev(base.BlockDev):
         base.ThrowError("drbd%d: can't attach local disk: %s",
                         minor, result.output)
 
+    def _WaitForMinorSyncParams():
+      """Call _SetMinorSyncParams and raise RetryAgain on errors.
+      """
+      if self._SetMinorSyncParams(minor, self.params):
+        raise utils.RetryAgain()
+
+    if self._NeedsLocalSyncerParams():
+      # Retry because disk config for DRBD resource may be still uninitialized.
+      try:
+        utils.Retry(_WaitForMinorSyncParams, 1.0, 5.0)
+      except utils.RetryTimeout as e:
+        base.ThrowError("drbd%d: can't set the synchronization parameters: %s" %
+                        (minor, utils.CommaJoin(e.args[0])))
+
   def _AssembleNet(self, minor, net_info, dual_pri=False, hmac=None,
                    secret=None):
     """Configure the network part of the device.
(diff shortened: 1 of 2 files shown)


lib/storage/drbd.py
```diff
@@ -397,6 +404,22 @@ def _AssembleLocal(self, minor, backend, meta, size):
...
+      else:
+        return []
+
+    if self._NeedsLocalSyncerParams():
+      # Retry because disk config for DRBD resource may be still uninitialized.
+      try:
+        utils.Retry(_WaitForMinorSyncParams, 1.0, 5.0)
+      except utils.RetryTimeout:
+        base.ThrowError("drbd%d: can't set the synchronization parameters: %s" %
+                        (minor, utils.CommaJoin(sync_errors)))
```
mbakke, line 421: `sync_errors` here is unbound.  Not sure how to propagate the error from `_WaitForMinorSyncParams` when `RetryTimeout` is reached.  Ideas?

rbott: I am not exactly sure I understood how utils.retry works but the following should preserve the sync_errors variable through the exception:
```python
  for cmd in cmds:
    result = utils.RunCmd(cmd)
    if result.failed:
      base.ThrowError("drbd%d: can't attach local disk: %s",
                      minor, result.output)

  def _WaitForMinorSyncParams():
    """Call _SetMinorSyncParams and raise RetryAgain on errors.
    """
    sync_errors = self._SetMinorSyncParams(minor, self.params)
    if sync_errors:
      raise utils.RetryAgain(sync_errors)
    else:
      return []
      
  if self._NeedsLocalSyncerParams():
    # Retry because disk config for DRBD resource may be still uninitialized.
    try:
      utils.Retry(_WaitForMinorSyncParams, 1.0, 5.0)
    except utils.RetryTimeout as e:
      base.ThrowError("drbd%d: timed out setting the synchronization parameters: %s" %
                      (minor, utils.CommaJoin(e.args[0])))
```

rbott:
I have successfully tested this with a full DRBD cluster QA run on [Debian Buster](https://salsa.debian.org/rbott/ganeti-build/-/jobs/843858) (DRBD 8.4.10, drbd-utils 9.5.0) and [Ubuntu Focal Fossa](https://salsa.debian.org/rbott/ganeti-build/-/jobs/844001) (DRBD 8.4.11, drbd-utils 9.11.0). I will also run this through Debian Bullseye tomorrow.

However, the error handling the in the retry code seems wrong, but that is already an open discussion/code comment:
https://github.com/ganeti/ganeti/blob/b5663b4f54dd768c1e5f539a5c164ff7088137bc/lib/storage/drbd.py#L407-L421

We need to sort this out but other than that, LGTM.

lib/storage/drbd.py
mbakke: Thanks for the hint, I've incorporated this suggestion now.  I tried testing it by setting a very low timeout, but could not trigger the error.  Probably my toy cluster is not busy enough!

I also removed the explicit return from `_WaitForMinorSyncParams` as nothing is checking it.

mbakke:
@rbott thanks a lot for reviewing, I think the error handling issue has been resolved with the latest revision.

lib/storage/drbd.py
```diff
@@ -315,6 +315,13 @@ def _GetShowInfo(self, minor):
     """
     return self._show_info_cls.GetDevInfo(self._GetShowData(minor))
 
+  @staticmethod
+  def _NeedsLocalSyncerParams():
+    # For DRBD >= 8.4, syncer init must be done after local, not in net.
+    info = DRBD8.GetProcInfo()
+    version = info.GetVersion()
+    return version["k_minor"] >= 4
```
apoikos, line 323: Additionally checking that `version["k_major"] == 8` would make the code more robust.

vladimir-ipatov:
> Thanks to both, @vladimir-ipatov and @mbakke!
> 
> In principle it looks good, however I'd like to link this PR to specific issues. Given that DRBD 8.4 generally works without this PR, what is it exactly that we are fixing here? Neither this PR, nor #1296 point to specific issues, and we need that rationale at least for the changelog. I'm not saying these changes are unnecessary, I just want a documented justification for them :)

Hello!

Main issue is:
Without this fixes ganeti with drbd 8.4 will not assemble drbd device if one of two nodes don't have lvm volumes for it (diskless).
According to my information, there is no issues on github about it.

It was mentioned in my pull request description (but may be uncertainly :) )"
"fixes for drbd >= 8.4: add drbdsetup new-device and new-mirror for network part assembling of drbd (needed for diskless states of drbd)"


May be there is some other minor issues, I don't remember.

apoikos:
Merged, thanks to both of you!

PR #1499, opened 2020-07-10
Cabal 3.0 compatibility

### LINKED ISSUE #1492: Build broken with Cabal 3
Cabal 3 changed its [APIs](https://github.com/haskell/cabal/pull/6022/files) as well as the [default command behavior](https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html), breaking the build. I've already fixed this, expect a PR soon.

apoikos:
This PR adds compatibility with Cabal 3.0 by adapting to Cabal API and command-line changes. You can see #1492 for details and links, but it basically boils down to:

 - inlining  `Distribution.Simple.LocalBuildInfo.externalPackageDeps` which was removed upstream
 - detecting and using the `v1-configure` command rather than plain `configure` which now does a NIX-style build
 - handling some renamed functions.

Closes #1492.

### DIFF
--- a/cabal/CabalDependenciesMacros.hs
+++ b/cabal/CabalDependenciesMacros.hs
@@ -5,7 +5,14 @@ import Control.Applicative
 import qualified Data.Set as Set
 import qualified Distribution.Simple.Build.Macros as Macros
 import Distribution.Simple.Configure (getPersistBuildConfig)
+#if MIN_VERSION_Cabal(3,0,0)
+import Distribution.Types.UnitId (UnitId)
+import Distribution.Types.MungedPackageId (MungedPackageId)
+import Distribution.Types.ComponentLocalBuildInfo (componentPackageDeps, componentUnitId)
+import Data.List (nub)
+#else
 import Distribution.Simple.LocalBuildInfo (externalPackageDeps)
+#endif
 import Distribution.PackageDescription (packageDescription)
 
 -- Common Cabal 2.x dependencies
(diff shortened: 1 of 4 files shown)


cabal/CabalDependenciesMacros.hs
```diff
@@ -30,6 +37,26 @@ import System.Environment (getArgs)
...
+#else
+generate = Macros.generate
+#endif
+
+#if MIN_VERSION_Cabal(3,0,0)
+externalPackageDeps :: LocalBuildInfo.LocalBuildInfo -> [(UnitId, MungedPackageId)]
+externalPackageDeps lbi =
+    -- TODO:  what about non-buildable components?
+    nub [ (ipkgid, pkgid)
+        | clbi            <- Graph.toList (LocalBuildInfo.componentGraph lbi)
```
iustin, line 51: This `Graph.toList` call is duplicated below. Please move it out to a local binding, as it will improve readability.

In my own version, I've used simply the `componentNameMap`, rather than going through the graph (we don't care about the structure here), but it's the same so good as it is.

cabal/ganeti.template.cabal
```diff
@@ -1,5 +1,5 @@
 name:                ganeti
-version:             2.16
+version:             3.0
```
iustin, line 2: Not that it matters, but I'm not sure calling it 3.0 already is best :) I would have chosen something like 2.99 for now.

PR #1503, opened 2020-07-16
Expose disk size to OS install scripts.

mbakke:
Hi,

This is a minor change to the OS API which allows OS providers to query the configured disk size.  It is useful for providers such as [instance-guix](https://github.com/mbakke/ganeti-instance-guix) which creates a disk image outside of Ganeti (no chroot etc) and copies it into place.  In particular, it is needed to reinstall file disks, but also to work with different storage providers without needing intimate knowledge of how to query the size.

### DIFF
--- a/lib/backend.py
+++ b/lib/backend.py
@@ -4305,6 +4305,7 @@ def OSEnvironment(instance, inst_os, debug=0):
     uri = _CalculateDeviceURI(instance, disk, real_disk)
     result["DISK_%d_ACCESS" % idx] = disk.mode
     result["DISK_%d_UUID" % idx] = disk.uuid
+    result["DISK_%d_SIZE" % idx] = str(disk.size)
     if real_disk.dev_path:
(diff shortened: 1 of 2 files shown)


iustin:
While the OS API version itself shouldn't change, I wonder if at least we shouldn't document from which version such a new variable is present? The way you documented it seems to imply it's been available forever.

Or, we keep a base set that was always available and a set of variables that appeared over time, all annotated with the version.

mbakke:
@iustin that makes sense.  I don't really have a strong opinion and opted for the first suggestion (annotate with the version it was added).

saschalucas:
Thanks for this PR. Too bad I didn't think of it myself. Every time I see/use `blockdev` or `qemu-img info`. This will be obsolete now. Good point by @iustin to document the version in which the variable was added. From my point it's fine, but I'll wait for @iustin to finish this PR.

PR #1504, opened 2020-07-16
Make gnt-cluster.8 manual page generation deterministic.

mbakke:
This PR sorts the ecode entries in `gnt-cluster(8)` by name.  In addition to being more human-friendly, this one-line change incidentally makes the Ganeti build process [reproducible](https://reproducible-builds.org/).


mbakke:
@saschalucas I believe the source is the `for (_, name, doc) in constants.CV_ALL_ECODES` loop just above this change.  Another fix would be to set `$PYTHONHASHSEED` to a fixed value, I suppose that's why Debian has not found it.

Issue #1493, opened 2020-06-30
Build broken with GHC 8.8


apoikos:
Base 4.13.0 completely removed `fail` from the `Monad` class (see [the migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/8.8)) and Ganeti currently does not build against GHC 8.8. This should be straightforward to address, expect a PR soon.

iustin:
I missed this bug, and I started on this. I (also) have now Ganeti working with GHC 8.8, the question is just what backwards compat we want before making the patch ready…  As keeping backwards compat with before 8.0 will be very painful, and even 8.0 might be ugly enough.

apoikos:
> I missed this bug, and I started on this. I (also) have now Ganeti working with GHC 8.8, the question is just what backwards compat we want before making the patch ready… As keeping backwards compat with before 8.0 will be very painful, and even 8.0 might be ugly enough.

It's okay, I didn't get a chance to do much. We already support only 8.0+ in ganeti master, so you can count on `MonadFail` being available.

iustin:
OK, docs are unclear, as INSTALL says GHC 7+ :)

The fun part is that after fixing the MonadFail part, there are a few actual other fixes needed that make me wonder how come the code was not issuing warnings before.

Question: why not bump GHC to 8.4? Do we care about oldstable/stretch, or some other distro that doesn't have 8.4?

iustin:
Ansgar made a good point on ganeti-devel that 18.04 LTS compatibility should be kept in mind. That ships 8.0.2, so it does look like 8.0 compat should be kept :)

iustin:
Made progress - it works with 8.8, but now I have to make it work well with older versions too :) so needs a bit more work.

iustin:
Sigh. Now only 8.4 is broken, in tests. Something about the MonadPlus instance not working correctly, but works in 8.0 and 8.8.

I'll give it a rest for a couple of days, tired :)

apoikos:
I think we can close this now that GHC 8.8 support has been released and tested. Thanks @iustin!

PR #1507, opened 2020-07-19
GHC 8.8 + Debian unstable compatibility

iustin:
This is an proposal for making master build with GHC 8.8 and Debian
unstable (not fully up-to-date as of 19th July, but almost), hopefully
closing issue #1493.

Specific notes:

- the simplest way to enable network-bsd but also old style was to add
  a new flag. I've tested building on sid and buster, but not older.
- test method: `./autogen.sh && ./configure --enable-haskell-tests`
  followed by `make` and `make hs-tests`
- tested with Ubuntu Bionic (ghc 8.0.2), Debian Buster (ghc 8.4.4) and
  Debian Sid (ghc 8.8.3)

I'm not entirely sure on the changes, but as tests pass, they should
be just fine, right? :)


iustin:
I've just looked, and it seems that the builds are only done on buster. If this goes in, we should probably expand the build matrix to all the above listed OS versions, so that we don't regress.

src/Ganeti/Objects/Lens.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE LiberalTypeSynonyms #-}
```
apoikos, line 2: Out of curiosity, why is this necessary?

mbakke:
Perhaps it's time to push a beta2 with the Haskell and DRBD fixes?

apoikos:
@mbakke, yes, we're planning to do so, thanks!

PR #1513, opened 2020-08-27
CI: use docker images from GitHub Packages

apoikos:
Our CI images now live on GitHub (https://github.com/ganeti/ci-images) and are automatically published to
GitHub Packages. Change the CI workflow to use these images instead of
the ones on DockerHub.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [bionic, buster, focal, debian-testing]
     container:
-      image: ganeti/ci:${{ matrix.os }}-py3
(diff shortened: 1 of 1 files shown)


apoikos:
Apparently the Github Packages registry requires authentication and there doesn't seem to be an easy way of doing this through Actions. Needs a bit more investigation.

apoikos:
Abandoning this, as there doesn't seem to be any "real" solution (other than ditching `container` and doing the whole docker dance by hand). See https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358 for reference.

apoikos:
I'll change the ci-images workflow to publish to dockerhub for the time being, until GitHub figures out what it is they want people to be able to do with the GHP registry.

PR #1514, opened 2020-08-31
Enable codeql security analysis

iustin:
I just got the notification that we're in the beta, so let's enable this. Only works for the Python part (and cpp/go/java/js/csharp, which we don't use), but better than nothing.


apoikos (approved):
Looks good, thanks! Not sure we need to do something about the missing base commit analysis, I guess it will update itself as soon as we merge this to master.

iustin:
> Looks good, thanks! Not sure we need to do something about the missing base commit analysis, I guess it will update itself as soon as we merge this to master.

It seems so, this is what it says:

> Missing analysis for base commit 0af56c7
> This usually happens when you enable code scanning for the first time, or if the build failed. […]

PR #1516, opened 2020-09-01
Do not override disk_cache settings for RBD disks

### LINKED ISSUE #822: Don't override disk_cache settings for RBD disks
Originally reported of Google Code with ID 766.

```
What software version are you running? Please provide the output of "gnt-
cluster --version", "gnt-cluster version", and "hspace --version".

gnt-cluster (ganeti v2.10.1) 2.10.1

Software version: 2.10.1
Internode protocol: 2100000
Configuration format: 2100000
OS api version: 20
Export interface: 0
VCS version: (ganeti) version v2.10.1

hspace (ganeti) version v2.10.1
compiled with ghc 7.6
running on linux x86_64

<b>What distribution are you using?</b>

Debian Jessie (testing)

<b>What steps will reproduce the problem?</b>
1. disk_cache settings other than none will be overridden by ganeti despite RBD being completely safe and well behaved when it comes to live migration and other cache sensitive issues.

2. See https://groups.google.com/forum/#!topic/ganeti/loF_Gp6hHew and especially the reference to https://www.mail-archive.com/ceph-users@lists.ceph.com/msg06890.html in that thread.

<b>What is the expected output? What do you see instead?</b>

ganeti should treat RBD like DRBD when it comes to cache safety. 

<b>Please provide any additional information below.</b>

Since the check that does the override lumps all externally mirrored disks into the same unsafe pot (and some of those probably are unsafe with caching) I don't know what the best way to solve this from a programming PoV is.
If making an exception for RBD is too involved maybe changing the override to a stern warning "you better know what you're doing" might be the faster way forward. 
Fixing this ASAP would vastly improve performance with RBD, especially with RBD Cache in the userspace model. 
```
Originally added on 2014-03-20 08:22:32 +0000 UTC.

oleeander:
Currently disk_cache settings other than none will be overridden to
none. RBD is considered safe and well behaved when it comes to
live migration and other cache sensitive issues.

Fixes #822.

Signed-off-by: Oleander Reis <oleander.reis@hostserver.de>

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1173,7 +1173,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     # Cache mode
     disk_cache = up_hvp[constants.HV_DISK_CACHE]
     for cfdev, link_name, uri in kvm_disks:
-      if cfdev.dev_type in constants.DTS_EXT_MIRROR:
+      if (cfdev.dev_type in constants.DTS_EXT_MIRROR
+          and cfdev.dev_type != constants.DT_RBD):
         if disk_cache != "none":
           # TODO: make this a hard error, instead of a silent overwrite
           logging.warning("KVM: overriding disk_cache setting '%s' with 'none'"


atta (approved):
Thanks for your Patch,

we have the same change in our live-clusters for quite some time and i'll be more than happy to merge it upstream.

PR #1517, opened 2020-09-03
Document gnt-instance add --tags

dannyman:
I have been using the `gnt-instance add --tags` feature for years. I recently discovered that the argument is not documented in `gnt-instance(1)` so I thought I might chip in. Cheers!


saschalucas (approved):
Thanks a lot. I use it too, but didn't ever noticed, that it is not documented.

PR #1518, opened 2020-09-07
don't expose RAPI to public w/o password

saschalucas:
This binds the RAPI port to localhost and requires always authentication, so that no information can be leaked by default. This is what [Debian packaging does](https://salsa.debian.org/ganeti-team/ganeti/-/blob/master/debian/rules#L111-112).

Signed-off-by: Sascha Lucas <sascha_lucas@web.de>

### DIFF
--- a/doc/examples/ganeti.default
+++ b/doc/examples/ganeti.default
@@ -1,6 +1,6 @@
 # Default arguments for Ganeti daemons
 NODED_ARGS=""
-RAPI_ARGS=""
+RAPI_ARGS="-b 127.0.0.1 --require-authentication"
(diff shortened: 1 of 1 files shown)


iustin (approved):
I'm not 100% onboard with this, in the sense that I (personally) would expect on the Ganeti nodes to not have random users logged in; so to me, either `-b 127.0.0.1` _or_ `--require-authentication` would be a better choice, but I'm not against it so to say.

One thing though - does `-b 127.0.0.1` means RAPI by defauly will only be reachable on ipv4 localhost?

In any case, this is examples, so LGTM.

PR #1521, opened 2020-09-13
htest: use printable unicode ByteStrings

### LINKED ISSUE #1506: Flaky config serialization test
As seen in https://github.com/ganeti/ganeti/runs/887095914?check_suite_focus=true:

```
2020-07-19T14:49:36.1571390Z ConfigData {configVersion = -9, configCluster = Cluster {clusterRsahostkeypub = "KD\479716", clusterDsahostkeypub = Just
2020-07-19T14:49:36.1577720Z Expected equality, but got mismatch
2020-07-19T14:49:36.1599464Z expected: ConfigData {configVersion = -9, configCluster = Cluster {clusterRsahostkeypub = "KD\479716", clusterDsahostkey
2020-07-19T14:49:47.9545941Z  but got: ConfigData {configVersion = -9, configCluster = Cluster {clusterRsahostkeypub = "KD\479716", clusterDsahostkey
2020-07-19T14:49:47.9552588Z (used seed -3390799639642185379)
```

Full output is too large to inline, attaching.
[8_Haskell tests.txt](https://github.com/ganeti/ganeti/files/4943900/8_Haskell.tests.txt)

apoikos:
For most of the config values, our test suite uses the generic String generator and decodes its output to a ByteString assuming UTF-8 encoding. The String arbitrary instance generates random unicode codepoints, avoiding only surrogate pairs. Unfortunately, as demonstrated in #1506, it also generates "non-characters"[1] that are usually rejected when found in a UTF-8 string, such as U+FFFE:

```Haskell
GHCi, version 8.8.3: https://www.haskell.org/ghc/  :? for help
Prelude> import qualified Data.ByteString.UTF8 as UTF8
Prelude UTF8> UTF8.toString $ UTF8.fromString "\65534"
"\65533"
```

Note how U+FFFE was replaced by the replacement character U+FFFD (�). This behavior breaks the serialization/deserialization tests, causing intermitent build failures.

Our expectations about config objects are that they should be serializable into JSON, and thus should contain valid Unicode text. We take extra care to encode arbitrary binary data, such as the gnt-network bitarrays, exactly because our config file is not a binary data store.

To this end, I believe it makes sense to restrict ByteString generation to printable Unicode strings. Do this using the modifiers introduced in QuickCheck 2.10.

This fixes #1506.

[1] https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Non-characters

### DIFF
--- a/test/hs/Test/Ganeti/Objects.hs
+++ b/test/hs/Test/Ganeti/Objects.hs
@@ -91,7 +91,7 @@ instance Arbitrary (Container DataCollectorConfig) where
       fromContainer = Map.fromList $ zip names configs }
 
 instance Arbitrary BS.ByteString where
-  arbitrary = fmap UTF8.fromString arbitrary
+  arbitrary = genPrintableByteString
 
 $(genArbitrary ''PartialNDParams)
 
--- a/test/hs/Test/Ganeti/TestCommon.hs
+++ b/test/hs/Test/Ganeti/TestCommon.hs
@@ -56,6 +56,7 @@ module Test.Ganeti.TestCommon
   , genPrintableAsciiChar
   , genPrintableAsciiString
   , genPrintableAsciiStringNE
+  , genPrintableByteString
   , genName
   , genFQDN
   , genUUID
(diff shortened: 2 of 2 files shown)


apoikos:
Oh well, I guess bionic only has QuickCheck 2.8. I'll add some compatibility code.

iustin (approved):
Thanks!

PR #1522, opened 2020-09-13
Fix docs by removing sphinx pre-1.3 support

saschalucas:
Currently building documentation is broken (missing style sheet), because `doc/css/style.css` includes default.css:

https://github.com/ganeti/ganeti/blob/f3f033a62c15f63cbc091be9924a2a370c874a39/doc/css/style.css#L1

instead of classic.css, which is the sphinx theme for >=1.3. The commit 268953c4c introduced differentiation between sphinx pre-1.3 and later, but missed `doc/css/style.css`.

This removes the differentiation between sphinx pre-1.3 and later, because the oldest shinx version currently in use is from Ubuntu Xenial (1.3)[1]. This means, switching to classic theme only. To not further hardcode the name of the theme (include statement), remove also the custom style (underlining links):

https://github.com/ganeti/ganeti/blob/f3f033a62c15f63cbc091be9924a2a370c874a39/doc/css/style.css#L3-L4

Mostly a revert of commit 268953c4c and dbee5c92.

[1] see footer of https://saschalucas.github.io/docs/docs/ganeti/2.16/html/search.html

This PR is against stable-2.16 in order to build docs there too. Later it will be merged into master.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -646,7 +646,6 @@ docinput = \
 	doc/cluster-keys-replacement.rst \
 	doc/cluster-merge.rst \
 	doc/conf.py \
-	doc/css/style.css \
 	doc/design-2.0.rst \
 	doc/design-2.1.rst \
 	doc/design-2.2.rst \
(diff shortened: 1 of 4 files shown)


apoikos (approved):
LGTM, thanks! It's always good to clean some cruft :)

PR #1524, opened 2020-09-18
Release 3.0.0~rc1

apoikos:
Wrap some overlong lines reported by `make check-local` (we should really check overlong lines at PR time), bump the version and set a release date.


saschalucas (approved):
Thanks! Please go ahead an make the RC1. Regarding `check-local` there is #1470 from @rbott. Maybe it's worth to look at it again.

PR #1529, opened 2020-10-12
automatic postcopy-ram live migration

saschalucas:
This implements automatic postcopy-ram live migration for the KVM hypervisor. Basically it is a cherry-pick from commit 041ed56d9 adopted to today's code. Credit goes to the original author @calumcalder.

It extends the migration status by `dirty_sync_count`, which is the number of times that dirty ram was synchronized (since qemu-2.1). If the migration capability `postcopy-ram` is set and `dirty_sync_count >=2`, live migration automatically switches to postcopy-ram migration.  In the future the existing HV parameter `migration_mode` can be extended by something like postcopy-live to let users better control live migrations.

Special Thanks to @cilq for his help on this PR.

### DIFF
--- a/lib/cmdlib/instance_migration.py
+++ b/lib/cmdlib/instance_migration.py
@@ -953,7 +953,10 @@ class TLMigrateInstance(Tasklet):
         raise errors.OpExecError("Could not migrate instance %s: %s" %
                                  (self.instance.name, msg))
 
-      if result.payload.status != constants.HV_MIGRATION_ACTIVE:
+      if ms.postcopy_status == constants.HV_KVM_MIGRATION_POSTCOPY_ACTIVE:
+        self.feedback_fn("* memory transfer has switched to postcopy")
+
+      if ms.status not in constants.HV_KVM_MIGRATION_ACTIVE_STATUSES:
         self.feedback_fn("* memory transfer complete")
         break
 
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -162,6 +162,8 @@ _SCSI_BUS = "scsi.0"
 
 _MIGRATION_CAPS_DELIM = ":"
 
+# in future make dirty_sync_count configurable
+_POSTCOPY_SYNC_COUNT_THRESHOLD = 2 # Precopy passes before enabling postcopy
 
 def _with_qmp(fn):
   """Wrapper used on hotplug related methods"""
@@ -2633,6 +2635,25 @@ class KVMHypervisor(hv_base.BaseHypervisor):
             migration_status.transferred_ram = \
                 query_migrate["ram"]["transferred"]
             migration_status.total_ram = query_migrate["ram"]["total"]
+
+            migration_status.postcopy_status = None
+            migration_caps = instance.hvparams[constants.HV_KVM_MIGRATION_CAPS]
+            # migration_caps is a ':' delimited string, so checking
+            # if 'postcopy-ram' is a substring also covers using
+            # x-postcopy-ram for QEMU 2.5
+            if migration_caps and "postcopy-ram" in migration_caps:
+              dirty_sync_count = query_migrate["ram"]["dirty-sync-count"]
+              if (migration_status.status == constants.HV_MIGRATION_ACTIVE
+                  and dirty_sync_count >= _POSTCOPY_SYNC_COUNT_THRESHOLD):
+                self.qmp.StartPostcopyMigration()
+                logging.info("switched live migration for instance %s to"
+                             " postcopy mode", instance.name)
+                # qmp.StartPostcopyMigration() seems asynchronous. Doing
+                # qmp.GetMigrationStatus() right after, the ["status"] won't be
+                # HV_KVM_MIGRATION_POSTCOPY_ACTIVE. This is why an extra field
+                # ["postcopy_status"] is introduced
+                migration_status.postcopy_status = \
+                    constants.HV_KVM_MIGRATION_POSTCOPY_ACTIVE
 
           return migration_status
         else:
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -710,6 +710,8 @@ class QmpConnection(MonitorSocket):
 
     arguments = {
       "max-bandwidth": max_bandwidth,
+      # TODO: available since qemu 3.0
+      #"max-postcopy-bandwidth": max_bandwidth,
       "downtime-limit": downtime_limit,
     }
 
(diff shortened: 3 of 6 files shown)


saschalucas:
fiendly ping @apoikos @rbott @iustin for review.

iustin:
Sorry, had not time for anything outside of work recently. Will try to take a look during the weekend.

lib/cmdlib/instance_migration.py
```diff
@@ -953,7 +964,14 @@ def _ExecMigration(self):
         raise errors.OpExecError("Could not migrate instance %s: %s" %
                                  (self.instance.name, msg))
 
-      if result.payload.status != constants.HV_MIGRATION_ACTIVE:
+      if (postcopy_enabled
```
apoikos, line 967: Continuing from my previous comment, I feel we have moved decision-making to the wrong level. I understand that this is in part because we don't monitor migration progress at the backend level, we only poll it regularly from the master via the RPC and so it feels natural to have the master take the decision instead of the node.

That said, I feel your approach can be improved to avoid bringing low-level details such as the memory pass count in cmdlib. I see two ways out of this:
 - Move the whole low-level decision logic in KVM's `GetMigrationStatus` and signal that an instance should move to postcopy mode via `MigrationStatus` (either through a special status, or through another field instead of `dirty_sync_count`).
OR
 - Trigger the switchover to postcopy from inside `GetMigrationStatus` itself. This is a bit ugly because it introduces side-effects to a "Get" function, but it has a couple of advantages:
   - It saves some trivial decision-making on the master
   - It does not require a new RPC call
   - It closes the gap between the time we get the migration status and the time we issue the postcopy start

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2590,6 +2590,21 @@ def MigrateInstance(self, cluster_name, instance, target, live_migration):
...
+    Requires that an instance is currently migrating, and that the
+    postcopy-ram (x-postcopy-ram on QEMU version 2.5 and below)
+    migration capability is enabled in the instance's hypervisor
+    parameters.
+
+    @type instance: L{objects.Instance}
+    @param instance: The instance being migrated.
+
+    """
+    self.qmp.StartPostcopyMigration()
```
apoikos, line 2606: Have we checked if this fails when issued against a completed migration? This is important because some time will have passed from the moment we issued the `migration_info` command and regular migration may have succeeded.

apoikos (changes requested):
Thanks for the PR @saschalucas and apologies for the long delay! Overall it's a nice feature, but I feel it can be improved by moving low-level decisions to the node level. Please see the inline comments.

saschalucas:
@apoikos, thanks for reviewing. I have tried to resolve all your points:

* make `_POSTCOPY_SYNC_COUNT_THRESHOLD` configurable: I assume the best is not in this PR but maybe in 3.1 to not change config data structure
* I moved decision making and  trigger to postcopy into `GetMigrationStatus`, saving extra RPC calls. Signal through another field instead of `dirty_sync_count`
* check what happens if we sent `migrate-start-postcopy` to an already finished migration: seems save from my point of view, the source instance seems no longer connected to the destination, after migartion has completed.
```json
{"QMP": {"version": {"qemu": {"micro": 0, "minor": 1, "major": 3}, "package": "Debian 1:3.1+dfsg-8+deb10u8"}, "capabilities": []}}

{ "execute": "qmp_capabilities" }
{"return": {}}

{ "execute": "query-migrate" }
{"return": {"status": "completed", "setup-time": 6, "downtime": 25, "total-time": 6750, "ram": {"total": 1091379200, "postcopy-requests": 0, "dirty-sync-count": 4, "multifd-bytes": 0, "page-size": 4096, "remaining": 0, "mbps": 270.694544, "transferred": 227928856, "duplicate": 212511, "dirty-pages-rate": 0, "skipped": 0, "normal-bytes": 225574912, "normal": 55072}}}

{ "execute": "migrate-set-capabilities", "arguments": { "capabilities": [ { "capability": "postcopy-ram", "state": true } ] } }
{"return": {}}

{ "execute": "migrate-start-postcopy" }
{"return": {}}

{ "execute": "query-migrate" }
{"return": {"status": "completed", "setup-time": 6, "downtime": 25, "total-time": 6750, "ram": {"total": 1091379200, "postcopy-requests": 0, "dirty-sync-count": 4, "multifd-bytes": 0, "page-size": 4096, "remaining": 0, "mbps": 270.694544, "transferred": 227928856, "duplicate": 212511, "dirty-pages-rate": 0, "skipped": 0, "normal-bytes": 225574912, "normal": 55072}}}
```

lib/cmdlib/instance_migration.py
```diff
@@ -953,7 +953,10 @@ def _ExecMigration(self):
         raise errors.OpExecError("Could not migrate instance %s: %s" %
                                  (self.instance.name, msg))
 
-      if result.payload.status != constants.HV_MIGRATION_ACTIVE:
+      if ms.postcopy_status == "switched":
```
apoikos, line 956: `"switched"` should be a constant, as it is also used in hv_kvm. You could even use the already-defined `HV_KVM_MIGRATION_POSTCOPY_ACTIVE` constant, since the value of the `postcopy_status` field is opaque.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2640,6 +2642,18 @@ def GetMigrationStatus(self, instance):
...
+            migration_status.postcopy_status = None
+            migration_caps = instance.hvparams[constants.HV_KVM_MIGRATION_CAPS]
+            # migration_caps is a ':' delimited string, so checking
+            # if 'postcopy-ram' is a substring also covers using
+            # x-postcopy-ram for QEMU 2.5
+            if migration_caps and "postcopy-ram" in migration_caps:
+              dirty_sync_count = query_migrate["ram"]["dirty-sync-count"]
+              if (migration_status.status == constants.HV_MIGRATION_ACTIVE
+                  and dirty_sync_count >= _POSTCOPY_SYNC_COUNT_THRESHOLD):
+                self.qmp.StartPostcopyMigration()
```
apoikos, line 2654: Since this call has a side-effect, it would be good to log something along the lines of "switching migration to post-copy mode' here, so that it also appears in the node daemon logs.

apoikos (commented):
Overall much cleaner than the initial version, thanks! I'm not approving yet, since I'd like to see both minor issues addressed, but other than that it looks quite nice and clean :)

saschalucas:
Thanks @apoikos. I have:

* replaced "switched" by HV_KVM_MIGRATION_POSTCOPY_ACTIVE
* inserted a log after switching to postcopy mode

apoikos (approved):
Thanks a lot @saschalucas!

PR #1534, opened 2020-12-11
KVM: fix multi-queue tap creation

apoikos:
We create multi-queue tap interfaces in a loop, by passing the tap name
obtained by the first TUNSETIFF ioctl() to the next ioctl() calls.
This loop assumes that the `name` variable is a string, but goes on to
reassign bytes to it causing the next iteration to fail with an
`AttributeError`.

Fix this by using a separate bytes variable to hold the interface name
to/from the ioctl() request.

This issue was reported by magnus1 on freenode.


skorpy2009:
I can confirm this fixes the issue.

skorpy2009 (approved):

saschalucas (approved):
thanks a lot for this fix. seems like a left over from python2->3? maybe we can add mutiqueue test into @rbott 's qa test?

please merge your changes.

PR #1535, opened 2020-12-17
KVM: use qemu monitor to set spice password instead of QMP

gedia:
Before this patch ganeti used the more modern QEMU Machine protocol to set
the spice password for an instance upon startup. This is in contrast to the
method used when setting a VNC password, which employs qemu monitor to
the same effect.

However, it appears that this implementation is currently broken. Instances
with a set spice password will fail to start due to QMP timeout.

Why this is the case is not clear to me. Either QMP is not ready to serve requests
for the instance at the time the message is sent out by ganeti, or the timing of
the QMP request is such during the VM startup process that the password-setting
command cannot be executed.

This patch changes the method used to set a spice password for the instance
to match the one used for VNC (i.e. via qemu monitor), enabling instances
with password-protected spice servers to start without manual user intervention.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2050,13 +2050,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
         raise errors.HypervisorError("Failed to open SPICE password file %s: %s"
                                      % (spice_password_file, err))
 
-      qmp = QmpConnection(self._InstanceQmpMonitor(instance.name))
-      qmp.connect()
-      arguments = {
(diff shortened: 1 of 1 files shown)


saschalucas:
Thanks a lot for this finding. Could you please try this [patch](https://github.com/ganeti/ganeti/commit/9d13658ba77024f5398c94c2d3a2ebb8da073699.patch) or my [fork/branch](https://github.com/saschalucas/ganeti/tree/fix_spice_pw) if this solves your problem?

saschalucas:
I have created #1536 to fix this problem in a way not going back to HMP. Thanks for finding this issue.

PR #1536, opened 2020-12-21
switch setting spice password to decorator

saschalucas:
It has been reported (and verified), that setting a spice password is broken (see #1535):

Setting the spice password via a self managed QMP connection will block further QMP commands. The instance stays paused, because it will not receive the "cont" command.

Instead of going back to HMP, switch the spice password setting to use the @_with_qmp decorator.


iustin (approved):
This looks clean, I guess explicitly closing the existing one-off connection would have worked as well, but that would just do more work needlessly.

PR #1537, opened 2020-12-22
kvm: get known instances from runtime and not PID files

### LINKED ISSUE #1440: Powering instance down from within results in ERROR_down rather than USER_down
Issue reported as suggested by Sascha Lucas.

This is on Debian Buster with ganeti 2.16.0-5 and qemu-system-x86 1:3.1+dfsg-8+deb10u4.

Initially reported by Christian Baumann, Message-Id: <7ad37664-0d85-4361-9d6d-8c489e563b7d@googlegroups.com>.

We sometimes have a need to power down instances from within the instance. This is accomplished in Ganeti using the "user_shutdown" at either the cluster or instance level and with the --user-shutdown cluster option.

This appears to be broken on the above platform. When running a 'poweroff' from within an instance, the instance correctly powers down. The instance status however is reported as ERROR_down. Ganeti watcher then comes along and restarts the instance.

$ sudo gnt-cluster show | grep -E "(shutdown|acpi)"
    acpi: True
    user_shutdown: True
  enabled user shutdown: True
$ sudo gnt-instance show mysql-01.amazing-internet.net | grep -E "(shutdown|acpi)"
    acpi: default (True)
    user_shutdown: default (True)

Thanks.

apoikos:
KVM's `GetAllInstancesInfo` enumerates all "known" instances using their
PID files. This should be fine as long as we care about live instances,
but it gets more complicated in reality because `GetAllInstancesInfo` also
tries to gather information about user-down instances which are no
longer alive.

Up until QEMU 3.1 this worked reliably because QEMU would leave the
pidfile behind after exiting. Since a user-down instance was not cleaned
up by Ganeti itself, its pidfile was left behind and `GetAllInstancesInfo`
would find it and detect that it was shutdown by the user.

However, as of QEMU 3.1 the pidfile is removed by QEMU itself on exit
(see QEMU commit 90a84d131c "Delete PID file on exit"), and as a result
user-down instances are now omitted from the loop and reported as
ERROR_down by the master.

Fix this by using the KVM runtime files rather than the PID files to
enumerate all "known" instances. In theory this should be safe:
excluding QEMU, Ganeti manages the PID and runtime files using the same
lifecycle.

This fixes #1440.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1130,7 +1130,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
 
     """
     data = []
-    for name in os.listdir(self._PIDS_DIR):
+    for entry in os.listdir(self._CONF_DIR):
+      name, _ = os.path.splitext(entry)
       try:
         info = self.GetInstanceInfo(name)
       except errors.HypervisorError:


saschalucas (approved):
Thanks @apoikos for finding the root cause.... qemu sounds obvious (to me, afterwards). Please consider if you like to PR against `stable-2.16`, because there is the problem, too. Otherwise LGTM.

Issue #1542, opened 2020-12-24
Stop using QEMU's human monitor


apoikos:
Right now we are using a mix of HMP and QMP calls to QEMU, depending on the feature. This adds complexity to the code and confusion to contributors over picking which method to use. Since QMP is supposed to be the machine interface of QEMU, it makes sense to switch everything over to it.

We currently use the human monitor for the following operations:

 - To get CPU thread information (`info cpus`, QMP command: `query-cpus`, since 0.14)
 - To set the VNC password (`change vnc password`, QMP command: `change` with arguments, since 0.14)
 - To hot-add drives (`drive_add`, QMP command: `blockdev-add`, since 2.9)
 - To hot-remove drives (`drive_del`, QMP command: `blockdev-del`, since 2.9)
 - To signal shutdown to the guest (`system_powerdown`, QMP command: `system_powerdown`, since 0.14)
 - To balloon the guest memory (`balloon`, QMP command: `balloon`, since 0.14)
 - To get version information (`info version`, QMP command: `query-version`, since 0.14)

(The information on the QMP commands can be found under `qapi/` in the QEMU source)

Migrating all these calls to QMP bumps our compatibility to QEMU >= 2.9, which was released in 2017. Since we are anyway no longer compatible with Xenial and Stretch because of Python 3/GHC, there's nothing to lose here.

rbott:
I entirely agree. I will try to tackle some of these, since I have already done some work on this end in the past month. I'll update this issue when I start working on any of the above tasks.

rbott:
Parts if this PR have been addressed by #1546 (and bumped the minimum QEMU version to 2.12). However, there is still some work to do :-)

rbott:
@saschalucas @apoikos there is now only the block device part left to address. However, this seems to a bit more complicated: according to [this slidedeck](http://events17.linuxfoundation.org/sites/events/files/slides/talk_11.pdf) `blockdev-add|del` in QMP do not directly correspond to `drive_del|add` in HMP and especially not to the `-drive` commandline parameter of QEMU.

The latter is deprecated and should be replaced by the `-blockdev` parameter (which in turn is in line with `blockdev-add`).

So I guess we need to entirely replace the `-drive/drive_add/drive_del` combination. AFAICT this also makes the `disk_cache` parameter obsolete as it does not exist any more in this context (other block device related parameters like `aio`, `discard` etc. are still available).

[...]

PR #1546, opened 2020-12-27
KVM: shift `info` queries from human monitor to QMP

rbott:
This PR partially addresses #1542 in moving the `info cpus` and `info version` human monitor requests over to QMP. Moreover, it uses `query-cpus-fast` instead of `query-cpus`, because the latter interrupts/stalls the VM for a brief moment (according to the QEMU/QMP documentation)

However, `query-cpus-fast` is only available since QEMU 2.12, bumping the minimum required QEMU version to this level. Debian Stretch comes with 2.8 but is out-of-scope due to Python/GHC versions anyways. Debian Buster is fine (QEMU 3.1). CentOS 7 comes with 2.12, should be also fine.

We could resort to compatibility code (e.g. use `query-cpus` on older QEMU versions) but I do not think this is required/helpful.

I checked the `info version` part against the QA suite (needs the `instance-device-hotplug` test enabled) and did manual testing on a QA cluster for the CPU masking feature.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -605,9 +605,6 @@ class KVMHypervisor(hv_base.BaseHypervisor):
 
   _VERSION_RE = re.compile(r"\b(\d+)\.(\d+)(\.(\d+))?\b")
 
-  _CPU_INFO_RE = re.compile(r"cpu\s+\#(\d+).*thread_id\s*=\s*(\d+)", re.I)
-  _CPU_INFO_CMD = "info cpus"
-
   _DEFAULT_MACHINE_VERSION_RE = re.compile(r"^(\S+).*\(default\)", re.M)
   _CHECK_MACHINE_VERSION_RE = \
     staticmethod(lambda x: re.compile(r"^(%s)[ ]+.*PC" % x, re.M))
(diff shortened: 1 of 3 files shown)


lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -779,6 +787,14 @@ def GetMigrationStatus(self):
 
     return self.Execute("query-migrate")
 
+  @_ensure_connection
+  def GetVersion(self):
```
saschalucas, line 791: perhaps this extra function `GetVersion()` can be avoided? a version information is already available from the QMP [`connect()`/`greeting`](https://github.com/ganeti/ganeti/blob/3368a7b6a7dc5e038caa68ebaa2ef96eab65efdb/lib/hypervisor/hv_kvm/monitor.py#L296). And used in [`SetMigrationParameters()`](https://github.com/ganeti/ganeti/blob/3368a7b6a7dc5e038caa68ebaa2ef96eab65efdb/lib/hypervisor/hv_kvm/monitor.py#L716)

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1005,27 +998,26 @@ def _AssignCpuAffinity(cls, cpu_mask, process_id, thread_dict):
...
-    output = self._CallMonitorCommand(instance_name, self._CPU_INFO_CMD)
-    for line in output.stdout.splitlines():
-      match = self._CPU_INFO_RE.search(line)
-      if not match:
-        continue
-      grp = [int(g) for g in match.groups()]
-      result[grp[0]] = grp[1]
+    cpu_info = self.qmp.GetFastCpuInformation()
+
+    result = {}
```
saschalucas, line 1014: is `result` here initialized twice?

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1005,27 +998,26 @@ def _AssignCpuAffinity(cls, cpu_mask, process_id, thread_dict):
       for i, vcpu in enumerate(cpu_list):
         cls._SetProcessAffinity(thread_dict[i], vcpu)
 
-  def _GetVcpuThreadIds(self, instance_name):
+  @_with_qmp
+  def _GetVcpuThreadIds(self, instance):
     """Get a mapping of vCPU no. to thread IDs for the instance
 
     @type instance_name: string
```
apoikos, line 1005: The argument type changes as well.

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -771,6 +771,14 @@ def StartPostcopyMigration(self):
 
     self.Execute("migrate-start-postcopy")
 
+  @_ensure_connection
+  def GetFastCpuInformation(self):
```
apoikos, line 775: I'd leave "fast" out of the name here.

lib/hypervisor/hv_kvm/__init__.py
rbott: Yes, you are right. Good catch! Must have been late :-)

lib/hypervisor/hv_kvm/__init__.py
rbott: Yes, will be fixed. Missed that in different place as well.

lib/hypervisor/hv_kvm/monitor.py
rbott: Yes, that's probably not relevant outside the monitoring class. Will fix.

lib/hypervisor/hv_kvm/monitor.py
rbott: Yes, I entirely overlooked that the version is already parsed from the QMP greeting/handshake. I reverted to using the `version` property.

saschalucas (approved):
Thanks @rbott. If @apoikos has no further sugestions, LGTM.

saschalucas:
friendly ping @rbott. I thinks this PR is ready to merge?

rbott:
Yes, thanks for the reminder. Let's go :-)

PR #1555, opened 2020-12-30
KVM: fix disk hotplugging with QEMU 4.0+

### LINKED ISSUE #1547: disk hotplug broken with Ubuntu Focal / QEMU 4.2
I recently enabled hotplug tests in the testsuite for my nightly QA runs and they [fail|https://salsa.debian.org/rbott/ganeti-build/-/jobs/1288613] on Ubuntu Focal.

```
Command: ssh gnt-test01 'gnt-instance modify --disk -1:add,size=1G --hotplug kvm-test-instance02'
Stdout was:
Mon Dec 28 09:55:53 2020  - INFO: Trying to hotplug device...
Mon Dec 28 09:55:53 2020  - WARNING: Could not hotplug device: Error while executing backend function: kvm: error executing the device_add command: Property 'virtio-blk-device.drive' can't find value 'disk-e38c1b24-92f7-4f51' (GenericError):
Mon Dec 28 09:55:53 2020  - INFO: Continuing execution..
Mon Dec 28 09:55:53 2020  - INFO: Waiting for instance kvm-test-instance02 to sync disks
Mon Dec 28 09:55:53 2020  - INFO: - device disk/2:  3.20% done, 35s remaining (estimated)
Mon Dec 28 09:56:29 2020  - INFO: Instance kvm-test-instance02's disks are in sync
Modified instance kvm-test-instance02
 - disk/2 -> add:size=1024,mode=rw
 - disk/2 -> hotplug:failed
Please don't forget that most parameters take effect only at the next (re)start of the instance initiated by ganeti; restarting from within the instance will not be enough.
```

Hotplugging a NIC seems to be no problem. I have not yet investigated this any further. Just for the record: Focal runs QEMU 4.2.

apoikos:
As of QEMU 4.0, file-backed block devices are opened in read-only mode
and upgraded to read-write mode the first time a write needs to happen.
QEMU calls this "dynamic auto-read-only" and it can be turned off by
appending `auto-read-only=off` to a drive definition.

For auto-read-only to work with fdsets, the fdset backing a drive must
now include two file decriptors, one in read-only and one in read-write
mode, but Ganeti only passes a read-write file descriptor.

We could disable auto-read-only for hotplugged devices, or we can pass a
second, read-only fd in the same fdset. The former would require
version-specific overrides once more and would go against QEMU defaults,
so let's do the latter.

Since QEMU only accepts individual descriptors in `add-fd`, we need to
slightly change _AddFd to allow using the fdset allocated by QEMU for
both descriptors.

This fixes #1547.

### DIFF
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -538,9 +538,15 @@ class QmpConnection(MonitorSocket):
 
     """
     if os.path.exists(uri):
-      fd = os.open(uri, os.O_RDWR)
-      fdset = self._AddFd(fd)
-      os.close(fd)
+      # For QEMU's dynamic auto-read-only feature to work (on by default since
+      # QEMU 4.0), we need two file descriptors, one in read-write and one in
+      # read-only mode.
+      fd_rw = os.open(uri, os.O_RDWR)
+      fd_ro = os.open(uri, os.O_RDONLY)
+      fdset = self._AddFd(fd_rw)
+      self._AddFd(fd_ro, fdset)
+      os.close(fd_rw)
+      os.close(fd_ro)
       filename = "/dev/fdset/%s" % fdset
     else:
       # The uri is not a file.
@@ -814,7 +820,7 @@ class QmpConnection(MonitorSocket):
       logging.info("Passing fd %s via SCM_RIGHTS failed: %s", fd, err)
       raise
 
-  def _AddFd(self, fd):
+  def _AddFd(self, fd, fdset_id=None):
     """Wrapper around add-fd qmp command
 
     Send fd to a running process via SCM_RIGHTS and then add-fd qmp command to
@@ -822,6 +828,9 @@ class QmpConnection(MonitorSocket):
 
     @type fd: int
     @param fd: The file descriptor to pass
+    @type fdset_id: int
+    @param fdset_id: The ID of an existing fdset to add the fd to. If None, a
+    new fdset will be created.
 
     @return: The fdset ID that the fd has been added to
     @raise errors.HypervisorError: If add-fd fails for some reason
@@ -830,8 +839,11 @@ class QmpConnection(MonitorSocket):
     self._check_connection()
     try:
       utils.SendFds(self.sock, b" ", [fd])
-      # Omit fdset-id and let qemu create a new one (see qmp-commands.hx)
-      response = self.Execute("add-fd")
+      if fdset_id is not None:
+        arguments = {'fdset-id': fdset_id}
+      else:
+        arguments = None
+      response = self.Execute("add-fd", arguments)
       fdset = response["fdset-id"]
     except errors.HypervisorError as err:
       logging.info("Passing fd %s via SCM_RIGHTS failed: %s", fd, err)


lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -538,9 +538,15 @@ def HotAddDisk(self, disk, devid, uri, drive_add_fn=None):
 
     """
     if os.path.exists(uri):
-      fd = os.open(uri, os.O_RDWR)
-      fdset = self._AddFd(fd)
-      os.close(fd)
+      # For QEMU's dynamic auto-read-only feature to work (on by default since
+      # QEMU 4.0), we need two file descriptors, one in read-write and one in
+      # read-only mode.
```
iustin, line 543: I would add here a note that fdsets work with older versions; the wording right now is unclear on this.

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -538,9 +538,15 @@ def HotAddDisk(self, disk, devid, uri, drive_add_fn=None):
...
+      fd_rw = os.open(uri, os.O_RDWR)
+      fd_ro = os.open(uri, os.O_RDONLY)
+      fdset = self._AddFd(fd_rw)
+      self._AddFd(fd_ro, fdset)
```
iustin, line 547: QQ: is there a need for ordering here (rw before ro), or not? The man page for `qemu-system` shows this ordering, but it's not clear if any order is accepted or not.

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -814,14 +820,17 @@ def _GetFd(self, fd, fdname):
       logging.info("Passing fd %s via SCM_RIGHTS failed: %s", fd, err)
       raise
 
-  def _AddFd(self, fd):
+  def _AddFd(self, fd, fdset_id=None):
```
iustin, line 823: Too bad a fdset can't be created empty, this way the call sequence looks a bit ugly…

iustin (approved):
A few minor comments below, but LGTM.

lib/hypervisor/hv_kvm/monitor.py
apoikos: No need for ordering, I verified this against QEMU code: it merely iterates over all descriptors in the fdset until it finds one with matching flags.

lib/hypervisor/hv_kvm/monitor.py
apoikos: Agreed

rbott:
I ran the QA suite on Ubuntu Focal against your PR but it still ends up with this error:
```Command: ssh gnt-test01 'gnt-instance modify --disk -1:add,size=1G --hotplug kvm-test-instance02'
Stdout was:
Wed Dec 30 23:53:08 2020  - INFO: Trying to hotplug device...
Wed Dec 30 23:53:08 2020  - WARNING: Could not hotplug device: Error while executing backend function: kvm: error executing the device_add command: Could not reopen file: Permission denied (GenericError):
Wed Dec 30 23:53:08 2020  - INFO: Continuing execution..
Wed Dec 30 23:53:08 2020  - INFO: Waiting for instance kvm-test-instance02 to sync disks
Wed Dec 30 23:53:09 2020  - INFO: - device disk/2:  3.20% done, 39s remaining (estimated)
Wed Dec 30 23:53:48 2020  - INFO: Instance kvm-test-instance02's disks are in sync
Modified instance kvm-test-instance02
 - disk/2 -> add:size=1024,mode=rw
 - disk/2 -> hotplug:failed
```

Complete QA output can be found [here|https://salsa.debian.org/rbott/ganeti-build/-/jobs/1296933]. Is there anything I more specific I could test?

This is the output of the HMP "conversation" when I tried to hotplug a disk once again after the testsuite failed.
```
QEMU 4.2.0 monitor - type 'help' for more information
(qemu) drive_add dummy file=/dev/fdset/2,if=none,id=disk-7c2d5004-0f10-4432,format=raw
warning: Opening a block device as a file using the 'file' driver is deprecated
OK
(qemu) 
```

apoikos:
OK, this means we fixed the `drive_add` part, now we need to see what's wrong with `device-add`.

apoikos:
Unfortunately it turns out the HMP `drive_add` & QMP `device_add` combination is broken when auto-read-only is enabled. OTOH, using `blockdev-add` over QMP instead of `drive_add` over HMP works for QEMU 5.1. I think we should temporarily turn off `auto-read-only` for hotplugged disks and release 3.0.1 with it, and then work on replacing `drive_add` with `blockdev-add`, which needs quite a bit of reading to get right.

apoikos:
I'm abandoning this PR in favor of disabling auto-read-only. Will open a new PR soon.

PR #1561, opened 2021-01-03
hotplug: set disk parameters

### LINKED ISSUE #1559: KVM: disk hotplug does not respect disk_cache
When a disk is hot pluged into an instance, I noticed different performance characteristics on the added disk.

Normally I set the hypervisor parameter `kvm:disk_cache=none` to force direct I/O. I think Ganeti overrides internally `kvm:disk_cache` to `none` for certain disk templates for working live migration/cache coherence? Looking at the cause, yields to that the hot pluged disk has a different (default) cache setting:

```
# Cluster=10.255.255.5 Ganeti=https://github.com/saschalucas/ganeti:backports/stable-3.0/disk_hotplug OS=ubuntu-focal 

root@node02:~# socat STDIO UNIX-CONNECT:/var/run/ganeti/kvm-hypervisor/ctrl/test.vm.monitor 
QEMU 4.2.1 monitor - type 'help' for more information
(qemu) info block
info block
disk-c3c6a083-74ef-44d3 (#block124): /var/run/ganeti/instance-disks/test.vm:0 (raw)
    Attached to:      /machine/peripheral/disk-c3c6a083-74ef-44d3/virtio-backend
    Cache mode:       writeback, direct
...
disk-bca4fa12-0194-4517 (#block324): /dev/fdset/0 (raw)
    Attached to:      /machine/peripheral/disk-bca4fa12-0194-4517/virtio-backend
    Cache mode:       writeback
```
One can see, that the hot plugged disk (/dev/fdset/0) has the default cache mode `writeback` without `direct`. I think this is wrong and unsafe for live migration i.e. on DRBD?

saschalucas:
When hot plugging a disk, paramters should match the current runtime. I.e. for live migration the cache mode is [critical](https://github.com/ganeti/ganeti/blob/74b82e754387e84913b7df372c7f27884a5656ee/lib/hypervisor/hv_kvm/__init__.py#L1206-L1217). Other parameters like aio, discard and unimplemented I/O (throtteling groups)[#1274] etc. are also relevant, but uncritical.

The documetation of HMP `drive_add` is unclear. It seems, that it supports the same parameters like -drive command line option???

Current cache settings can be checked by:
* HMP: `info block -n`
* QMP: `{ "execute": "query-named-block-nodes" }`

I was not able to check for aio and discard. So I must assume, that the HMP/`drive_add` correctly accepts this parameters.

This closes #1559

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2298,6 +2298,14 @@ class KVMHypervisor(hv_base.BaseHypervisor):
           # after QEMU 4.0. auto-read-only first appeared in 3.1, but 4.0
           # changed its behavior in a way that breaks hotplugging. See #1547.
           cmd += ",auto-read-only=off"
+        # When hot plugging a disk, parameters should match the current runtime.
+        # I.e. for live migration, the cache mode is critical.
+        if up_hvp[constants.HV_DISK_CACHE] != constants.HT_CACHE_DEFAULT:
+          cmd += ",cache=%s" % up_hvp[constants.HV_DISK_CACHE]
+        if up_hvp[constants.HV_KVM_DISK_AIO] == constants.HT_KVM_AIO_NATIVE:
+          cmd += ",aio=%s" % up_hvp[constants.HV_KVM_DISK_AIO]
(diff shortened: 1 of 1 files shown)


atta (approved):
looks good to me

PR #1562, opened 2021-01-04
optimize available queue time on larger clusters

saschalucas:
On larger clusters with DRBD instances (i.e. 350+ instance disks) the disk verification takes a long time (1.5 minutes). Running this every 5 minutes via cron reduces the window where other jobs can run to 3.5 minutes (66%). This is very annoying. Also I did not observe that DRBD connections are horribly unstable and disconnect every time. So checking them once per hour should be enough.

### DIFF
--- a/doc/examples/ganeti.cron.in
+++ b/doc/examples/ganeti.cron.in
@@ -3,8 +3,11 @@ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin
 # On reboot, continue a Ganeti upgrade, if one was in progress
 @reboot root [ -x @SBINDIR@/gnt-cluster ] && @SBINDIR@/gnt-cluster upgrade --resume
 
-# Restart failed instances (every 5 minutes)
-*/5 * * * * root [ -x @SBINDIR@/ganeti-watcher ] && @SBINDIR@/ganeti-watcher
+# Restart failed instances (every 5 minutes, expect at minute 0)
+5-55/5 * * * * root [ -x @SBINDIR@/ganeti-watcher ] && @SBINDIR@/ganeti-watcher --no-verify-disks
+
+# Verify and repair degraded DRBD disks and restart failed instances (once per hour, at minute 0)
+0 * * * * root [ -x @SBINDIR@/ganeti-watcher ] && @SBINDIR@/ganeti-watcher
 
 # Clean job archive (at 01:45 AM)
 45 1 * * * @GNTMASTERUSER@ [ -x @SBINDIR@/ganeti-cleaner ] && @SBINDIR@/ganeti-cleaner master


rbott:
I have not yet observed flaky DRBD connections (e.g. "once they are up, everything is fine) - from that angle, an extended time period seems doable.

However, sometimes DRBD connections are not established at all at instance creation (e.g. due to firewalling issues or other configuration problems outside of DRBD/Ganeti itself). If you run the verify code only once an hour, you might miss a broken DRBD setup for quite a while.

On top of that I have not yet looked at the verify code for other storage backends. There might be reasons for more frequent checks.

The obvious other way would be to improve the speed of DRBD checks but that sounds like a more invasive and/or impossible path to go down.

Just my two cents :-)

saschalucas:
> However, sometimes DRBD connections are not established at all at instance creation (e.g. due to firewalling issues or other configuration problems outside of DRBD/Ganeti itself).

I think instance creation will fail if not all disks are in sync. Maybe you are referring to `--no-wait-for-sync`? Can you give an example how this situation can happen? ATM I can't imagine. If there exists configuration problems outside of Ganeti, is it likely that they are resolved 5 minutes later?

> On top of that I have not yet looked at the verify code for other storage backends. There might be reasons for more frequent checks.

According to `ganeti-watcher(8)` and `gnt-cluster(8)` verify-disks seems only related to DRBD.

iustin:
Hmm, why does verify disks prevent _all_ other jobs from running?

saschalucas:
> Hmm, why does verify disks prevent _all_ other jobs from running?

With _all_ I mean things I'm doing mostly: `instance add|modify|reboot`. Somehow `verify-disks` locks the entire node group, so other commands needing a cluster level lock are waiting. If some of my usual commands do not respond, I take a look at the clock, and see: ah it's `*/5` cron time for `verify-disks`, I have to wait until `*/5+1.5`.

iustin:
> > Hmm, why does verify disks prevent _all_ other jobs from running?
> 
> With _all_ I mean things I'm doing mostly: `instance add|modify|reboot`. Somehow `verify-disks` locks the entire node group, so other commands needing a cluster level lock are waiting. If some of my usual commands do not respond, I take a look at the clock, and see: ah it's `*/5` cron time for `verify-disks`, I have to wait until `*/5+1.5`.

That is interesting (I forgot that behaviour).

I don't know what the best imediate fix is - I think one hour is likely too much, I would suggest 15 minutes (that would make it 10% of the time, rather than 33%) for a start. Or 30 minutes.

A better fix would be to remove the node group lock, if possible - needs to be investigated - and/or make it not verify all node groups at once/all instances at once.

zen2:
I wonder what is the accurate purpose of `verify-disks` .
I looked at `cmd_lib/instance_storage.py` and `storage/drbd.py` and it seems that `verify-disks` try to `assemble` all drbd disks in order to do his check.
Can't we rely on `/proc/drbd` status tor that ?
Or miss I something really important in drbd `assemble` code ?

saschalucas:
> Can't we rely on `/proc/drbd` status tor that ?

I thinks yes. Checking `/proc/drbd` should show if there are somewhere degraded disks. This does not sound like needing a look on the node group, unless something is to repair. Maybe here is some/huge optimization possible?

Last time on the [attic](https://github.com/ganeti/ganeti/wiki/AtticMasterCherrypicks), I discovered this commits:
* commit beff5977e "Prevent watcher from holding lock on verify group" (does not apply to DRBD?)
* commit 3a366d647 "introduce a non-strict flag for group verify disk"

I think both do not solve my problem, because they are related to that the watcher can run more often because of other blocking jobs. In my case the watcher/verify-disks is the most longest blocking job.

iustin:
I didn't comment yet since I need to refresh my memory about how this LU works. At least ages ago, it was not possible to rely only on `/proc/drbd` since entirely missing minors are not there. But not sure if static minor allocation doesn't solve that.

In any case, it is very likely that this can be significantly improved, indeed.

saschalucas:
> ... At least ages ago, it was not possible to rely only on `/proc/drbd` since entirely missing minors are not there. ...

There is definitely a situation where DRBD minors are not in `/proc/drbd`: instances which have secondaries on nodes that have been rebooted. However, to every DRBD minor not being there belongs an other minor on the primary side being not `UpToDate/UpToDate`. The idea could be, that this is enough to trigger a repair?

iustin:
> > ... At least ages ago, it was not possible to rely only on `/proc/drbd` since entirely missing minors are not there. ...
> 
> There is definitely a situation where DRBD minors are not in `/proc/drbd`: instances which have secondaries on nodes that have been rebooted. However, to every DRBD minor not being there belongs an other minor on the primary side being not `UpToDate/UpToDate`. The idea could be, that this is enough to trigger a repair?

Not even primaries are always reliable - at least, not 100%, due to drdb module bugs (might have been fixed 10 years ago).

I think that the whole logic should be rewritten, in any case, instead of relying on that file as authoritative. Have some idea, but need to check the code more.

In the meantime, I still think 15-30 minutes is a good compromise.

zen2:
If a drbd device is missing on a node, it will be reported as a problem on his twin node.

I don't know if that can help but this is my nagios check script that I'm using since years to monitor drbd status and it's always the first alert I got in case of problem especially for problem related to drbd network link.
It look on primary and secondary drbd minors status and avoid drbd device in sync.
It generate a exit status:
- STATE_OK=0 if no problem at all
- STATE_WARNING=1 in case of problem on secondary drbd minors
- STATE_CRITICAL=2 in case of problem on primary drbd minors


```
#!/bin/bash
#
# check_drbd.sh 
#
# check /proc/drbd for no UpToDate disks
#
# jfmaeyhieux[at]atanar.net
#

STATE_OK=0
STATE_WARNING=1
STATE_CRITICAL=2
STATE_UNKNOWN=3

DRBD_P="$(grep "^[ 0-9]*: cs:" /proc/drbd | egrep -v " ro:Secondary/|^[ 0-9]*: cs:Connected ro:Primary/Secondary ds:UpToDate/UpToDate |^[ 0-9]*: cs:SyncSource ro:Primary/Secondary ds:UpToDate/Inconsistent |^[ 0-9]*: cs:PausedSyncS ro:Primary/Secondary ds:UpToDate/Inconsistent ")"
DRBD_S="$(grep "^[ 0-9]*: cs:" /proc/drbd | egrep -v " ro:Primary/|^[ 0-9]*: cs:Connected ro:Secondary/Primary ds:UpToDate/UpToDate |^[ 0-9]*: cs:SyncTarget ro:Secondary/Primary ds:Inconsistent/UpToDate |^[ 0-9]*: cs:PausedSyncT ro:Secondary/Primary ")"
DRBD_P_STATUS="$( echo -e "$DRBD_P" | grep -v "^$" | wc -l)"
DRBD_S_STATUS="$( echo -e "$DRBD_S" | grep -v "^$" | wc -l)"

case $DRBD_P_STATUS in
        0) STATUS_P=$STATE_OK ; out_p="-All Primary DRBD devices are UpToDate" ;;
        *) STATUS_P=$STATE_CRITICAL ; out_p="-The following Primary DRBD devices are not UpToDate:\n$DRBD_P" ;;
esac

case $DRBD_S_STATUS in
        0) STATUS_S=$STATE_OK ; out_s="-All Secondary DRBD devices are UpToDate" ;;
        *) STATUS_S=$STATE_WARNING ; out_s="-The following Secondary DRBD devices are not UpToDate:\n$DRBD_S" ;;
esac

out="\n${out_p}\n${out_s}"

if [ $STATUS_P -gt $STATUS_S ]; then
	STATUS=$STATUS_P
else
	STATUS=$STATUS_S
fi

case $STATUS in
        $STATE_OK)       echo -e "OK $out" ;     exit $STATE_OK ;;
        $STATE_WARNING)  echo -e "WARNING $out" ;  exit $STATE_WARNING ;;
        $STATE_CRITICAL) echo -e "CRITICAL $out" ; exit $STATE_CRITICAL ;;
        *)               echo -e "UNKNOWN $out" ;  exit $STATE_UNKNOWN ;;
esac
```

saschalucas:
I'm closing this in favor of #1565, which seems the root cause, why `verify-disks` takes so long.

PR #1563, opened 2021-01-05
fix pattern substitution in man pages for focal

saschalucas:
Ganeti man pages contain patterns that are replaced later via a generated sed script (`autotools/replace_vars.sed`). Examples are the version (`@GANETI_VERSION@`) or local state dir (`@LOCALSTATEDIR@`). Pandoc transforms this `@` into `\@`, which is why some sed commands replace it back (`s/\\@/@/g`). For man pages (`-t man`) with Ubuntu Focal pandoc 2.5 `@` is now transformed into `\[at]`, which leads to `autotools/replace_vars.sed` not matching any more.

The solution here replaces the patterns before they are fed to pandoc, avoiding additional replace-back-patterns. As a side effect the existing replace-back-pattern and the intermediate step for `maninputs` (*.in) can be removed.

Old steps were:
* ./autotools/docpp: *.rst -> *.gen
* pandoc -t man/html: *.gen + footer.rst -> *.in
* replace_vars.sed: \*.in -> man/*.html

New steps are:
* ./autotools/docpp|replace_vars.sed: *.rst -> *.gen
* pandoc -t man/html: \*.gen + footer.rst -> man/*.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,7 +111,6 @@
 # man
 /man/*.[0-9]
 /man/*.html
-/man/*.in
 /man/*.gen
 
 # test/hs
--- a/Makefile.am
+++ b/Makefile.am
@@ -294,7 +294,6 @@ APIDOC_PY_DIR = $(APIDOC_DIR)/py
 APIDOC_HS_DIR = $(APIDOC_DIR)/hs
 
 MAINTAINERCLEANFILES = \
-	$(maninput) \
 	doc/install-quick.rst \
 	doc/news.rst \
 	doc/upgrade.rst \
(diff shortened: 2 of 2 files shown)


iustin (approved):
This looks very clean and simple solution, so LGTM. Can you confirm you looked at the results and they look ok? (e.g. pandoc doesn't mangle the replaced strings in any way?).

Small nitpick: commit title should have first word capitalized 'Fix'.

saschalucas:
> ... Can you confirm you looked at the results and they look ok? (e.g. pandoc doesn't mangle the replaced strings in any way?).

Yes I can confirm. `autotools/replace_vars.sed` contains not much special at all, even less related to man pages. I can only identify patterns like path names and the version. For this patterns I took samples... it turns out, that this way, it might be even better? E.g. the `~` in `3.1.0~alpha1` gets transformed to `\[ti]`. Viewing the man page, the version is rendered as expected (`3.1.0~alpha1`). The html variant of the man pages seems not effected at all (either before nor after this PR).

> Small nitpick: commit title should have first word capitalized 'Fix'.

Will fix this.

Thanks @iustin.

PR #1568, opened 2021-01-18
Update monitor.py in line with #1268 and #1326

regnauld:
Vectors is 2N+2 (updates #1268 #1326)
Fix typo


saschalucas (approved):
Thank you for discovering.

PR #1572, opened 2021-01-23
kvm: support multiple devices for boot order preference

gedia:
This patch adds multiple device support for the boot_order
setting. Valid values are still disk, cdrom, network and floppy,
but the user can now supply a comma-separated list of those instead
of a single value only.

Additionally, this patch will now use the newer
'-boot order=drives' syntax for the qemu command line,
instead of the deprecated '-boot drives'.

### DIFF
--- a/lib/hypervisor/hv_base.py
+++ b/lib/hypervisor/hv_base.py
@@ -167,6 +167,21 @@ def ParamInSet(required, my_set):
   err = ("The value must be one of: %s" % utils.CommaJoin(my_set))
   return (required, fn, err, None, None)
 
+def AllParamsInSet(required, my_set):
+  """Builds parameter checker for set membership for collection of items.
+  @type required: boolean
+  @param required: whether this is a required parameter
+  @type my_set: tuple, list or set
+  @param my_set: allowed values set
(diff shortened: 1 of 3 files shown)


saschalucas:
Thanks for your contribution. This looks almost reviewable for me (with my limited capabilities).

Without looking in depth I think the extended `boot_order` syntax should be documented in the man page of `gnt-instance`, too.

I wonder if we should not take Ganeti downgrades into account. Say this will be included in Ganeti-3.1, there must be a possibility to trim the extended `boot_order` list back to a single (the first?) value, so that a downgraded Ganeti-3.0 will understand the cluster config. What do @apoikos and @iustin think?

Ah, and please check maximum line length to be <=80. It looks like there are a view longer one.

rbott:
I agree with @saschalucas here, we need some downgrade code in lib/tools/cfgupgrade.py. AFAIR this means we have to create a new helper method (e.g. DowngradeBootOrder()) and reference it from DowngradeAll(). Also, the `gnt-instance` documentation needs an update on the `boot_order` parameter.

Other than that, LGTM :-)

PR #1575, opened 2021-02-02
SocketOperation: return the number of chars written

apoikos:
Callers deal with strings, so it's the number of chars we should return
rather than the number of bytes.

Note that this is only relevant if we pass non-ascii information over
HTTP.


saschalucas (approved):
Thanks. Please merge your changes.

PR #1579, opened 2021-02-03
Add default arguments for the metadata daemon

jfut:
The metadata daemon accepts arguments at startup.

There was a case where we wanted to change the default port 80 with the `-p` option.


iustin (approved):
Looks good, thanks!

Issue #1574, opened 2021-02-01
xen migration fails due to missing xen config


Wurzelmann:
OS: Debian Buster
ganeti software version: 2.16 / 3.0.0
ganeti Debian version: 3.0.0-1~bpo10+1
hypervisor: xen (version 4.11)

This has been tested on ganeti 2.16 and 3.0 and both are affected.

When migrating instances from one node to another, it fails when there's no xen config for said instance (e.g. "test") in **/etc/xen/**. This config file only gets created when adding the instance and is not necessary for the migration at all (case in point: xl migrate is perfectly able to migrate said instance without it) and disappears after the first migration.

Example error message:

```
2021-01-26 13:30:30,091: job-197 pid=11195 ERROR Failed fetching source
migration information from node2: Failed to fetch migration information: 
Failed to load Xen config file: [Errno 2] No such file or directory: 
'/etc/xen/test'
```

This problem has already been reported in 2018 ([relevant Google Groups Thread](https://groups.google.com/g/ganeti/c/IerQ9r8axSs/m/ZmlDOSgaAwAJ)) and seems to not have been resolved yet.

We are currently trying to decide whether we want to use ganeti for our clusters and are therefore testing ganeti, but this issue is a hurdle we have to overcome. A crude hack [has been created](https://gist.github.com/Wurzelmann/790a8d57350c634db73d4eea4f594e6c), but this is far from a great solution and we'd be glad if this could be fixed in a more sustainable fashion. Additionally we'd be glad to test any fixes you throw our way, because this situation is less than ideal. :-)

saschalucas:
Hi,

thanks for reporting this issue. Sadly the support for Xen hypervisor in Ganeti is undefined. Currently no one in the organization has the knowledge and/or time for testing and fixing Xen (all using KVM only). Any help would be highly appreciated.

Wurzelmann:
I see, thank you for the insight. Since XEN does not depend on the config files in **/etc/xen/** for migration, we wondered if this had been a requirement for KVM instances and has been implemented for XEN too, although not needed?

PR #1582, opened 2021-02-05
fix live migration of xen instances

Wurzelmann:
This PR fixes issue #1574. Config deletion has been unified (setting the config path first and deleting the file, if it exists and creating it) and an improved runtime check has been introduced.

The test used before just took the cpu usage in seconds into account, which lead to problems on our new cluster due to its quite beefy AMD cpus; just migrating an idle instance did not increase the number of cpu seconds above 0.


saschalucas (approved):
Thanks for your fix!

PR #1584, opened 2021-02-16
fix parameter check on attaching ext disks

### LINKED ISSUE #1583: bug: attaching previously detached disk of type ext always fails
Hi,

I found that attaching a previously detached disk of type ext will always fails because it simultaneously requires *only* a name xor uid and an provider for extstorage as a parameter.
As I understand it the error is raised by the call to CheckDiskExtProvider in [1] and this would need to a) check which action is used on the disk or b) read the provider from config in order tor work correctly.

Cheers, Valentin
[1] https://github.com/ganeti/ganeti/blob/0b037dc450a34da6550ae8587bc4418069a7ab1f/lib/cmdlib/instance_storage.py#L368 

## process to reproduce:
* create instance with ext storage and 2 disks
* detach the secondary disk
* attach the secondary disk

## result:
```
$ gnt-instance modify --disk 1:attach XX
Failure: prerequisites not met for this operation:
error type: wrong_input, error details:
Missing provider for template 'ext'
$ gnt-instance modify --disk 1:attach,provider=lvm XX
Failure: prerequisites not met for this operation:
error type: wrong_input, error details:
Only one argument is permitted in attach op, either name or uuid
```

## expected result:
attach the disk

vali-um:
Only check for provider if a disk is added, not on attaching.
This fixes  #1583.

Signed-off-by: Valentin <valentin@ominous.space>

### DIFF
--- a/lib/cmdlib/instance_set_params.py
+++ b/lib/cmdlib/instance_set_params.py
@@ -211,10 +211,12 @@ class LUInstanceSetParams(LogicalUnit):
       if name is not None and name.lower() == constants.VALUE_NONE:
         params[constants.IDISK_NAME] = None
 
-    # This check is necessary both when adding and attaching disks
+    # These checks are necessary when adding and attaching disks
     if op in (constants.DDM_ADD, constants.DDM_ATTACH):
       CheckSpindlesExclusiveStorage(params, excl_stor, True)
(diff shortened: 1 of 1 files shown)


saschalucas (approved):
Thanks for your fix.

PR #1586, opened 2021-02-17
fix lvcreate for debian11

### LINKED ISSUE #1585: disk: "Device creation failed" debian bullseye with lvm backend
## issue
on debian11 bulseye creation of disk devices with lvm backed staorage templates fails if the underlying blockdevice contains any filesystem signature.
This is due to a change in lvcreate. Previously (e.g. under buster) upon abortion of signature wiping the lvcreate command succesfully continued wih a warning printed. In Bullseye it aborts and returns exit code 5.

## steps to reproduce (on bullseye)
```
gnt-instance add --disk 0:size=10g,name=foo-disk  --disk-template plain --os-type debootstrap+default foo
gnt-instance remove foo
gnt-instance add --disk 0:size=10g,name=foo-disk  --disk-template plain --os-type debootstrap+default foo
```

## result
```
Wed Feb 17 12:20:13 2021 * disk 0, size 10.0G
Wed Feb 17 12:20:13 2021 * creating instance disks...
Wed Feb 17 12:20:13 2021  - WARNING: Device creation failed
Failure: command execution error:
Can't create block device <LogicalVolume(/dev/XX/87936f68-7d39-4817-b365-ecbd4ee1a72a.disk0, visible as /dev/disk/0, size=10240m)> on node whiskey for instance foo: Can't create block device: LV create failed (exited with exit code 5): WARNING: ext4 signature detected on /dev/XX/87936f68-7d39-4817-b365-ecbd4ee1a72a.disk0 at offset 1080. Wipe it? [y/n]: [n]
  Aborted wiping of ext4.
  1 existing signature left on the device.
  Failed to wipe signatures on logical volume whiskey-vg/87936f68-7d39-4817-b365-ecbd4ee1a72a.disk0.
  Aborting. Failed to wipe start of new LV.
```

## expected result
successful instance creation

vali-um:
fixes #1585

### DIFF
--- a/lib/storage/bdev.py
+++ b/lib/storage/bdev.py
@@ -215,7 +215,8 @@ class LogicalVolume(base.BlockDev):
     # create an optimally-striped volume; in that case, we want to try
     # with N, N-1, ..., 2, and finally 1 (non-stripped) number of
     # stripes
-    cmd = ["lvcreate", "-L%dm" % size, "-n%s" % lv_name]
+    # Use -Wn to not wipe signatures as trying to do so may fail on debian11
+    cmd = ["lvcreate", "-Wn", "-L%dm" % size, "-n%s" % lv_name]
     for stripes_arg in range(stripes, 0, -1):
       result = utils.RunCmd(cmd + ["-i%d" % stripes_arg] + [vg_name] + pvlist)
       if not result.failed:
--- a/lib/storage/bdev.py
+++ b/lib/storage/bdev.py
@@ -215,7 +215,9 @@ class LogicalVolume(base.BlockDev):
     # create an optimally-striped volume; in that case, we want to try
     # with N, N-1, ..., 2, and finally 1 (non-stripped) number of
     # stripes
-    # Use -Wn to not wipe signatures as trying to do so may fail on debian11
+    # When run non-interactively, newer LVM versions will fail (unless
+    # `--yes` is specified) when an existing filesystem signature is
+    # encountered while creating a new LV. Using `-Wn` disables this check.
     cmd = ["lvcreate", "-Wn", "-L%dm" % size, "-n%s" % lv_name]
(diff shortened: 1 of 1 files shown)


lib/storage/bdev.py
```diff
@@ -215,7 +215,8 @@ def Create(cls, unique_id, children, size, spindles, params, excl_stor,
     # create an optimally-striped volume; in that case, we want to try
     # with N, N-1, ..., 2, and finally 1 (non-stripped) number of
     # stripes
-    cmd = ["lvcreate", "-L%dm" % size, "-n%s" % lv_name]
+    # Use -Wn to not wipe signatures as trying to do so may fail on debian11
```
apoikos, line 218: I think the comment here can be a bit more descripting, something along the lines:

```
# When run non-interactively, newer LVM versions will fail (unless `--yes` is specified) when an existing filesystem signature is encountered while creating a new LV. Using `-Wn` disables this check.
```

The same should probably be mentioned in the commit message as well. Note that this is not a debian-specific issue, it's a change in the behavior of `lvcreate` in LVM2 >= 2.03.10.

apoikos (commented):
Thanks for the bug report and the PR @vali-um!

It would be perfect if we could have a more descriptive commit message, other than that it looks good.

Note to self and others: `-Wn` was added in LVM2 2.02.105, January 2014, so this change won't break any systems that Ganeti is expected to run on.

lib/storage/bdev.py
vali-um: Thanks for your suggestion. I added the comment

vali-um:
Thanks for your response.
I hope a new commit with this message was the intended way to add this comment.
If not please tell me what to do better next time.
I will definitely try to write mire descriptive commit messages.

apoikos (approved):

apoikos:
Merged, thanks @vali-um!

PR #1587, opened 2021-02-21
Unbreak and enhance the chroot hypervisor

saschalucas:
When using the chroot hypervisor it turns out, that it is currently broken/can not be started:

> "Error while executing backend function: Instance test.vm is not running"

The trace from `noded.log` yields to `GetInstanceInfo`, which returns a HypervisorError, when then instance is not running. Fixing this by returning `None` like other HVs do.

The second change enhances the console command to use `/bin/sh` as a chroot shell, which is more likely to be available than `/bin/bash`, what ever the chroot may contain.


iustin (approved):
Both changes look good. Oh my, `raise HypervisorError`!

PR #1589, opened 2021-04-12
switch name- and ip-checking to opt-in

### LINKED ISSUE #1384: Make "--no-*-check" the default or cluster-wide configurable
Many people use the ``--no-{ip,name}-check`` toggles if creating instances. Probably, this should be either the default, which then would introduce a breaking change, or, probably better, be a cluster-wide option to set, which then can still be overwritten manually if there is a need to.

saschalucas:
As often discussed (at Ganeticons etc.) many people like to just create
instances without the need to resolv their name via DNS or /etc/hosts.
So a common boilerplate in the add command is to use `--no-name-check`
`--no-ip-check`. This commit inverts the logic to an opt-in model.

During changing the code, I disconverd an undocumented value of the
instance's NIC: `ip=auto`, which resolves the instance name and sets the
IP to the result. Until now, I was not aware of this.

This fixes #1384


rbott:
Thanks for the PR @saschalucas 

I have a few thoughts on this:
Changing the CLI might break scripts. We should probably - at least for a deprecation period - also keep the `--no-[something]` parameters and just change and document the default behavior. Maybe we can have the CLI tools issue a stderr warning if the deprecated parameters are used?

Another aspect: we should include this explicitly in the QA suite. From a quick glance I think it currently validates e.g. the name check but just because it does not disable it explicitly. I would opt (but possibly in a separate PR?) to test both variants in the QA suite.

What do you think?

saschalucas:
> Changing the CLI might break scripts. We should probably - at least for a deprecation period - also keep the `--no-[something]` parameters and just change and document the default behavior. Maybe we can have the CLI tools issue a stderr warning if the deprecated parameters are used?

It will also break scripts that rely on the old default name-/ip-check. For this case, there is no backwards compatibility. However, I've readded the old options with a deprecation warning. PTAL.
 
> Another aspect: we should include this explicitly in the QA suite. From a quick glance I think it currently validates e.g. the name check but just because it does not disable it explicitly. I would opt (but possibly in a separate PR?) to test both variants in the QA suite.

Indeed. I'm still unfamiliar with the QA suite, but it looks like that a fix needs to be addressed in this PR:
https://github.com/ganeti/ganeti/blob/592c0e945cc2c7b0013f813ea8c9d8ec0d5bab98/qa/qa_instance.py#L452-L456
I'll need some time to get in touch with QA.

rbott:
I just ran the QA suite against this PR and it breaks because the QA suite actually tests the name/ip-check behavior (somewhat implicitly).

The test `TestInstanceRenameAndBack` manipulates the `/etc/hosts` and expects the rename to fail:
```
---- FAILED [TestInstanceRenameAndBack] gnt-instance rename: Command 'gnt-instance rename kvm-test-instance01.staging.ganeti.org kvm-test-instance01.staging.ganeti.org' on node disco-stu.staging.ganeti.org was expected to fail but didn't
```

With the checks disabled the command does not fail any more and thus is unexpected to the QA suite. Would could alter the tests to
- accept the new default behavior and not expect the command to fail
- add an extra step where we add the --name-check/--ip-check parameters and expect _that_ step to fail

This way we would have both variants covered.

lib/cli_opts.py
```diff
@@ -818,15 +822,33 @@ class CliOption(Option):
...
 
-NOIPCHECK_OPT = cli_option("--no-ip-check", dest="ip_check", default=True,
-                           action="store_false",
-                           help="Don't check that the instance's IP"
-                           " is alive")
+IPCHECK_OPT = cli_option("--ip-check", dest="ip_check", default=False,
+                           action="store_true",
+                           help="Check that the instance's IP is alive (ping)")
+
+def WarnDeprecatedOption(option, opt_str, value, parser):
```
apoikos, line 829: This should probably live in `lib/cli.py`.

apoikos (approved):
@saschalucas thanks for this, we had been discussing it for quite a while.

If I recall correctly, the default behavior of checking IP and DNS had to do with how Ganeti was deployed initially at Google as part of a pipeline where an earlier step would do IPAM/DNS provisioning. It is true that most setups are simpler than that, and virtually every one I've seen uses `--no-name-check` and `--no-ip-check` for instance creation.

man/gnt-instance.rst
```diff
@@ -208,9 +207,9 @@ an instance with that name has already been added to the configuration
 as a forthcoming instance and the request is to replace this instance
 by the newly created real one.
 Note that if the reason for reserving an instance is that DNS names
-still need to be propagated, the reservation has to be done with
-``--no-name-check`` and ``--no-ip-check`` as these options are not
-implied by ``--forthcoming``.
+still need to be propagated, the reservation has to be done without
+``--name-check`` and ``--ip-check`` as these options are not implied
```
iustin, line 211: The paragraph here doesn't make a lot of sense (the negation when they're negated already). I would just drop the "as these options are not implied by --forthcoming".

man/gnt-instance.rst
```diff
@@ -1590,30 +1589,21 @@ options.
...
 Renames the given instance. The instance must be stopped when running
-this command. The requirements for the new name are the same as for
-adding an instance: the new name must be resolvable and the IP it
-resolves to must not be reachable (in order to prevent duplicate IPs
-the next time the instance is started). The IP test can be skipped if
-the ``--no-ip-check`` option is passed.
+this command. The ``--name-check`` checks for the new instance name via
+the resolver (e.g. in DNS or /etc/hosts, depending on your setup) and
+that the resolved name matches the new name. In addition the
+``--ip-check`` can be used to prevent duplicate IPs, by the new name's
```
iustin, line 1600: "by checking that the new…"

test/py/cmdlib/instance_unittest.py
```diff
@@ -259,7 +261,7 @@ def testAutoIpNoNameCheck(self):
                          ip_check=False,
                          name_check=False)
     self.ExecOpCodeExpectOpPrereqError(
-      op, "IP address set to auto but name checks have been skipped")
+      op, "IP address set to auto but name checks not enabled")
```
iustin, line 264: "but the name checks are not enabled" (add verb to mirror the first half of the sentence).

iustin (commented):
Wow, talk about a blast from the past. I've left a couple of minor wording changes, but looks good from my side.

iustin:
> If I recall correctly, the default behavior of checking IP and DNS had to do with how Ganeti was deployed initially at Google as part of a pipeline where an earlier step would do IPAM/DNS provisioning. It is true that most setups are simpler than that, and virtually every one I've seen uses `--no-name-check` and `--no-ip-check` for instance creation.

The details were slightly different, but I agree that in normal setups, these being the default are not needed.

rbott:
Nice to hear from you @iustin :-)

lib/cli_opts.py
saschalucas: I have tried to move `WarnDeprecatedOption` to `lib/cli.py`, but ended up with an circular import error, because `lib/cli.py` also imports `cli_opts`:
```
# make
PYTHONPATH=. ./autotools/run-in-tempdir \
  /tmp/022a060c4e2e293f642f2a46006698f849e40834/saschalucas_colon_name_ip_check-tmp/./autotools/build-bash-completion --compact > doc/examples/bash_completion
  File "/tmp/022a060c4e2e293f642f2a46006698f849e40834/saschalucas_colon_name_ip_check-tmp/./autotools/build-bash-completion", line 50, in <module>
    from ganeti import cli
  File "/tmp/gntbuild.eKSZ4Lo2/ganeti/cli.py", line 58, in <module>
    import ganeti.cli_opts
  File "/tmp/gntbuild.eKSZ4Lo2/ganeti/cli_opts.py", line 831, in <module>
    action="callback", callback=cli.WarnDeprecatedOption,
                                ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: partially initialized module 'ganeti.cli' has no attribute 'WarnDeprecatedOption' (most likely due to a circular import)
make: *** [Makefile:4502: doc/examples/bash_completion] Error 1
make: *** Deleting file 'doc/examples/bash_completion'
```
Any hints?

iustin: Probably in this case it's better to leave it here…

saschalucas:
> I just ran the QA suite against this PR and it breaks because the QA suite actually tests the name/ip-check behavior (somewhat implicitly).

I have blindly tried to address this issue by keeping as much as possible as before (see commit a230119). @rbott can you try to do a QA run?

rbott:
I tried running it against `xen-pvm` (which is the only option that currently works without any problems unrelated to this PR) and it went well :smile: 

I also checked against kvm, but that fails on bookworm due to hotplug issues. I will do one more run against buster, but I think we are safe to merge. 

I think it would be better to either make a single commit out of this PR or to sqash-merge it into a single commit, what do you think @saschalucas?

rbott (approved):

saschalucas:
> I think it would be better to either make a single commit out of this PR or to sqash-merge it into a single commit, what do you think @saschalucas?

Thanks @rbott. Of cause I'll squash the commits... if I do not forget it again :-).

PR #1591, opened 2021-04-18
add rename function to networks

saschalucas:
Currently it is not possible to rename a network, besides removing and
recreate it with the new name. In order to remove a network, it must not
contain instances and is not connected to any node group. During
removing/recreating the network information like network address/mask,
gateway, link, vlan etc. must be saved externally. With this change
networks can be renamed like most other Ganeti config objects.

### DIFF
--- a/lib/client/gnt_network.py
+++ b/lib/client/gnt_network.py
@@ -322,6 +322,21 @@ def RemoveNetwork(opts, args):
   SubmitOrSend(op, opts)
 
 
+def RenameNetwork(opts, args):
+  """Rename a network.
+
+  @param opts: the command line options selected by the user
+  @type args: list
+  @param args: a list of length 2, [old_name, new_name]
+  @rtype: int
+  @return: the desired exit code
+
+  """
+  network_name, new_name = args
+  op = opcodes.OpNetworkRename(network_name=network_name, new_name=new_name)
+  SubmitOrSend(op, opts)
+
+
 commands = {
   "add": (
     AddNetwork, ARGS_ONE_NETWORK,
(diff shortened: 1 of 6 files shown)


rbott:
Hi @saschalucas , thanks for the PR.

Could you add a test for this to the QA suite? I think it should go here: [/qa/qa_network.py](https://github.com/ganeti/ganeti/blob/master/qa/qa_network.py). 

Other than that it _looks_ fine, but I night some time to go through the details of this PR.

On a sidenote: for completeness we should also provide the rename feature through Ganeti RAPI as `/2/networks/[network_name]/rename` (analogue to the instance rename API endpoint). Should this be handled in a separate PR?

saschalucas:
> Could you add a test for this to the QA suite? I think it should go here: [/qa/qa_network.py](https://github.com/ganeti/ganeti/blob/master/qa/qa_network.py).

Yes, I will get in touch with QA and add a test there (will take some time).

> On a sidenote: for completeness we should also provide the rename feature through Ganeti RAPI as `/2/networks/[network_name]/rename` (analogue to the instance rename API endpoint). Should this be handled in a separate PR?

This is already done. The python unit test don't let you introduce new OPCODES without implementing them in the RAPI server/client. See the second commit 3af7d2b in this PR. Luckily tests are strict, protecting the code base from noobs like me, doing just 50% of the work :smile:

dannyman:
👍🏻 I just hit this use case and would have appreciated this feature! :)

rbott (approved):
LGTM and has passed a QA run. This feature needs to be on the list of "things that need to be addressed in the QA suite" :-)

PR #1594, opened 2021-05-05
Legacy Tekram dc390 scsi emulation support

ksanislo:
Would it be possible to also allow the configuration of the old 'dc390' cards in the official ganeti releases? I've got a bunch of very old freebsd (3.x, 4.x) instances which I've been supporting via ganeti for quite some time, but some of the more recent changes have broken my ability to continue as-is.

Adding this ability to configure things with the "dc390" card is needed since the new QEMU scsi config since 2.16 hasn't been able to be configured via hv/kvm_extra alone like in the older style.


ksanislo:
Sorry about the duplicate PRs. I've reorganized my fork's branches as I've found additional things that require patching. Closing this PR in favor of the branch version.

Issue #1530, opened 2020-10-20
hspace shows different results from textfile and luxid


ironmaya:
With `hspace -L --disk-template=drbd --standard-alloc=20480,2048,2` the output is:
```
The cluster has 16 nodes and the following resources:
  MEM 2058553, DSK 27603812, CPU 824, VCPU 4944.
There are 319 initial instances on the cluster.
Tiered (initial size) instance spec is:
  MEM 32768, DSK 1048576, CPU 12, using disk template 'drbd'.
Tiered allocation results:
  -   1 instances of spec MEM 32768, DSK 1048576, CPU 12
  -   1 instances of spec MEM 32768, DSK 681728, CPU 12
  -   1 instances of spec MEM 32768, DSK 602880, CPU 12
  -   1 instances of spec MEM 32768, DSK 139520, CPU 12
  -   1 instances of spec MEM 32768, DSK 120320, CPU 12
  -   1 instances of spec MEM 32768, DSK 111616, CPU 12
  -   1 instances of spec MEM 32768, DSK 66816, CPU 12
  -   1 instances of spec MEM 32768, DSK 58112, CPU 12
  -   1 instances of spec MEM 32768, DSK 5632, CPU 12
  - most likely failure reason: FailDisk
  - initial cluster score: 16.03100126
  -   final cluster score: 15.43443732
  - memory usage efficiency: 61.00%
  -   disk usage efficiency: 90.99%
  -   vcpu usage efficiency: 34.81%
Standard (fixed-size) instance spec is:
  MEM 2048, DSK 20480, CPU 2, using disk template 'drbd'.
Normal (fixed-size) allocation results:
  -  67 instances allocated
  - most likely failure reason: FailDisk
  - initial cluster score: 16.03100126
  -   final cluster score: 11.50773354
  - memory usage efficiency: 53.34%
  -   disk usage efficiency: 80.39%
  -   vcpu usage efficiency: 35.34%
```

With `hspace -t LOCAL.data --disk-template=drbd --standard-alloc=20480,2048,2` the output is: 

```
The cluster has 16 nodes and the following resources:
  MEM 2058553, DSK 27603812, CPU 824, VCPU 4944.
There are 319 initial instances on the cluster.
Tiered (initial size) instance spec is:
  MEM 32768, DSK 1048576, CPU 12, using disk template 'drbd'.
Tiered allocation results:
  -   1 instances of spec MEM 19456, DSK 1048576, CPU 12
[...]

PR #1595, opened 2021-05-18
hscan should save corrected fixdata

ksanislo:
When [this PR](https://github.com/ganeti/ganeti/pull/34) was merged back in 2017, hscan was updated to look at the new data and present it to stdout, however the change was never applied to the saved copy (LOCAL.data) of the information.

When this incorrect data is reloaded through hbal, under some conditions the invalid memory information can cause it to silently stop considering otherwise viable instance moves.

Correcting hscan to actually save the fixed data for external use, instead of saving the unmodified copy keeps things working as expected.

### DIFF
--- a/src/Ganeti/HTools/Program/Hscan.hs
+++ b/src/Ganeti/HTools/Program/Hscan.hs
@@ -144,7 +144,7 @@ writeDataInner nlen name opts cdata fixdata = do
   hFlush stdout
   when (isJust shownodes) .
        putStr $ Cluster.printNodes nl (fromJust shownodes)
-  writeFile (oname <.> "data") (serializeCluster cdata)
+  writeFile (oname <.> "data") (serializeCluster fixdata)
   return True
 
 -- | Main function.


rbott:
@apoikos @iustin do you have any insights on this? From my rather limited understanding of Haskell, it looks good.

iustin:
Hmm, I'm not sure this is the right thing. The basic question is whether a saved data file will have "corrected" or "uncorrected" data baked in. The code as it is before this commit, seems to keep "uncorrected" data and hbal should correct it on the fly as well, via the `ExtLoader.loadExternalData`, no? Is that not happening?

The patch would change things so that saved data is "corrected", but then we would need to know to prevent all the downstream uses from again applying the correction.

ksanislo:
@iustin The problem is that hbal won't correct data when loading it from a dump.

Please keep in mind that I'm not particularly knowledgeable in Haskell, and it has been 6 months since I looked into this, but in my limited understanding it seemed that the data gets 'fixed' only when it is missing. By saving the original uncorrected data, when that dump gets reloaded in the future, it now has what seems to be valid data in those fields. Since the saved LOCAL.data file doesn't specify if nodes are Xen vs KVM or even something else, there seems to be no way to determine that a specific datapoint would even need changed on reloading. In this case, trusting the saved file content happens automatically, so ensuring the corrected data is what gets written into the file seems like the proper choice.

saschalucas:
In a debugging session with @thielema (with the help of PR #1849) it turns out, that this PR seems valid. An example diff of `hscan` is:

```
--- /tmp/writedatainner-cdata.hs        2025-05-21 17:33:48.588000000 +0200
+++ /tmp/writedatainner-fixdata.hs      2025-05-21 17:33:57.964000000 +0200
@@ -58,46 +58,46 @@
               , alias =
                   "master.38293f9e9d7563f67f87b536dc174b47aeae7da6.gnt.test"
               , tMem = 7939.0
-              , nMem = 747
-              , iMem = 0
+              , nMem = 4096
+              , iMem = 1972
               , fMem = 7192
               , fMemForth = 7192
-              , xMem = 0
+              , xMem = -5321
               , tDsk = 30716.0
               , fDsk = 29736
               , fDskForth = 29736
               , tCpu = 8.0
               , tCpuSpeed = 1.0
               , nCpu = 8
-              , uCpu = 8
-              , uCpuForth = 8
+              , uCpu = 12
+              , uCpuForth = 12
               , tSpindles = 1
               , fSpindles = 0
               , fSpindlesForth = 0
-              , pList = []
-              , pListForth = []
-              , sList = []
-              , sListForth = []
+              , pList = [12, 8, 7, 3]
+              , pListForth = [12, 8, 7, 3]
+              , sList = [10, 2, 1]
+              , sListForth = [10, 2, 1]
               , idx = 0
-              , peers = []
-              , failN1 = True
+              , peers = [(2, 660), (3, 620), (1, 552)]
+              , failN1 = False
               , failN1Forth = True
-              , rMem = 0
+              , rMem = 660
               , rMemForth = 0
-              , pMem = 0.0
+              , pMem = 0.23567199899231642
               , pMemForth = 0.9059075450308604
               , pDsk = 0.968094804010939
               , pDskForth = 0.968094804010939
-              , pRem = 0.0
+              , pRem = 8.31338959566696e-2
               , pRemForth = 0.0
-              , pCpu = 1.0
-              , pCpuForth = 1.0
+              , pCpu = 1.5
+              , pCpuForth = 1.5
               , mDsk = 0.0
               , loDsk = 0
               , hiCpu = 32
               , hiSpindles = 32.0
-              , instSpindles = 0.0
-              , instSpindlesForth = 0.0
+              , instSpindles = 7.0
+              , instSpindlesForth = 7.0
               , offline = False
               , isMaster = True
               , nTags = []
```

With this PR applied, `hbal --print-nodes` and `hcheck` leads to the same states/moves/scores etc. regardless of the source via live data `-L` or from saved data `-t LOCAL.data`.

It is likely that this will also resolve #1530.

saschalucas (approved):
Thanks @ksanislo and sorry for the delay. LGTM.

PR #1596, opened 2021-05-18
reallow dc390 scsi card emulation choice with kvm

ksanislo:
With the advent of ganeti 2.16's revamped scsi configuration, we lost the previous ability to configure a VM with a scsi card as 'dc390' through the use of the kvm_extra parameter alone.

I have existing legacy production infrastructure deployed via freebsd (3.x and 4.x) on ganeti which requires this SCSI card to operate. I would like to have the 'dc390' card put back in as a fully fledged scsi card choice that can be configured in the proper manner.


robbat2 (approved):

rbott:
Hi @ksanislo,

could you rebase the patch on the current master and force-push it to your branch/fork? That would make automated testing easier. Thanks :-)

ksanislo:
@rbott I've rebased my branch. Let me know if you need anything else, thanks!

rbott:
This branch works fine against QA suite. I'll do some more tests and then will move forward and merge it if nothing serious happens :-)

(sorry for the delay!)

rbott:
@ksanislo Thanks for updating/rebasing :-)

@saschalucas I guess we don't need upgrade code here but we do need downgrade code so that `dc390` gets replaced with a valid value _if_ someone decides to downgrade a cluster after configured `dc390` for an instance (or as cluster default), right?

PR #1609, opened 2021-09-23
Add information on Ganeti package sources to docs

rbott:
This picks up where #44 left and adds links to package sources to the INSTALL guide.


rbott:
Thanks @saschalucas, I fixed the line-width and will merge that now.

PR #1612, opened 2021-10-08
Fix unsupported keymap include in >=qemu-4.0

saschalucas:
[Thomas Rieschl reported on ganeti mailinglist](https://groups.google.com/g/ganeti/c/8BouDdnUpd0/m/dLNznGjgBAAJ), that VNC keymap is broken due to Qemu>=4.0 does no longer support keymap include syntax:

`kvm: -vnc :5100: keymap include files are not supported any more`

The initial commit 4f580fef2e3eb3fec8223dad174804e6b5887d7b for VNC keymap support does not have further explanations, why this strategy with `include` was chosen. The comments implies, that there must have been problems with wrong keymaps, that can later not be found. Today Qemu refuses to start if it can't find the supplied keymap:

`qemu-system-x86_64: -vnc :5100: could not read keymap file: 'doof'`

So I decided to drop keymap-include entirely. Live migration from Qemu-3.x with old style `-k /path/to/file_with_include` to Qemu-5.x with new style `-k <map>` works as expected.

[1] https://groups.google.com/g/ganeti/c/8BouDdnUpd0/m/dLNznGjgBAAJ

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1884,7 +1884,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
 
     # the VNC keymap
     keymap = conf_hvp[constants.HV_KEYMAP]
-    kvm_cmd.extend(["-k", keymap])
+    if keymap:
+      kvm_cmd.extend(["-k", keymap])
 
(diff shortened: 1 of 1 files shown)


saschalucas:
@rbott friendly ping && PTAL

PR #1614, opened 2021-10-13
relax VLAN check with VETHs

### LINKED ISSUE #1533: bridge_vlan_add check for VLAN interface doesn't allow for veth interfaces.
I have a veth pair (ve-gw-0@ve-vm-0) with one end (ve-vm-0) associated with a bridge (br-vm). ve-vm-0 has two VLANs, 1 and 2.
```
~# bridge vlan show dev ve-vm-0
port    vlan ids
ve-vm-0  1 PVID Egress Untagged
         2
```

I created a traditional VLAN interface to handle packets associated with VLAN 2:
```
ip link add link ve-gw-0 name ve-gw-0.2 type vlan id 2
```

When I add instances to VLAN 2 I receive the following error:
````
#/usr/lib/ganeti/net-common 
88:      echo "VLAN ${VID} is in use by interface ${vlan_interface}"
````
Despite the comment below, the veth device (ve-gw-0.2) functions correctly when I disable the check. The tap interfaces added to the bridge are able to communicate with each other and the IP associated with ve-gw-0.2.
```
  # when the VID is in use by traditional VLAN interface, then the
  # tap on the VLAN aware bridge will not receive traffic
```

Here is a small diagram of the setup:
```
                       +--------------------------------+
                       |             br-vm              |
+-------------+        |  +-----------+                 |
|  ve-gw-0    +-----------+  ve-vm-0  |                 |
|             |        |  | VLANs: 1,2|                 |
+------+------+        |  +-----------+                 |
       |               |                                |
       |               +--------------------------------+
+------+------+
|  ve-gw-0.2  |
|             |
+-------------+
```
I'm not sure what's the best way to handle this configuration. Maybe adding a force=True option to the vlan= option to bypass the check?

saschalucas:
Normally one can not use identical VLAN on VLAN aware bridge and traditional VLAN interfaces connected to the same lower device. A common workaround is to insert a veth pair between the VLAN aware bridge and traditional VLAN interfaces. This way one can stack VLAN aware bridge and traditional bridges, too. This commit relaxes the VLAN check, to be skipped, if the lower interface is of type `veth`.

This fixes #1533

### DIFF
--- a/tools/net-common.in
+++ b/tools/net-common.in
@@ -76,9 +76,15 @@ function bridge_vlan_add {
   # tap on the VLAN aware bridge will not receive traffic
   if [ -r /proc/net/vlan/config ]; then
     local vlan_interface="$(awk -F '[| ]*' -v vlan="^${VID}$" 'match($2, vlan) { print $1 }' /proc/net/vlan/config)"
+    local lower_devs="$(awk -F '[| ]*' -v vlan="^${VID}$" 'match($2, vlan) { print $3 }' /proc/net/vlan/config)"
     if [ -n "${vlan_interface}" ]; then
-      echo "VLAN ${VID} is in use by interface ${vlan_interface}"
-      exit 1
+      for i in ${lower_devs}; do
+        # allow bridge stacking and vlan overlap for veth devices
+        if [ "$(ip -o link show dev ${i} type veth | wc -l)" -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


rbott:
It looks like it doesn't break anything. Therefore, LGTM. Please proceed :-)

rbott (approved):

PR #1618, opened 2021-10-20
fix KVM's HotplugSupported() method 

rbott:
HotplugSupported() tries to access the `version` property of the QMP instance. However, it only exists _after_ a QMP connection has been established, which is not the case when this method is the first interaction with the QMP socket.

I could reproduce the error with issuing the following on a freshly started instance:
```
gnt-instance modify --net -1:add --hotplug test-instance.example.org
```

Usually QMP functions are accessed through methods of the QmpConnection class which always use the decorator `@_ensure_connection` which takes care of opening the QMP socket if not yet done. This simple fix makes sure HotplugSupported() always has a working QMP connection available.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2151,8 +2151,9 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     @raise errors.HypervisorError: in one of the previous cases
 
     """
+
     try:
-      version = self.qmp.version
+      version = self.qmp.GetVersion()
     except errors.HypervisorError:
       raise errors.HotplugError("Instance is probably down")
 
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -702,6 +702,18 @@ class QmpConnection(MonitorSocket):
     if "netdev_add" not in self.supported_commands:
       _raise("netdev_add qmp command is not supported")
 
+
+  @_ensure_connection
+  def GetVersion(self):
+    """Return the QMP/qemu version field
+
+    Accessing the version attribute directly might result in an error
+    since the socket might not be yet connected. This getter method
+    uses the @_ensure_connection decorator to work around this problem.
+    """
+    return self.version
+
+
   @_ensure_connection
   def HasDynamicAutoReadOnly(self):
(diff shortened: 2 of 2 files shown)


saschalucas:
> Actually there is not much cleanup going on right now. Will the Python garbage collector simply close the socket when the `KVMHypervisor` instance (and the `QmpConnection` instance along with it) get destroyed?

Thanks, implicit connection close must be the case, else I would expect further problems like in #1536?
 
> There are other examples where `KVMHypervisor` connects and uses the `QmpConnection` e.g. through the `Execute()` method and also does not take care of closing anything after that.

It seems that `GetInstanceInfo` is the only one left using `QmpConnection.Execute()`?

What about a minimal compromise adding a explicit `close()`? This way the fix stays lightweight and does not rely on implicit behavior.

Anyway, I already approved this PR, so let's fix this.

rbott:
Sorry for the hassle, I changed the implementation a bit :-D I think this is now a better/more consistent way of doing things: Instead of directly accessing the `version` attribute I have introduced a `GetVersion()` method which makes use of the @_ensure_connection decorator. I will run it through the QA suite and update this post with the result.

saschalucas:
> I have introduced a `GetVersion()` method which makes use of the @_ensure_connection decorator.

Looks good. LGTM!

saschalucas:
I'm removing the label "Backport to 3.0" here, because this PR relies on PR #1546.

#1546 itself has a dependency to >=qemu-2.12, which indirectly obsoletes the `HotplugSupported()` check for >=qemu-1.7.

I'm not sure, how to read/get the qemu-Version in CentOS-7. According to [this](https://pkgs.org/search/?q=qemu) I read `qemu-kvm-1.5.3-175.el7_9.3.x86_64.rpm`. However this seems an additional reason not to backport this to 3.0.

jfut:
In CentOS 7, we use the QEMU provided by the [centos-qemu-ev](https://wiki.centos.org/SpecialInterestGroup/Virtualization) repository when using Ganeti 3.0 beacause the QEMU provided by the default repository is very old (qemu-1.5.3).

- Latest: `qemu-kvm-ev-2.12.0-44.1.el7_8.1.x86_64.rpm`
  - http://mirror.centos.org/centos/7/virt/x86_64/kvm-common/Packages/q/

PR #1622, opened 2021-11-07
Qemu-Guest-Agent (QGA): use implicit PCI slot

saschalucas:
According to `doc/design-hotplug.rst`, explicit PCI slot management via KVM runtime is for Disks and NICs only (see also comment on `update_hvinfo()`). Current PCI slot assignment is calculated from KVM runtime, not live query. This way, misuse of PCI slots, like QGA, are not detected. While QMP `query-pci` would cover virtio-blk-pci and virtio-net-pci, it does not cover the SCSI disk case. Keeping the fix simple, let QGA use an implicit Qemu managed PCI slot (like spice agent virtio-serial).

This fixes #1620

PR #1624, opened 2021-11-11
Allow to run QA suite's instance tests with alternating HV parameters

rbott:
The QA suite currently does not test different values to hypervisor parameters (e.g. `use_chroot`, `use_guest_agent` etc.). This PR implements automatic cycling through values of boolean parameters and parameters which values are known to Ganeti (e.g. `disk_cache`).

It does not cover parameters which accept custom values and it also does not cover parameters which need to be set along with others (e.g. `security_model` and `security_domain`). However, this could be implemented on top of this patch later on. 

The hvparam tests have their own switch in the QA suite configuration as they run **all** enabled instance tests (e.g. create/migrate/hotplugging etc.) **for each hypervisor parameter**, which is very time consuming. If they are disabled the QA suite will fall back to the old behaviour and only run instance tests with the default parameters.

Currently, only KVM HV parameters are used, but it is also possible to add checks for Xen/LXC parameters.

@saschalucas This patch correctly uncovers the bug you are trying to fix in #1622. Once this patch gets merged, I can have the QA suite verify if our PR fixes the problem without any sideeffects.

### DIFF
--- a/qa/ganeti-qa.py
+++ b/qa/ganeti-qa.py
@@ -825,65 +825,91 @@ def RunInstanceTests():
         qa_config.IsTemplateSupported(templ)):
       inodes = qa_config.AcquireManyNodes(num_nodes)
       try:
-        instance = RunTest(create_fun, inodes)
-        try:
-          RunTestIf("instance-user-down", qa_instance.TestInstanceUserDown,
-                    instance)
-          RunTestIf("instance-communication",
-                    qa_instance.TestInstanceCommunication,
-                    instance,
-                    qa_config.GetMasterNode())
-          RunTestIf("cluster-epo", qa_cluster.TestClusterEpo)
-          RunDaemonTests(instance)
-          for node in inodes:
-            RunTestIf("haskell-confd", qa_node.TestNodeListDrbd, node,
-                      templ == constants.DT_DRBD8)
-          if len(inodes) > 1:
-            RunTestIf("group-rwops", qa_group.TestAssignNodesIncludingSplit,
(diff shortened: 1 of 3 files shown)


saschalucas (approved):
Thanks @rbott for your work on the QA suite. Please LGTM and merge your changes. If possible take a look at AssertClusterHvParameterModify(), there seems a longer than 80 char  line in call to AssertCommand().

In future, to save some iterations, it can be an option to turn on as many options as possible (not only boolean one, like a Christmas tree) and only iterate trough those, that can't be used at the same time (i.e. vnc/spice).

rbott:
> Thanks @rbott for your work on the QA suite. Please LGTM and merge your changes. If possible take a look at AssertClusterHvParameterModify(), there seems a longer than 80 char line in call to AssertCommand().

Wow, I always manage to slip in at least one overlength line :roll_eyes: :disappointed:
A fix for that has been pushed.

> In future, to save some iterations, it can be an option to turn on as many options as possible (not only boolean one, like a Christmas tree) and only iterate trough those, that can't be used at the same time (i.e. vnc/spice).

Yes of course - the default configuration of the testing environment also plays a great role here. _Actually_ we would also need to "mix and match" different combinations of parameters but then again that would take ages run.

Let's start with simple steps :-)

Once this has been merged to master, could you please rebase #1622 @saschalucas? That way, I could run your fix against the modified QA suite, as the QA setup always uses the QA suite code that ships with the given PR. In any case, thanks for reviewing!

Issue #1628, opened 2021-12-17
Migration from 2.15.2 to 3.0.1


zen2:
First thanks to everyone that participate to this project that we are using happily for 10 years now.

We want to upgrade our current cluster that is gentoo based and use currently Ganeti 2.15.2.
There is not any more Ganeti packages definitions (ebuild) on Gentoo Linux.

We have build successfully on an up to date Gentoo linux a Ganeti 3.0.1 version.
We can't build a 2.16.2 version because Haskell toolchain is too much recent.
We are aware that we need to upgrade first to 2.16.2 to be able to upgrade to 3.0.1.

So I want to know if it could be possible to upgrade manually /var/lib/ganeti data from a 2.15.2 to 2.16.2 version in order to be able  to try an upgrade to Ganeti 3.0.1.
We are aware that we can't downgrade if needed but we can rollback the whole system by using "gnt-node evacuate" and reboot one by one nodes on a ready to use system backup present on a second partition.

Do you think it's doable ?

saschalucas:
Hi,

the good news are, that I "think" 2.15, 2.16 and 3.0 are 100% compatible regarding `/var/lib/ganeti/config.data`. So in practice, there is nothing to change expect the version itself.

I have no possibility to test my assumption, because I'm used to build Ganeti versions with distro's Haskell packages, too (which makes multiple Ganeti versions complicated). However with careful doing, it should be possible to upgrade from 2.15 to 3.0 in one step.

Before Ganeti 2.10, only one version can be installed at the same time and this still is true if configured without `--enable-symlinks`. This way you have to do a **manual upgrade** (without `gnt-cluster upgrade --to x.x`). See [docs](https://docs.ganeti.org/docs/ganeti/3.0/html/upgrade.html) section `2.1 and above`. If configured with `--enable-symlinks` at step `5. Install new Ganeti version on all nodes` switching symlinks in `/etc/ganeti` must be added.

I think code preventing upgrade from 2.15 to 3.0 is only in `lib/utils/version.py` `UpgradeRange()`. If I'm right, it can be tuned to support 2.15 as well. Even not running `cfgupgrade` at all and edit `/var/lib/ganeti/config.data` by hand (new config version) should be possible.

All nodes must be upgraded at once (while your instances keep running). If things break, your cluster is not controllable. Things like `gnt-node evacuate` won't work. To get back control, one just has to bring back the old state of `/var/lib/ganeti` and old Ganeti version itself.

[...]

PR #1634, opened 2022-01-10
Drop the `xm` toolstack for the Xen hypervisor

### LINKED ISSUE #1611: xen: drop support for `xm` command
Some years ago the Xen project  deprecated and finally dropped the `xm` command and replaced it with the `xl` command. Ganeti contains code for both commands and also defaults to using `xm`, leading to a bad user experience for users of the Xen hypervisor.

We should:
* drop all code/settings/tests related to `xm`
* remove the `xen_cmd` property
* remove all references to `xm` from the documentation

rbott:
This commits drops support for the `xm` toolstack from the Xen hypervisor implementation in Ganeti. It has been deprecated in Xen version 4.1 and finally dropped in 4.5 (around 2016). There is no need for compability any more.

All of the abstractions built into `XenHypervisor` to support two toolstacks have been removed and unit tests have been updated accordingly. 

This fixes #1611


rbott:
QA Suite for xen-pvm has been triggered [here](https://github.com/ganeti/ganeti/runs/4767456065?check_suite_focus=true)

saschalucas (approved):
Thanks @rbott for this work. Since this removes the undocumented HV parameter `xen_cmd` it should be removed from the config on upgrade, too? ... and on downgrade set to `xl`. Feel free to add it here or later.

It seems my review is the best we can get ATM. So LGTM.

rbott:
Ahhh yes, I kind of forgot the upgrade/downgrade handling. I'll take care of that and update the PR accordingly.

rbott:
@saschalucas I have added the up/downgrade part and successfully upgraded stable-3.0 to 3.1. However, I could not test the downgrade (yet) as it is broken. It runs the `DowngradeSshKeyParams()` Method and ends up with this error:

> The current Ganeti setup is using non-DSA SSH keys, and versions below 2.16 do not support these. To downgrade, please perform a gnt-cluster renew-crypto using the  --new-ssh-keys and --ssh-key-type=dsa options, generating DSA keys that older versions can also use.

That is probably not what is supposed to happen on downgrading from 3.1 to 3.0 :-) I'll try to look into that as well and see if I can do anything about at (possibly in a different PR or at least in a separate commit). If you or anybody else can add anything to the solution, please feel free to proceed :-)

saschalucas:
> @saschalucas I have added the up/downgrade part and successfully upgraded stable-3.0 to 3.1. However, I could not test the downgrade (yet) as it is broken. It runs the `DowngradeSshKeyParams()` Method and ends up with this error:
>
> ```
> Verifying 3.0 present on all nodes
> ...
> Downgrade failed:
> Removing SSH parameters failed:
> The current Ganeti setup is using non-DSA SSH keys, and versions below 2.16 do not support these. To downgrade, please perform a gnt-cluster renew-crypto using the  --new-ssh-  keys and --ssh-key-type=dsa options, generating DSA keys that older versions can also use.
> ```

What I understand from `tools/cfgupgrade.py` is, that all config upgrades are always applied (idempotent). That's why one can upgrade from i.e. 2.15 to 3.0 like in #1628.

Downgrade is supposed to work for one minor version. Because Ganeti 2.16 and 3.0 use the same config model, I assume that:
* downgrade from 3.0 to 2.16 was never implemented/tested (MAJOR version differ)
* non-DSA SSH keys are introduced with 2.16, which is why the code for downgrade to 2.15 is still present

So I suppose, that for Ganeti-3.1 `DowngradeSshKeyParams()`should be removed from `DowngradeAll()`???

rbott:
> So I suppose, that for Ganeti-3.1 `DowngradeSshKeyParams()`should be removed from `DowngradeAll()`???

Oh boy, you are entirely right. Sorry for not seeing that earlier :-) I will remove the code and test again!

rbott:
I am currently running once more a QA for xen-pvm ([Link](https://github.com/ganeti/ganeti/runs/4827912778?check_suite_focus=true)) and will add one more for xen-hvm after that. Could you take a quick look again @saschalucas? Unless you see anything, I guess we are good to go then.

saschalucas:
> Could you take a quick look again @saschalucas? Unless you see anything, I guess we are good to go then.

Thanks @rbott. Upgrade/Downgrade looks good. Please merge your changes.

rbott:
Did some additional (non-functional) cleanup/added documentation updates. Tests & QA are still fine.

Merging now, thanks @saschalucas for the review & input! :-)

PR #1635, opened 2022-01-10
Fix Byte vs String comparison

rbott:
Fixes a leftover from the 2to3 migration. OpenSSL now returns byte objects which need to be decoded to strings for comparison with other strings. Found by @pboguslawski in #1632, thanks!

### DIFF
--- a/lib/tools/common.py
+++ b/lib/tools/common.py
@@ -99,7 +99,7 @@ def _VerifyCertificateStrong(cert_pem, error_fn,
   cert_encoded = OpenSSL.crypto.dump_certificate(OpenSSL.crypto.FILETYPE_PEM,
                                                  cert)
   complete_cert_encoded = key_encoded + cert_encoded
-  if not cert_pem == complete_cert_encoded:
+  if not cert_pem == complete_cert_encoded.decode('ascii'):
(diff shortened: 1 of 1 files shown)


rbott:
I have started a QA run in the KVM environment, output can be found [here](https://github.com/ganeti/ganeti/runs/4767522814?check_suite_focus=true).

saschalucas (approved):
Thanks @rbott. Please fix line longer than 80 chars (maybe in a way as suggested by @pboguslawski, doing the `decode()` after concatenation directly on the RHS of the comparison? it shouldn't make any differences?).

Do I correctly assume QA is failing because of an other reason? Other than that, LGTM.

rbott:
Oh boy. I'll put up a tip jar and throw in 5€ every time I'll hand in a PR with overlong lines. Sorry :-( 

I will update the PR accordingly.

rbott:
Fixed the line length issue. I will run the QA Suite just for the sake of completeness (output [here](https://github.com/ganeti/ganeti/runs/4794970068?check_suite_focus=true)). It should not crash this time, I have disabled the QA step that currently prevents the KVM QA from finishing (need to look into that separately).

Once the QA passes, I'll merge and close this PR.

PR #1603, opened 2021-07-04
work around dynamic auto-ro for live migration

saschalucas:
QEMU 4.0 introduced dynamic auto-read-only for file-backed drives. Since then, live migration will break, when the qemu process does not run as root (aka security_model=user|pool):
```
==> /var/log/ganeti/kvm/test.vm.log <==
kvm: Could not open '/var/run/ganeti/instance-disks/test.vm:0': Permission denied
```
Disable auto-read-only if supported by the -blockdev help message. This allows migrations back and forth between qemu versions that support auto-read-only and those that do not. See also commit 6d6adbb7187b4130c74e877e114222f039d35e11 which solves a similar problem for hotpluged disks.

This fixes #1601

PR #1637, opened 2022-01-16
Backports fixes from the last months to stable-3.0

rbott:
This backports all kinds of bugfixes from current master to the stable-3.0 branch in preparation of releasing a 3.0.2 version.


rbott:
BTW we can keep this open until all currently open PRs with "backport-to-3.0" tag have been resolved (either merged or rejected).
Currently these are: #1622, #1614 and #1603

saschalucas:
Thanks @rbott for pushing towards a 3.0.2 release. Two things before my final check:

1. while switching to the backports model, we agreed on @iustin's hint to cherry-pick with `-x` to indicate which commit this change was cherry-picked from. Do I see it right, that this is missing here?
2. there are some open PRs I like to see in 3.0.2, but actually not in this PR:
* #1603 seems very important for working live migration with `security_model=user|pool` **I see no reason for not merging, please review**
* #1614 just a small fix for VLAN-aware bridge users (like me) **works for me :smile:, please review**
* maybe #1622 (but let me double check again what's happening during live migration between a instance started by the unpatched version and live migrated by the patched one).

saschalucas:
>     * maybe [Qemu-Guest-Agent (QGA): use implicit PCI slot #1622](https://github.com/ganeti/ganeti/pull/1622) (but let me double check again what's happening during live migration between a instance started by the unpatched version and live migrated by the patched one).

I've checked it. It's safe for live migration. The QGA pci slot is stable during migration and changed only after reboot with patched Ganeti. Please review #1622, to include it here.

rbott:
Ah yes, I forgot to use `-x` :-( I will fix that asap!

saschalucas:
so next step is #1603, this seems the most anoying bug people struggling with in Bullseye (see ganeti mailing list).

saschalucas:
> so next step is #1603, this seems the most anoying bug people struggling with in Bullseye (see ganeti mailing list).

Just for further explanation: The problem solved in #1603 can not be discovered by testing (single??) live migration, because the migration will succeed. Sadly the instance becomes paused and is frozen. In order to cover this case, i.e. in the QA-suite, additional liveliness probes like ICMP pinging the instance must be done.

rbott:
I have also pulled in #1646, #1647 and #1648

rbott:
So, there's only #1614 left now.

@apoikos if you put on your Debian hat, is there anything that shouldn't go into 3.0 (or which _should_ go into 3.0 but is not yet in this PR)?

saschalucas:
> So, there's only #1614 left now.

It would be no problem, if we decide not to backport it. It's a rare use case, affecting me and @newellz2 in #1533 (more precise I learned from there, how VLAN aware an traditional bridges can coexist within the same VLAN). OTOH VLAN aware bridge support is new in 3.0 and probably not wildly used.

> @apoikos if you put on your Debian hat, is there anything that shouldn't go into 3.0 (or which _should_ go into 3.0 but is not yet in this PR)?

Going through the labels "Backport to 3.0" I can't find #1615, which seems debian related. @apoikos included it in his patches: https://metadata.ftp-master.debian.org/changelogs//main/g/ganeti/ganeti_3.0.1-3_changelog

rbott:
> > So, there's only #1614 left now.
> 
> It would be no problem, if we decide not to backport it. It's a rare use case, affecting me and @newellz2 in #1533 (more precise I learned from there, how VLAN aware an traditional bridges can coexist within the same VLAN). OTOH VLAN aware bridge support is new in 3.0 and probably not wildly used.

If reviewed it and it looks good to me :-) Actually all of our Ganeti 3 clusters use vlan aware bridging now, thanks for that feature!

> 
> Going through the labels "Backport to 3.0" I can't find #1615, which seems debian related.

Oh, that one slipped through. I'll ad it!

saschalucas (approved):
Thanks @rbott. I have nothing else to add. Please decide on your own, if you want to backport #1614.

Let's make 3.0.2.

rbott:
I ran some tests today (both with the QA suite and manually) and so far everything looks fine, apart from #1649. I am not sure if we should postpone a 3.0.2 release any longer because of that issue. The current workaround would be to try and run the hotplug command again, but it does not break anything essential.

What do you think @saschalucas? If you agree, I will merge this PR and we can prepare a release.

saschalucas:
> What do you think @saschalucas? If you agree, I will merge this PR and we can prepare a release.

I agree with merging this PR. #1649 can be solved later. If you think, Bugfix #1652 should be included, feel free ...

LGTM.

rbott:
@saschalucas could you take a final look again? After that, this PR will be merged at last!

saschalucas (approved):
Thanks @rbott! Please merge your changes and LGTM.

PR #1638, opened 2022-01-19
Adjust for Pyparsing 3.0.

mbakke:
`operatorPrecedence` was renamed to `infixNotation` in pyparsing 1.5.7 (from 2012) and finally removed in 3.x.

Please also backport to 3.0.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -38,7 +38,7 @@ Before installing, please verify that you have the following programs:
 - `Python OpenSSL bindings <https://www.pyopenssl.org/>`_
 - `simplejson Python module <https://simplejson.readthedocs.io/>`_
 - `pyparsing Python module <https://pyparsing-docs.readthedocs.io/>`_, version
-  1.4.6 or above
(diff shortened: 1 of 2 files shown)


saschalucas (approved):
Thanks @mbakke. ATM there is no Debian based distro with >=pyparsing-3.0, so we didn't noticed this breakage yet. Good to have you, detecting such things early :+1:

Backport to 3.0 is OK for Ubuntu 18.04 and Debian Buster, both with pyparsing-2.2.0, but would break CentOS 7 with pyparsing-1.5.6. What does @jfut think?

//Edit: Forget my CentOS 7 confusion. Python3 is from EPEL repo, there it seems pyparsing-2.4.0

jfut:
@saschalucas
Thanks for the confirmation. Our RPMs for CentOS 7 since Ganeti 3.0 depend on python36-pyparsing (currently: 2.4.0-1.el7) from EPEL, so that should be fine :+1:

PR #1639, opened 2022-01-23
Improve Ganeti QA runtime/logging

rbott:
This PR addresses recent problems with the QA suite:
* it improves where #1626 and #1624 left:
  * it is not in all cases desirable to iterate through all defined values of a hypervisor parameter (e.g. KVM's `disk_type`)
  * there was an unnecessary step added to the test loop
  * the instance tests that are run for each hypervisor parameter have been reduced to a useful minimum which saves several hours of QA runtime
  * fix resetting the hypervisor parameter to its default, when the default was `None`
* it creates a separate log file ($log_path/qa-profile.log) which only stores the test names and their duration to ease analyzing of the QA suite's performance

### DIFF
--- a/qa/ganeti-qa.py
+++ b/qa/ganeti-qa.py
@@ -64,10 +64,13 @@ from ganeti import utils
 from ganeti import rapi # pylint: disable=W0611
 from ganeti import constants
 from ganeti import netutils
+from ganeti import pathutils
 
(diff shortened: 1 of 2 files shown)


saschalucas (approved):
Thanks @rbott for your continued work on the QA suite! LGTM.

PR #1640, opened 2022-01-23
Warn users that changing DRBD parameters does not affect running instances

### LINKED ISSUE #781: Warn users about the resync rate not being automatically updated
Originally reported of Google Code with ID 725.

```
When the resync rate of DRBD is updated using gnt-cluster modify, the resync rate of existing instances remains the same until the disks are reestablished.

Either warn users of this behavior (easy fix), or make gnt-cluster activate the disks to force the new resync rate to be used (the right thing to do).
```
Originally added on 2014-02-20 10:54:53 +0000 UTC.

rbott:
When a user changes DRBD parameters (e.g. sync-related parameters) they will be only used in future DRBD setups and not applied to currently existing ones. This only adds a warning. A better solution would be to have `activate-disks` apply changed parameters where possible and either add this information to the warning message or remove it and run activate-disks automatically for all existing devices.

Partially fixes #781

### DIFF
--- a/lib/cmdlib/cluster/__init__.py
+++ b/lib/cmdlib/cluster/__init__.py
@@ -1272,6 +1272,16 @@ class LUClusterSetParams(LogicalUnit):
     self._CheckDrbdHelper(vm_capable_node_uuids,
                           drbd_enabled, drbd_gets_enabled)
 
+    if (self.op.diskparams is not None and
+            constants.DT_DRBD8 in self.op.diskparams):
+      self.LogWarning("Changing DRBD parameters only affects devices created "
+                      "in the future, not existing ones")
+      self.LogWarning("You need to shutdown and start (not reboot!) existing "
+                      "instances to adopt the changes")
+      self.LogWarning("Alternatively you can swap the secondary node by "
+                      "running `gnt-instance replace-disks --new-secondary "
+                      "$instance`")
+
     # validate params changes
     if self.op.beparams:
       objects.UpgradeBeParams(self.op.beparams)


rbott:
Started QA suite run here: https://github.com/ganeti/ganeti/runs/4914451814?check_suite_focus=true

saschalucas (approved):
Thanks @rbott. The issue addressed here seems known to me somehow. However I've verified and like to add, that the following will update DRBD parameters for existing instances/disks:
* a cycle of `gnt-instance shutdown` && `gnt-instance start` (`gnt-instance reboot` won't be enough, maybe this does not run a `deactivate/activate disks` cycle)
* doing a `gnt-instance replace-disks` (just quick tried `--new-secondary`)

If you like, you may add some more information to this warning. Otherwise LGTM.

rbott:
Thanks @saschalucas, I have added your suggestions to the warning output. Should we move forward with #781 and close that issue or should we keep it open until someone decides to implement re-syncing the settings with `activate-disks` to existing devices? I don't know if that is even possible for all parameters, tbh.

saschalucas (approved):
Thanks @rbott. Please proceed and merge your changes. I'm also in favor for closing the linked issue. Eventually in future times, we can construct a more modern approach like a state machine, that can apply a desired state to DRBD disk parameters. I'm doing this regular under the hood via `drbdsetup`.

zen2:
We do it under the hood too on on our cluster.
In fact you can modify drbd parameters online.
The only important thing is to identify correct drbd minor on each node.

Issue #1621, opened 2021-11-03
Nic hotplug fails when running qemu chrooted.


glance-:
When running a kvm instance with use_chroot=True and trying to hotplug a nic this happens:
```
Wed Nov  3 17:11:38 2021  - WARNING: Could not hotplug device: Error while executing backend function: kvm: error executing the device_add command: failed to find romfile "efi-virtio.rom" (GenericError):
```

One idea how this could be solved would be using rom_bar=off when hotplugging nic's in a use_chroot=True enviorment. There might be some other way to pass in the rom to the chroot.

saschalucas:
Confirmed.

Don't know if NIC hotplug and chroot ever worked? The man page QEMU.1(1) and `--help` does not show anything regarding `rom_bar` or `rombar`. From previous experience with NICs ROM size mismatch between different qemu versions, I suspect care must be taken to not break live migration and switching rom=on|off in combination with hotplug.

PR #1644, opened 2022-01-27
Workaround for NIC hotplugging with chrooted qemu and QA suite fix for ACPI/hotplugging

rbott:
This PR addresses two problems:

* hotplugging does not work when ACPI is disabled - as this might be the case during the QA runs (e.g. due to alternating through hvparams), the QA suite now only runs hotplug tests when ACPI is enabled for the guest
* while fixing the above I stumbled on another problem: when QEMU runs chrooted, one can not hotplug a NIC because qemu tries to load a PXE rom file which is not present in the chroot (`failed to find romfile "$depends_on_the_driver.rom"` error message). When chrooting is enabled, QMPs `device_add` will now be instructed to not load a rom file from disk

I guess the latter is more of a workaround. Either Ganeti should setup the chroot environment correctly (e.g. provide required rom files inside the chroot) or have the actual hotplug command fail early with a note that hotplugging nics is not possible while running qemu in a chroot. OTOH I am not sure how relevant PXE roms are when NICs are hot-added to already running systems.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2291,12 +2291,14 @@ class KVMHypervisor(hv_base.BaseHypervisor):
       self.qmp.HotAddDisk(device, kvm_devid, uri, drive_add_fn)
     elif dev_type == constants.HOTPLUG_TARGET_NIC:
       kvmpath = instance.hvparams[constants.HV_KVM_PATH]
+      is_chrooted = instance.hvparams[constants.HV_KVM_USE_CHROOT]
       kvmhelp = self._GetKVMOutput(kvmpath, self._KVMOPT_HELP)
       devlist = self._GetKVMOutput(kvmpath, self._KVMOPT_DEVICELIST)
       features, _, _ = self._GetNetworkDeviceFeatures(up_hvp, devlist, kvmhelp)
       (tap, tapfds, vhostfds) = OpenTap(features=features)
       self._ConfigureNIC(instance, seq, device, tap)
-      self.qmp.HotAddNic(device, kvm_devid, tapfds, vhostfds, features)
+      self.qmp.HotAddNic(device, kvm_devid, tapfds, vhostfds, features,
+                         is_chrooted)
       utils.WriteFile(self._InstanceNICFile(instance.name, seq), data=tap)
 
     self._VerifyHotplugCommand(instance, kvm_devid, True)
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -465,7 +465,8 @@ class QmpConnection(MonitorSocket):
     return ret
 
   @_ensure_connection
-  def HotAddNic(self, nic, devid, tapfds=None, vhostfds=None, features=None):
+  def HotAddNic(self, nic, devid, tapfds=None, vhostfds=None, features=None,
+                is_chrooted=False):
     """Hot-add a NIC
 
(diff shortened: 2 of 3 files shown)


lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2291,12 +2291,13 @@ def drive_add_fn(filename):
...
     elif dev_type == constants.HOTPLUG_TARGET_NIC:
       kvmpath = instance.hvparams[constants.HV_KVM_PATH]
+      is_chrooted = instance.hvparams[constants.HV_KVM_USE_CHROOT]
       kvmhelp = self._GetKVMOutput(kvmpath, self._KVMOPT_HELP)
       devlist = self._GetKVMOutput(kvmpath, self._KVMOPT_DEVICELIST)
       features, _, _ = self._GetNetworkDeviceFeatures(up_hvp, devlist, kvmhelp)
       (tap, tapfds, vhostfds) = OpenTap(features=features)
       self._ConfigureNIC(instance, seq, device, tap)
-      self.qmp.HotAddNic(device, kvm_devid, tapfds, vhostfds, features)
+      self.qmp.HotAddNic(device, kvm_devid, tapfds, vhostfds, features, is_chrooted)
```
saschalucas, line 2300: this is longer than 80c

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -500,6 +502,7 @@ def HotAddNic(self, nic, devid, tapfds=None, vhostfds=None, features=None):
         fdnames.append(fdname)
 
       arguments.update({
+        "romfile": "",
```
saschalucas, line 505: Did you discover, that an empty romfile is needed, too, when `vhost_net` is enabled?

saschalucas (approved):
Thanks @rbott. Nice work, also solving #1621. I agree, that passing a ROM for hotpluged NIC seems not to make sense during runtime, just for startup/boot. However we are fine here to leave the ROM empty only when chrooted.

Please fix the 80c issue and double check, why the `romfile` is empty, too, for `vhost_net`. Other than that, LGTM. 

Generally the ROM seems useless, as long as there is no intention to boot from PXE? Maybe Qemu (or Ganeti) should not load a ROM, unless explicitly requested (i.e. by `boot_order`)?

saschalucas:
Added label for Backporting to 3.0. Probably only commit 7c0dec9765d59552db4fea8888351ed100501d6d (fixing the chroot-nic-hotplug-issue). ATM the QA suites improvements are not on the backports list.

lib/hypervisor/hv_kvm/monitor.py
rbott: uh, thanks for spotting. This is actually wrong here ('romfile' is only relevant to the device_add structure, not to netdev_add)

lib/hypervisor/hv_kvm/__init__.py
rbott: fixed

rbott:
Thanks @saschalucas. Sorry for the 80c and the leftover "romfile"-parameter. This has been fixed now. I'll wait for all tests to pass and will merge after that :-)

rbott:
> Generally the ROM seems useless, as long as there is no intention to boot from PXE? Maybe Qemu (or Ganeti) should not load a ROM, unless explicitly requested (i.e. by `boot_order`)?

I am not sure if we should derive the need for a boot rom from the `boot_order`. I think qemu automatically falls through to network boot of there is no OS on disk present. But I might be wrong here.

Issue #1645, opened 2022-01-28
Hot-Adding Disks with KVM / disk_cache=writethrough is broken


rbott:
Hot-Adding disks with KVM and `disk_cache` set to `writethrough` fails with the following error:

```
gnt-instance modify --disk -1:add,size=1G --hotplug kvm-test-instance02.staging.ganeti.org
Fri Jan 28 13:47:27 2022 * disk 2, size 1.0G
Fri Jan 28 13:47:27 2022  - INFO: Trying to hotplug device...
Fri Jan 28 13:47:27 2022  - WARNING: Could not hotplug device: Error while executing backend function: kvm: error executing the device_add command: Property 'virtio-blk-device.drive' can
't find value 'disk-bc85ea66-fcf2-4cd3' (GenericError):
Fri Jan 28 13:47:27 2022  - INFO: Continuing execution..
Fri Jan 28 13:47:27 2022  - INFO: Waiting for instance kvm-test-instance02.staging.ganeti.org to sync disks
Fri Jan 28 13:47:27 2022  - INFO: Instance kvm-test-instance02.staging.ganeti.org's disks are in sync
Modified instance kvm-test-instance02.staging.ganeti.org
 - disk/2 -> add:size=1024,mode=rw
 - disk/2 -> hotplug:failed
Please don't forget that most parameters take effect only at the next (re)start of the instance initiated by ganeti; restarting from within the instance will not be enough.
```

These are the relevant noded logs:
```
RunCmd pvs --noheadings --nosuffix --units=m --unbuffered '--separator=|' -opv_name,vg_name,pv_free,pv_attr,pv_size,pv_name
RunCmd lvcreate -Wn -L1024m -ne229d4ae-8405-49d1-9a53-1664db1f5f06.disk2 -i1 gnt /dev/vdb
RunCmd lvs --noheadings '--separator=|' --units=k --nosuffix -ovg_name,lv_name,lv_attr,lv_kernel_major,lv_kernel_minor,vg_extent_size,stripes,devices
RunCmd lvchange -ay /dev/gnt/e229d4ae-8405-49d1-9a53-1664db1f5f06.disk2
RunCmd lvs -o tags --noheadings --nosuffix /dev/gnt/e229d4ae-8405-49d1-9a53-1664db1f5f06.disk2
[...]

PR #1646, opened 2022-01-29
Refactor KVM parameters checks/validation and validate disk_cache/disk_aio

rbott:
This PR moves all current KVM Parameter check/validation logic into separate functions with unit tests.

It also adds new checks for `disk_aio`/`disk_cache` to address #1645. As the `CheckParameterSyntax()` method is also used in cluster-verify-runs, broken configurations should also be detected once a user upgrades to a Ganeti version that ships this code and there should be no separate handling during the cluster upgrade required.

With the new checks in place we could either drop some compatibility code from `_GenerateKVMBlockDevicesOptions()`_or_ add some more code to fix situations like #1645. The latter could be backported to 3.0.x, whereas the improvements that are currently in this PR are probably not suited for a backport.


saschalucas (approved):
Thanks @rbott. Really nice work! Please merge your changes.

Regarding a possible backport to 3.0, I find it not necessary, but I have trouble to see why you think it is "not suited". Can you please spot some light, where the problems would be?

rbott:
> Thanks @rbott. Really nice work! Please merge your changes.

Thanks :-) 

> Regarding a possible backport to 3.0, I find it not necessary, but I have trouble to see why you think it is "not suited". Can you please spot some light, where the problems would be?

On the one hand it would fix/uncover possible misconfiguration for everyone - but on the other hand it would also mean that gnt cluster-verify might suddenly fail after a minor update (because it detects conflicting parameters). Probably the pros outweigh the cons here :-)

rbott:
It seems like the current code still has problems. I initially assumed "disk_cache=default" would work with hotplugging, but that still leads to an error.

e.g. the following two statements work:
```
drive_add dummy file=/dev/fdset/0,if=none,id=disk-d584c5d9-367f-4368,format=raw,cache=none,aio=native
drive_add dummy file=/dev/fdset/0,if=none,id=disk-36d0f64d-505d-4369,format=raw,cache=none
```

...whereas this does not work (with disk_aio set to native and disk_cache set to default):
```
drive_add dummy file=/dev/fdset/0,if=none,id=disk-ed836934-3437-4719,format=raw,aio=native
```

I will look into the PR again under two aspects:
* disk_aio = native should only work with disk_cache = none (this PR allows none _and_ default)
* the code which generates the drive_add string should be revised to only create working parameter sets

I'll update the PR as soon as I find the time to look into it!

saschalucas:
> ...whereas this does not work (with disk_aio set to native and disk_cache set to default):
> 
> ```
> drive_add dummy file=/dev/fdset/0,if=none,id=disk-ed836934-3437-4719,format=raw,aio=native
> ```

A, thanks for spotting this out. Sorry for being late, but i should have known this. From man page qemu-system-x86_64(1):

```
       -drive option[,option[,option[,...]]]
...
         cache=cache
               cache is "none", "writeback", "unsafe", "directsync" or "writethrough"
...
               The default mode is cache=writeback.
```
With `cache=writeback` per default your research in https://github.com/ganeti/ganeti/issues/1645#issuecomment-1024430539 applies: "If another cache mode is used, then the IO mode will silently be switched back to the default" ... by Qemu.

That's why a `drive_add` with `aio=native` fails.

rbott:
I have just made an attempt at using the same code for the evaluation of aio/cache during KVM start and hotplugging. I will run some tests on this branch and see how that turns out. At least the new code is unittested :-)

rbott:
oh well... check-local seems a bit to strict here :-)
> lib/hypervisor/hv_kvm/validation.py: Line 4: Must contain copyright information (Copyright (C) <year>[, <year> ...] Google Inc.)

I guess we need to update/fix the validation to not test for "Google Inc" as the (sole) copyright holder :-D 
However, just switching it to something like "the Ganeti project" will probably break all other files. But that should be fixed in a separate PR, I'll ignore that for now.

saschalucas:
> I guess we need to update/fix the validation to not test for "Google Inc" as the (sole) copyright holder

PR #1648 allows Google and Ganeti project copyright. The new file `lib/hypervisor/hv_kvm/validation.py` needs the following header:

```
#
#

# Copyright (C) 2022 the Ganeti project
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are
# met:
#
# 1. Redistributions of source code must retain the above copyright notice,
# this list of conditions and the following disclaimer.
#
# 2. Redistributions in binary form must reproduce the above copyright
# notice, this list of conditions and the following disclaimer in the
# documentation and/or other materials provided with the distribution.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
# IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
# TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
# PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

saschalucas (approved):
Thanks for your work. Looks good, please merge your changes.

rbott:
So far my testing looks good. However, I keep running into occasional QMP handshake errors while hotplugging disks or nics. I have seen these before, so this has not been introduced with these code changes. I will open up an issue for that as well, but ignore it for this PR. As soon as #1648 has been merged, the tests should pass on this PR and I will merge it.

rbott:
I have pulled in the check-local changes from the current master and now all tests pass. Merging! :-)

PR #1647, opened 2022-01-29
add `check-local` to the github CI

saschalucas:
The `check-local` is part of making a release tarball. It checks i.e. for code lines longer than 80 characters, trailing white space etc. This is rather late, fixing code style at release time.

PR #1470 was the first try to address this issue. However we couldn't agree on several things:
* if the .github folder should be excluded from Makefile
* if the locale problem should be fix generally

Now I leave the above questions open and work around them in the command invocation.

Motivation is: better to have `check-local` in CI than have not.


rbott:
I assume the check-local part needs to run before the other targets, right? Right now it looks like it struggles with __pycache__ folders created during e.g. tests.

saschalucas:
> I assume the check-local part needs to run before the other targets, right? Right now it looks like it struggles with **pycache** folders created during e.g. tests.

Yes, seems so. I moved it before tests.

Also I know now, that github CI uses it's definition from inside the PR branch. That's how one can mine bitcoins in other projects CI. This also means, I need to commit a followup fixing some whitespace etc. to make `check-local ` happy.

rbott (approved):
LGTM, please merge & thanks for picking up where I left :-)

rbott (approved):
LGTM, please merge & thanks for picking up where I left :-)

saschalucas:
This is partially relevant for backport-3.0 to prepare for a minor release. @rbott please pick commit 53d59c9d76eb36c6f3d41f6e4676470d3c3fe2a6 into #1637.

PR #1662, opened 2022-03-05
KVM: Move ballooning and system-shutdown from HMP to QMP

rbott:
This PR addresses #1542 by moving the ballooning and shutdown actions from HMP to QMP.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2037,8 +2037,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
       utils.WriteFile(self._InstanceNICFile(instance.name, nic_seq), data=tap)
 
     if vnc_pwd:
-      change_cmd = "change vnc password %s" % vnc_pwd
-      self._CallMonitorCommand(instance.name, change_cmd)
+      self.qmp.SetVNCPassword(vnc_pwd)
 
     # Setting SPICE password. We are not vulnerable to malicious passwordless
     # connection attempts because SPICE by default does not allow connections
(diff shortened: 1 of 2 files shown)


rbott:
Update: this has passed a full KVM QA run on Debian Bullseye. Buster is still running, but should be fine as well. The VNC settings (or spice, for that matter) settings are currently not tested with the QA suite but I tested it manually of course. I

On a sidenote: I will extend the QA suite to cover Bookworm (current Debian testing) soon.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2433,8 +2432,22 @@ def _GetDefaultMachineVersion(cls, kvm_path):
     else:
       return "pc"
 
-  @classmethod
-  def _StopInstance(cls, instance, force=False, name=None, timeout=None):
+  def _WaitForInstanceToShutdown(self, timeout, name):
+    """Wait for QEMU to shut down after timeout or kill it
```
saschalucas, line 2436: It seems nothing is killed here, so the comment might be to much?

saschalucas (approved):
Thanks @rbott. Other than my inline comment above, I have nothing to add. If you have verified, please merge your changes.

lib/hypervisor/hv_kvm/__init__.py
rbott: oh, that is a leftover comment from an earlier version of the method. I'll fix that and merge the PR.

Thanks for spotting this! :-)

rbott: Fixed

PR #1663, opened 2022-03-05
KVM Instance poweroff Init Script: adapt to current socat parameter usage

### LINKED ISSUE #1629: ganeti-kvm-poweroff socat and timeout bugs
https://github.com/ganeti/ganeti/blob/master/doc/examples/ganeti-kvm-poweroff.initd.in does not shutdown guests in ganeti on debian 11 (works fine on debian 9). Seems that -U socat parameter causes problem in debian 11.

There is also variable name bug in this script that makes its timeout function useless.

Attached please find patch that works fine for us in ganeti on debian 9 and 11:

[ganeti-kvm-poweroff.diff.gz](https://github.com/ganeti/ganeti/files/7795996/ganeti-kvm-poweroff.diff.gz)

rbott:
This fixes #1629 by implementing the patch @pboguslawski suggested. Thanks!

Tested on Debian Bullseye.

### DIFF
--- a/doc/examples/ganeti-kvm-poweroff.initd.in
+++ b/doc/examples/ganeti-kvm-poweroff.initd.in
@@ -27,7 +27,7 @@ do_kvm_poweroff () {
     # shutdown VMs and remove sockets of those not running
     for vm_monitor in $CONTROL_PATH/*.monitor; do
         if ! echo system_powerdown | \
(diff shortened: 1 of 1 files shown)


saschalucas (approved):
thanks @rbott and @pboguslawski. please merge.

PR #1597, opened 2021-05-20
kvm_qmp_timeout hypervisor parameter

ksanislo:
This change makes the kvm hypervisor QMP timeout a settable hypervisor parameter. The default timeout remains the preconfigured 5 seconds, but can now be easily extended for edge cases that create QMP timeouts regularly.

PR #1665, opened 2022-03-30
KVM: repeat QMP shutdown request every 10 seconds

rbott:
When instances start and stop in short order (which happens e.g. during the QA suite tests) the guest usually is not yet able to receive ACPI shutdown events, hence triggering the 120 second shutdown timeout. This expands the QA runtime from a few hours to sometimes > 12 hours.

Ganeti now signals the KVM guest to shut down every 10 seconds until the shutdown timeout has been reached. This avoids the aforementioned scenario (or any other case where the guest might miss the shutdown event). According to some tests, this does not affect guests in negative ways, but it surely needs some in-depth testing.

I will update this PR as soon as I have tested several different guest systems.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2432,20 +2432,6 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     else:
       return "pc"
 
-  def _WaitForInstanceToShutdown(self, timeout, name):
-    """Wait for QEMU to shut down within a given timeout
-
-    """
-    tick = 1
-    while tick <= timeout:
-      _, _, alive = self._InstancePidAlive(name)
-      if not alive:
-        logging.info("KVM: instance %s finished shutdown after %d seconds"
-                     % (name, tick))
-        break
-      tick += 1
-      time.sleep(1)
-
   @_with_qmp
   def _StopInstance(self, instance, force=False, name=None, timeout=None):
     """Stop an instance.
@@ -2466,8 +2452,6 @@ class KVMHypervisor(hv_base.BaseHypervisor):
         utils.KillProcess(pid)
       else:
         self.qmp.Powerdown()
-        if timeout is not None:
-          self._WaitForInstanceToShutdown(timeout, name)
 
     self._ClearUserShutdown(instance.name)
 


lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2467,7 +2453,25 @@ def _StopInstance(self, instance, force=False, name=None, timeout=None):
...
+                           % (name, tick))
+              break
+
+            # re-issue ACPI shutdown events every 10 seconds
+            # this works around instances which did not react to earlier events
+            # because they were e.g. still booting
+            # this _mostly_ addresses the QA suite which starts and stops
+            # instances in short order, sometimes triggering 120 second
+            # default timeout
+            if not tick % 10:
```
apoikos, line 2470: This is a bit of a double negation and needs some brain cycles to parse correctly. I would write it explicitly as
```
if tick % 10 == 0:
```

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2467,7 +2453,25 @@ def _StopInstance(self, instance, force=False, name=None, timeout=None):
       else:
         self.qmp.Powerdown()
         if timeout is not None:
-          self._WaitForInstanceToShutdown(timeout, name)
+          tick = 1
+          while tick <= timeout:
```
apoikos, line 2457: You know `timeout` here, you could replace `while tick <= timeout` with `for tick in range(1, timeout+1)`.

saschalucas:
> Ganeti now signals the KVM guest to shut down every 10 seconds until the shutdown timeout has been reached. This avoids the aforementioned scenario (or any other case where the guest might miss the shutdown event). According to some tests, this does not affect guests in negative ways, but it surely needs some in-depth testing.

Thanks @rbott. I know I've approved #1662 and should have asked there. But ATM I'm a bit confused how shutdown via HMP worked before and how it compares to QMP now.

I know by observation, that via old HMP an instance already receives multiple ACPI power button events. Are the following assumptions right?

* the parameter `timeout` corresponds to CLI option `--shutdown-timeout`?
* the HMP `system_powerdown` is synchronous and returns after the instance is shut down?
* the HMP  `system_powerdown` was responsible for repeating ACPI power button events?
* the QMP Powerdown is asynchronous and returns imitatively?
* that's why for QMP repeating ACPI power button is implemented here?

rbott:
> Thanks @rbott. I know I've approved #1662 and should have asked there. But ATM I'm a bit confused how shutdown via HMP worked before and how it compares to QMP now.

In general, it works the same, although HMP is _way_ slower. See this comment from `_CallMonitorCommand()` in the KVM hypervisor implementation:
```
    # TODO: Replace monitor calls with QMP once KVM >= 0.14 is the minimum
    # version. The monitor protocol is designed for human consumption, whereas
    # QMP is made for programmatic usage. In the worst case QMP can also
    # execute monitor commands. As it is, all calls to socat take at least
    # 500ms and likely more: socat can't detect the end of the reply and waits
    # for 500ms of no data received before exiting (500 ms is the default for
    # the "-t" parameter).
```

If a timeout is given, `_CallMonitorCommand()` simply wraps the call to `socat` with the UNIX `timeout` command.

> 
> I know by observation, that via old HMP an instance already receives multiple ACPI power button events. Are the following assumptions right?

Yes and no. I have looked a bit more into the code and the KVM hypervisor implementation never generated more than one power button event by itself (not in current `master` and also not with HMP). However, the `StopInstance()` Method of the HV implementation gets called from `InstanceShutdown` in `backend.py` and _that_ actually does implement retries which I did not notice before. I will have to review that a bit more and figure out if #1662 actually was wrong here.

> 
> * the parameter `timeout` corresponds to CLI option `--shutdown-timeout`?

More or less, yes. `hv.StopInstance()` might be called through other means, but in the case of the CLI command it corresponds to it.

> * the HMP `system_powerdown` is synchronous and returns after the instance is shut down?
I assumed it to be asynchronous (however, interacting with HMP through `socat` _is_ way slower than using QMP). 

> * the HMP  `system_powerdown` was responsible for repeating ACPI power button events?
No, it was not (see above).

> * the QMP Powerdown is asynchronous and returns imitatively?
That definitely is true.
 

I will review this PR and #1662 again and try to find a correct solution. I am not yet sure how the `timeout` value plays along with the hardcoded retry-count of `5` in `backend.py/ShutdownInstance()`. In theory this should bump the real timeout to `5 * timeout` which is not what I have observed. I'll update this PR as soon as I have come up with something.

Thanks alot for pointing me in the right direction @saschalucas :-)

rbott:
So I did some testing with the "old" code (pre #1662) and figured out that I should have done that earlier. The following happens:

* `InstanceShutdown()` from `backend.py` runs the `hv.StopInstance()` method every 5 seconds until `timeout` is reached (essentially generating the ACPI power button event over and over again)
* `timeout` is propagated to `hv.StopInstance()` and in the case of KVM used to wrap the `socat` call to kill it in case it hangs indefinetly (but usually socat returns after ~500ms, so does `hv.StopInstance()` and gets called again 5 seconds later)

So in theory - if `socat` hangs/blocks and needs to be killed - only one ACPI power button event will be generated. But in practice it does not hang.

With #1662 I made the `hv.StopInstance()` basically blocking and handling the timeout entirely itself and thus changed the behaviour from continuous ACPI power button events to just a single one. 

I will now update this PR to revert to the old behaviour, but with QMP. That means `hv.StopInstance()` will return immediately and actually do nothing at all with `timeout`. In theory, QMP calls _could_ hang (see #1597) but according to [QEMU people](https://github.com/ganeti/ganeti/pull/1597#issuecomment-846567632) this should not happen and would be considered a bug in QEMU (hence, does not need to be handled in Ganeti).

lib/hypervisor/hv_kvm/__init__.py
rbott: You are absolutely right here. However, this part of the code is now obsolete :-)

saschalucas (approved):
Thanks @rbott for your explanation and fixup. Sorry again for overlooking this before in #1662.

I think the issue for the QA suite is solved in the right way now. Also code, where @apoikos had valid objections, is now obsolete.

So please merge your changes.

PR #1667, opened 2022-04-06
KVM: replace -drive by -blockdev and make HMP obsolete

### LINKED ISSUE #1542: Stop using QEMU's human monitor

rbott:
This PR replaces the deprecated `-drive` commandline parameter of QEMU with `-blockdev`. It also switches hotplugging to use `blockdev-add|del` via QMP and hence cuts the last ties from Ganeti to QEMU's human monitor (HMP).

*Please consider this PR work-in-progress!*

This PR also touches several of KVMs parameters:
* ~~`disk_type` `ide` is now named `ide-hd`~~
* ~~`disk_cache` now only accepts `writethrough`/`writeback` (default: `writeback`)~~
* `disk_aio` now also supports `io_uring` when QEMU 5.0+ is present (default: `threads`)
* `disk_discard` now only accepts `ignore`/`unmap` (default: `ignore`)

~~Instead of changing parameters to reflect QEMU (e.g. `ide` -> `ide-hd`) we could also try to keep parameters as they are and just translate in the code. What would you prefer? Regarding the `disk_cache` parameter, there are really only two sensible choices (dicated by the blockdev parameter `cache.direct=on|off`). I think it makes it easier for the user to not have Ganeti offer four different choices when there really are only two.~~

**What has been tested?**
- [x] The code works fine on Debian Bullseye locally with plain disk instances and Hot-Add/Hot-Removal
- [x] The code works fine with the QA suite (3-Node Debian Bullseye clusters with various storage modes + Live Migration & Hotplugging)
  - [x] DRBD
  - [x] File
  - [x] SharedFile
  - [x] Plain
  - [x] RBD (including userspace access)
  - [x] GlusterFS (including userspace access) _Note_: these tests have actually been carried out manually. Not through the QA suite
- [x] hot-adding disks after cluster upgrades work (tested with 3.0 -> 3.1/HEAD with and without instance restarts)
- [x] updates to the documentation (e.g. updated parameters)
- [x] the cdrom code needs to be refactored to use `-blockdev`
- [x] the floppy code needs to be refactored to use `-blockdev`
- [x] migration code for the disk_discard parameter (no more "default" value)

**What is currently missing?**
- [ ] ExtStorage disk type is entirely untested (also see [this post](https://github.com/ganeti/ganeti/pull/1667#issuecomment-1512141687))

@apoikos @iustin @saschalucas it would be great if you can find some time to review so we can try to improve this PR. Once this is done, we can close #1542 and aim for a 3.1 release :-)

### DIFF
--- a/doc/design-3.1.rst
+++ b/doc/design-3.1.rst
@@ -1,0 +1,7 @@
+=================
+Ganeti 3.1 design
+=================
+
+The following designs have been implemented in Ganeti 3.1
+
+- :doc:`design-qemu-blockdev`
--- a/doc/design-qemu-blockdev.rst
+++ b/doc/design-qemu-blockdev.rst
@@ -1,0 +1,175 @@
+======================================
+Implementing the qemu blockdev backend
+======================================
+
+:Created: 2022-Apr-20
+:Status: Implemented
+:Ganeti-Version: 3.1
+
+.. contents:: :depth: 2
+
+This is a design document explaining the changes inside Ganeti while
+transitioning to the blockdev backend of QEMU, making the currently used
+`-drive` parameter obsolete.
+
+
+Current state and shortcomings
+==============================
+
+Ganeti's KVM/QEMU code currently uses the `-drive` commandline parameter to add
+virtual hard-disks, floppy or CD drives to instances. This approach has been
+deprecated and superseded by the newer `-blockdev` parameter which has been
+considered stable with the 2.9 release of QEMU.
+
+Furthermore, the use of `-drive` blocks the transition from QEMU's human
+monitor to QMP, as the latter has never seen an implementation of the relevant
+methods to hot-add/hot-remove storage devices configured through `-drive`.
+
+Currently, Ganeti QEMU/KVM instances support the following storage devices:
+
+``Virtual Hard-disks``
+  An instance can have none to many disks which are represented to guests as the
+  selected `disk_type`. Ganeti supports various device types like `paravirtual`
+  (VirtIO), `scsi-hd` (along with an emulated SCSI controller), `ide` and the
+  like. The disk type can only be set per instance, not per disk.
+
+  A disk's backing storage may be file- or block-based, depending on the
+  available disk templates on the node.
+
+  Disks can be hot-added to or hot-removed from running instances.
+
+  An instance may boot off the first disk when not using direct kernel boot,
+  specified through the `boot_order` parameter.
+
+  Ganeti allows tweaking of various disk related parameters like AIO modes,
+  caching, discard settings etc.
+  Recent versions of QEMU (5.0+) have introduced the `io_uring` AIO mode which
+  is currently not configurable through Ganeti.
+
+``Virtual CD Drives``
+  Ganeti allows for up to two virtual CD drives to be attached to an instance.
+  The backing storage to a CD drive must be an ISO image, which needs to be
+  either a file accessible locally on the node or remotely through a HTTP(S)
+  URL. Different bus types (e.g. `ide`, `scsi`, or `paravirtual`) are supported.
+
+  CD drives can not be hot-added to or hot-removed from running instances.
+
+  Instances not using direct kernel boot may boot from the first CD drive, when
+  specified through the `boot_order` parameter. If the `boot_order` is set to
+  CD, the bus type will silently be overridden to `ide`.
+
+``Virtual Floppy Drive``
+  An instance can be configured to provide access to a virtual floppy drive
+  using an image file present on the node.
+
+  Floppy drives can not be hot-added to or hot-removed from running instances.
+
+  Instances not using direct kernel boot may boot from the floppy drive, when
+  specified through the `boot_order` parameter.
+
+
+Proposed changes
+================
+
+We have to eliminate all uses of the `-drive` parameter from the Ganeti codebase
+to ensure compatibility with future versions of QEMU. With this, we can also
+switch all hotplugging-related methods to use QMP and drop all code relating to
+the human monitor interface.
+
+Ganeti should support the AIO mode `io_uring` as long as the QEMU version on
+the node is recent enough.
+
+Implementation
+==============
+
+I/O Methods
++++++++++++
+
+With QEMU 5.0, support for `io_uring` has been introduced. This should be
+supported by Ganeti as well, given a recent enough QEMU version is present on
+the node. Ganeti will still default to using the `threads` mode on new
+installations.
+
+Disk Cache
+++++++++++
+
+Using the following table found in `man 1 qemu-system-x86_64` we can translate
+the disk cache modes known to Ganeti into the settings required by `-blockdev`:
+
+  .. code-block:: none
+
+    ┌─────────────┬─────────────────┬──────────────┬────────────────┐
+    │             │ cache.writeback │ cache.direct │ cache.no-flush │
+    ├─────────────┼─────────────────┼──────────────┼────────────────┤
+    │writeback    │ on              │ off          │ off            │
+    ├─────────────┼─────────────────┼──────────────┼────────────────┤
+    │none         │ on              │ on           │ off            │
+    ├─────────────┼─────────────────┼──────────────┼────────────────┤
+    │writethrough │ off             │ off          │ off            │
+    ├─────────────┼─────────────────┼──────────────┼────────────────┤
+    │directsync   │ off             │ on           │ off            │
+    ├─────────────┼─────────────────┼──────────────┼────────────────┤
+    │unsafe       │ on              │ off          │ on             │
+    └─────────────┴─────────────────┴──────────────┴────────────────┘
+
+The table also shows `directsync` and `unsafe`, which are currently not
+implemented by Ganeti and may be addressed in future changes. The Ganeti value
+of `default` should internally be mapped to `writeback`, as that reflects the
+values which QEMU assumes when not given explicitly according to the
+documentation.
+
+Construction of the command line parameters
++++++++++++++++++++++++++++++++++++++++++++
+
+The code responsible for generating the commandline parameters to add disks,
+cdroms and floppies needs to be changed to use the combination of `-blockdev`
+and `-device`. This will not change they way a disk is actually presented to
+the virtual guest.
+A small exception will be cdrom: Ganeti used to overwrite the cdrom disk type to
+`ide` when `boot_order` is set to `cdrom`. This is not required any more - QEMU
+will also boot off VirtIO or SCSI CD drives.
+
+Hot-Adding Disks
+++++++++++++++++++
+
+The code needs to be refactored to make use of the QMP method "blockdev-add",
+using the same parameters as for its commandline counterpart (e.g. blockdev node
+id, caching, aio-mode).
+
+Hot-Removing Disks
+++++++++++++++++++
+
+Hot-removing a disk consists of two steps:
+
+- removing the virtual device (or rather: ask the guest to release it)
+- releasing the storage backend
+
+The first step always returns immediately (QMP request `device_del`) but signals
+its `actual` result asynchronously through the QMP event `DEVICE_DELETED`.
+Ganeti currently does not support receiving QMP events - implementing this will
+be out of scope for this change.
+
+In Ganeti releases up to 3.0 the human monitor was used to delete the device.
+Executing commands through that interface was very slow (500ms to 1s) and
+seemingly slow enough to let the following request to remove the drive succeed
+as the guest had enough time to actually release the device.
+With the switch to QMP, both requests will fire in direct succession. Since QEMU
+cannot release a block device (`blockdev-del` through QMP) which is still in use
+by a device inside the guest, hot-removing disks will always fail.
+
+Without support for QMP events, the only feasible way will be to mimic the slow
+human monitor interface and block for one second after sending the `device_del`
+request to the guest.
+
+On upgraded clusters it will **not** be possible to hot-remove a disk before the
+instance has been either restarted or live-migrated (thus "upgrading" all disk
+related parameters to `-blockdev`). Disks added with `-drive` can not be removed
+using `blockdev-dev`.
+
+
+
+.. vim: set textwidth=72 :
+.. Local Variables:
+.. mode: rst
+.. fill-column: 72
+.. End:
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -82,7 +82,10 @@ from ganeti.hypervisor.hv_kvm.validation import check_boot_parameters, \
                                                 validate_machine_version, \
                                                 validate_security_model, \
                                                 validate_spice_parameters, \
-                                                validate_vnc_parameters
+                                                validate_vnc_parameters, \
+                                                validate_disk_parameters
+
+import ganeti.hypervisor.hv_kvm.kvm_utils as kvm_utils
 
 _KVM_NETWORK_SCRIPT = pathutils.CONF_DIR + "/kvm-vif-bridge"
 _KVM_START_PAUSED_FLAG = "-S"
@@ -182,8 +185,9 @@ def _with_qmp(fn):
     return fn(self, *args, **kwargs)
   return wrapper
 
+
 def _GetDriveURI(disk, link, uri):
-  """Helper function to get the drive uri to be used in --drive kvm option
+  """Helper function to get the drive uri to be used in -blockdev kvm option
 
   Invoked during startup and disk hot-add. In latter case and if no userspace
   access mode is used it will be overriden with /dev/fdset/<fdset-id> (see
@@ -439,6 +443,13 @@ def _UpgradeSerializedRuntime(serialized_runtime):
       # Replace the original vnc argument with the new ones
       kvm_cmd[idx:idx+2] = tls_obj_cmd + vnc_cmd
 
+    # with 3.1 the 'default' value for disk_discard has been dropped
+    # and replaced by 'ignore'
+    if constants.HV_DISK_DISCARD in hvparams \
+      and hvparams[constants.HV_DISK_DISCARD] not in \
+        constants.HT_VALID_DISCARD_TYPES:
+      hvparams[constants.HV_DISK_DISCARD] = constants.HT_DISCARD_IGNORE
+
   return kvm_cmd, serialized_nics, hvparams, serialized_disks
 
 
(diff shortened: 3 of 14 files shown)


src/Ganeti/Constants.hs
```diff
@@ -2731,7 +2731,7 @@ htHvmValidVifTypes = ConstantUtils.mkSet [htHvmVifIoemu, htHvmVifVif]
 -- * Disk types
 
 htDiskIde :: String
-htDiskIde = "ide"
+htDiskIde = "ide-hd"
```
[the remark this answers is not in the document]
rbott: You are right, I will change that back!

src/Ganeti/Constants.hs
```diff
@@ -2803,12 +2803,6 @@ htKvmValidScsiControllerTypes =
                        htScsiControllerVirtio,
                        htScsiControllerMegasas]
 
-htCacheDefault :: String
-htCacheDefault = "default"
-
-htCacheNone :: String
-htCacheNone = "none"
```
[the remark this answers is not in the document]
rbott: Yes, this would require upgrade/downgrade code and of course lots of testing.

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -545,74 +546,66 @@ def HotDelNic(self, devid):
...
-
     """
     self.Execute("device_del", {"id": devid})
-    #TODO: uncomment when drive_del gets implemented in upstream qemu
-    # self.Execute("drive_del", {"id": devid})
+
+    # TODO: implement receiving of QMP events
+    # We need to wait for the DEVICE_DELETED event via QMP before proceeding.
+    # The old implementation using HMP only worked because it is so slow.
+    time.sleep(1)
```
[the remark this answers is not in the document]
rbott: Yeah, entirely agree. That was more of a workaround to get things in a working state. OTOH it pretty much reflects what happened before (although not as explicit) and implementing events in QMP is probably way out of scope for this PR.

saschalucas:
Thanks @rbott for your work. Without going into depth, I would prefer to keep old values for instance parameters `disk_type` (ide case) and specially `disk_cache`.

I think it might be not enough to just support/translate `writethrough/writeback` into `cache.direct`. Basically `writethrough` means read-cache-only. Which can't be `cache.direct=on`??? We also need `none` for no-cache.

In `qemu-system-x86_64(1)` there is a nice translation table from `-drive` to `-device` (column 2) and `-blockdev` (columns 3+4) cache options.
```
             │ cache.writeback   cache.direct   cache.no-flush
────────────┼─────────────────────────────────────────────────
writeback    │ on                off            off
none         │ on                on             off
writethrough │ off               off            off
directsync   │ off               on             off
unsafe       │ on                off            on
```
The first column matches Ganeti's `disk_cache`. Maybe we can add `unsafe` for faster Ganeti-QA runs :smile:.

Also I think we can't drop the hassle of valid cache combinations with aio/disk template type. Especially for `rbd`'s feature of coherent distributed caching?

Without much investigation, initially I thought we need to use the table above?

rbott:
> In `qemu-system-x86_64(1)` there is a nice translation table from `-drive` to `-device` (column 2) and `-blockdev` (columns 3+4) cache options.
> 
> ```
>              │ cache.writeback   cache.direct   cache.no-flush
> ────────────┼─────────────────────────────────────────────────
> writeback    │ on                off            off
> none         │ on                on             off
> writethrough │ off               off            off
> directsync   │ off               on             off
> unsafe       │ on                off            on
> ```
> 
> The first column matches Ganeti's `disk_cache`. Maybe we can add `unsafe` for faster Ganeti-QA runs smile.
> 

Thanks, I actually did not see that table while looking at the documentation. I guess I will add `unsafe` then and update the PR accordingly. I will let you know once I have changed the code. I also need to figure out why the python tests are hanging indefinitely on GH Actions without any output (they happily pass when I run them in the same container locally).

To safe yourself some time, please wait with further reviews until I have dealt with disk_type and disk_cache :-)

rbott:
I have udpated the PR:

- `disk_cache` now accepts its original values again
- `disk_type` now accepts `ide` instead of `ide-hd` again
- actual cache settings are now derived from the table @saschalucas posted above

src/Ganeti/Constants.hs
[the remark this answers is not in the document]
rbott: fixed.

src/Ganeti/Constants.hs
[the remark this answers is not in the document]
rbott: fixed (disk_cache reverted to the old list of accepted values)

rbott:
I have also opened #1669 because I can not test all Ganeti disk_types (but they do not work on the stable-3.0 or master branch as well). But I don't think we should solve/address that as part of this PR.

rbott:
I have added one fix: HotAddDisk() now has the dev_type of the disk-to-add available (plain, drbd8, file etc.) to determine if the underlying storage is file- or blockdevice-based. The previous code checked the wrong variable which broke hot-adding file-based disks.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -459,6 +460,37 @@ def _AnalyzeSerializedRuntime(serialized_runtime):
...
+  """Return cache settings suitable for use with -blockdev
+
+  @param cache_type: string
+  @return: (writeback, direct, no_flush)
+  @rtype: tuple
+  """
+
+  if cache_type == constants.HT_CACHE_NONE:
+    return True, True, False
+  elif cache_type in [constants.HT_CACHE_WBACK, constants.HT_CACHE_DEFAULT]:
```
saschalucas, line 473: Probably `gnt-instance(8)` should be adopted, too. ATM it's documented `default` means "to not pass any cache option". Here it's explicitly `writeback` (which might be the same as not passing any cache option?). Probably we also should think of better defaults. i.e. default to `none`?

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1125,40 +1157,6 @@ def GetAllInstancesInfo(self, hvparams=None):
...
-    @return: parameter string suitable for QEMU drive parameters
-
-    """
-
-    if (dev_type in constants.DTS_EXT_MIRROR
-            and dev_type != constants.DT_RBD):
-      logging.warning("KVM: overriding disk_cache setting '%s' with 'none'"
-                      " to prevent shared storage corruption on migration",
-                      disk_cache)
-      disk_cache = constants.HT_CACHE_NONE
```
saschalucas, line 1148: I wonder if we no longer need to override the cache mode to `none` for shared storage?

rbott: Good catch! I must have overlooked this while refactoring.

lib/hypervisor/hv_kvm/__init__.py
rbott: I have not yet done any documentation work (e.g. `io_uring` is also still missing). I will add that in one of the next iterations and tick the relevant checkbox of the top post.

Regarding `default` I think we should make the documentation more explicit. "not passing any cache option" also lead to explicit cache settings in QEMU before, they just depended on whatever default the QEMU version in question assumed (although I think not much has changed in that regard in the past years).

lib/hypervisor/hv_kvm/__init__.py
rbott: I have re-introduced this as part of the `_GetCacheSettings()` helper in the latest push.

rbott:
**CDROM code transitioned to -blockdev**

The latest push also updates the  `_CdromOption()` method to use `-blockdev` + `-device`  instead of `-drive`. The code has been tested for:

* add one or two CD drives with types `scsi-cd`, `ide` (actually `ide-cd`) or `paravirtual` (actually `virtio-blk-pci`)
* use of local ISOs or HTTP(s) URLs as source (the latter only works on Debian if the package `qemu-block-extra` is installed - but that should be no different from the old code)
* boot from CD with all three disk types

Differences to old code:
* setting `boot_order=cdrom` silently set the cdrom-disk-type to `ide` - that is not happening any more
* disk types other than the three named above will raise a `HypervisorError` (the other types should not work anyways and prevent QEMU from starting)

Limitations of the code (which were already present in the old code):
* using disk_type `paravirtual` actually inserts the CD drives *ahead* of the other virtio disks (e.g. they appear as `vda` and  `vdb`, followed by the first virtio disk as `vdc`)
* no hotplugging (or rather in case of cd drives: ejecting/changing the disk (-iso))
* the code is not covered by unit tests or the QA

rbott:
**CDROM code transitioned to -blockdev**

The KVM hypervisor code has been extended with a `_FloppyOption()` method (analogue to `_CdromOption()`) which extends the KVM commandline with `-blockdev`/`-device` parameters to add a virtual floppy drive using a provided image file.

**Design Document for -blockdev**

I have also added a design document for 3.1, describing the code changes introduced with this PR.


With this, we are now only lacking upgrade/downgrade code + proper testing of cluster upgrades. Could you please take another look, @apoikos @saschalucas?

saschalucas:
> **floppy code transitioned to -blockdev**
...
> **Design Document for -blockdev**

> With this, we are now only lacking upgrade/downgrade code + proper testing of cluster upgrades. Could you please take another look, @apoikos @saschalucas?

Thanks @rbott for your work. I'll have a look. I'm also going to make some tests, especially with `rbd` and `access=userspace`.

doc/design-qemu-blockdev.rst
```diff
@@ -0,0 +1,175 @@
...
+++++++++++++++++++
+
+The code needs to be refactored to make use of the QMP method "blockdev-add",
+using the same parameters as for its commandline counterpart (e.g. blockdev node
+id, caching, aio-mode).
+
+Hot-Removing Disks
+++++++++++++++++++
+
+Hot-adding a disk consists of two steps:
```
saschalucas, line 142: Shouldn't this be Hot-removing, instead of Hot-adding?

doc/design-3.1.rst
```diff
@@ -0,0 +1,7 @@
+=================
+Ganeti 3.0 design
```
saschalucas, line 2: Ganeti 3.1 here?

doc/design-qemu-blockdev.rst
rbott: Yes, absolutely. Thanks for spotting this!

doc/design-3.1.rst
rbott: Uh, copy/paste error. Sorry :-)

rbott:
> Thanks @rbott for your work. I'll have a look. I'm also going to make some tests, especially with `rbd` and `access=userspace`.

Did you already have the time to run some of the checks @saschalucas?

rbott:
I have added migration code to take care of the obsolete `default` value of the `disk_discard` hvparam. During upgrade, all occurances of `default` will now be replaced with `ignore`.

The cluster upgrades well from 3.0.2 to the code in this PR. Of course it does not do anything to running instances with regards to storage devices/disks. _Trying to hot-remove a disk from a running instance after the migration will fail, until the instance gets re-created with -blockdev parameters for drives._ OTOH, re-creating an instance (stop/start or reboot) works exactly as intended.

I yet have to test the following scenario (need multiple nodes for that):
* have a multi-node cluster with 3.0.2 running
* create a DRBD/sharedfile instance
* upgrade cluster to the code in this PR
* live-migrate instance (this should re-create the instance on the target with -blockdev parameters without problems)

If anyone reading this has the time to carry out the test themselves, please proceed :-)

saschalucas:
> Did you already have the time to run some of the checks @saschalucas?

Sorry, very busy ATM. I'll try in the next days.

doc/design-qemu-blockdev.rst
atta: just tested the rbe userspace and got an error
```
cat /var/log/ganeti/kvm/test.vm.log
qemu-system-x86_64: -blockdev driver=raw,node-name=disk-222ec87a-b27d-4756,auto-read-only=off,cache.direct=off,cache.no-flush=off,discard=ignore,file.aio=threads,file.driver=host_device,file.filename=rbd:rbd/4fdf6419-ffd5-4dff-a60f-7a4274a00566.rbd.disk0: Could not open 'rbd:rbd/4fdf6419-ffd5-4dff-a60f-7a4274a00566.rbd.disk0': No such file or directory 
```

rbott:
I have updated this PR and added code to support gluster (both kernel- and userspace). @atta this also adds the groundwork to support other userspace storage backends (e.g. rbd). I still do not have a ceph setup at hand, but if you could provide me with an example rbd URI I could adopt the code to possibly also support rbd userspace access. QEMU seems to [expect the following](https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#qapidoc-807) options: pool, image - and a bunch of optional parameters as well.

For gluster, I simply extracted the required information from the gluster URL (gluster://host:port/volume/path) and filled the required qemu fields with that information.

rbott:
_Finally_ I had some time to revisit this PR. I have now added/verified support for blockdev with RBD storage (both userspace/kernelspace access). There is a lot of room for improvement in the code (especially with the additions for gluster & rbd) and I would like to revise it before we move forward with this.

However, recent QA runs have also shown that with Debian Bookworm hotplugging with the legacy QEMU storage implementation (e.g. in stable-3.0) is broken again. This PR should fix this issue. However I suppose its way to big of a change to get included into Debian Bookworm at his point (we are already knee deep into the freeze/release schedule).

I have improved the QA suite setup to also support RBD setups (only kernelspace access for now). This should make it easier to test.

There is only one main problem left: I have not tested ext-storage yet. According to the documentation, "userspace" access is also supported:

> After successful attachment the script returns to its stdout a string, which is the full path of the block device to which the volume is mapped. e.g:: /dev/dummy1
> [...]
> In case the storage technology supports userspace access to volumes as well, e.g. the QEMU Hypervisor can see an RBD volume using its embedded driver for the RBD protocol, then the provider can return extra lines denoting the available userspace access URIs per hypervisor. The URI should be in the following format: <hypervisor>:<uri>. For example, a RADOS provider should return kvm:rbd:<pool>/<volume name> in the second line of stdout after the local block device path (e.g. /dev/rbd1).
>
> So, if the access disk parameter is userspace for the ext disk template, then the QEMU command will end up having file=<URI> in the -drive option.

This is some sort of an issue now. The current implementation would just pass the userspace URI over to qemu (which will either understand it or not). With the blockdev implementation ganeti actually needs to parse the URI, select the corresponding driver and fill whatever parameters might be required with the information it found in the URI. This somewhat undermines the sense of the ext storage provider. Unless we also change the ext-storage-provider model to require e.g. passing of valid JSON configuration for qemu instead an URI. What would be your take on this @atta @apoikos @saschalucas?

To illustrate this a bit - this is how the RBD userspace access looks like in the JSON syntax for QMP (the commandline syntax is just a flattened-string-representation of the same):

```json
{
   "driver":"raw",
   "node-name":"disk-234d1e63-effa-45b4",
   "discard":"ignore",
   "cache":{
      "direct":true,
      "no-flush":false
   },
   "file":{
      "driver":"rbd",
      "pool":"ganeti",
      "image":"891ef23c-068e-44f2-9636-1bafb5f48c87.rbd.disk4"
   }
}
```

In this case, `pool` and `image` have been extracted from the RBD userspace URL by Ganeti. The same applies to userspace access with GlusterFS. With this in mind, the `file` section would actually have to be provided by the ext-storage-provider.

saschalucas:
thanks @rbott for working on this important topic. I hope to find some time for review soon....

rbott:
This PR has finally left the draft stage :-) 

The latest revision contains the following changes:
- code cleanup
- remove of duplicate code logic in vm-startup and hotplugging disks (both parts shared re-implemented the same logic to find the correct QEMU parameters)
- QEMU commandline parameters now get auto-generated from the corresponding QMP data structure for block devices
- a more generic approach to URI parsing for userspace variants of RBD and Gluster
- unit tests for the above logic

I still have not tested this against an implementation of the ext-storage-interface. However, in theory this _should_ work when the storage provider returns a block device (which is the default) and also _should_ work when it returns an additional userspace URI, as long as it is either RBD or gluster. Other URIs could also work but need to be implemented in the `_ParseStorageUriToBlockdevParam()` function first (which kind of defeats the purpose of the ext storage interface).

I'd say this is really good for testing now. Go for it :-)

doc/design-qemu-blockdev.rst
```diff
@@ -0,0 +1,175 @@
+==========================
+Ganeti daemons refactoring
```
candlerb, line 2: I think this heading needs changing (copy-pasted from design-daemons.rst ?)

doc/design-qemu-blockdev.rst
```diff
@@ -0,0 +1,175 @@
...
+
+  Instances not using direct kernel boot may boot from the floppy drive, when
+  specified through the `boot_order` parameter.
+
+
+Proposed changes
+================
+
+We have to eliminate all uses of the `-drive` parameter from the Ganeti codebase
+to ensure compatibility with future versions fo QEMU. With this, we can also
```
candlerb, line 75: s/ fo / of /

doc/design-qemu-blockdev.rst
rbott: Fixed, thanks :-)

doc/design-qemu-blockdev.rst
rbott: Fixed, thanks :-)

codefritzel (commented):
Hi @rbott,
The code looks good to me.

**Just a general point:**
I think all new helper functions (e.g. `_TranslateBoolToOnOff`, `_ParseStorageUriToBlockdevParam`, ...) should be moved to a new `utils.py`.
The helper functions in `hv_kvm/__init__.py` makes the codefile larger and more complex and at some point the old code should be modernised (snake_case, typehints,...).

rbott:
Thanks @codefritzel for your comments. I have moved the helper functions to a separate utils file.

I will go forward with this and merge this now. However, I will open an issue to keep track of @apoikos' comment above to implement a (basic) way of event handling to replace the `sleep()` call.

PR #1668, opened 2022-04-08
PyYAML 6.0 compatibility

mbakke:
PyYAML 6.0 and later requires an explicit `Loader` argument.  The first commit in this PR adds a `Loader` that preserves previous behavior, while the second commit switches to `SafeLoader` for good measure (mainly to prevent cargo-culting a poor example).

I suppose this should be backported to 3.0 (will there be a 3.0.3?).

### DIFF
--- a/qa/qa_utils.py
+++ b/qa/qa_utils.py
@@ -450,7 +450,7 @@ def GetObjectInfo(infocmd):
   master = qa_config.GetMasterNode()
   cmdline = utils.ShellQuoteArgs(infocmd)
   info_out = GetCommandOutput(master.primary, cmdline)
-  return yaml.load(info_out)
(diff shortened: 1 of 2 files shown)


saschalucas (approved):
Thanks @mbakke for this compatibility fix. `Loader=yaml.SafeLoader` seems to work down to python3-yaml=3.13 from Debian Buster. So a backport to 3.0 should be fine.

rbott:
I am currently doing QA runs on Debian Bullseye and Buster and will merge this PR as soon as both of them pass.

saschalucas:
> I am currently doing QA runs on Debian Bullseye and Buster and will merge this PR as soon as both of them pass.

ACK.

PR #1671, opened 2022-05-03
kvm: Switch to newer -boot order invocation

akosiaris:
Per

https://manpages.debian.org/bullseye/qemu-system-x86/qemu-system-x86_64.1.en.html#boot

the -boot drives format is considered legacy and while still supported
its use is discouraged as it may be removed from future versions.

The newer form is more expressive and flexible, by e.g. allowing to show
splash screens, menus, specify boot orders and then override once etc

Switch the -boot <drive> invocations to -boot order=<drive> as a first
step

Signed-off-by: Alexandros Kosiaris <akosiaris@gmail.com>


saschalucas (approved):
Thanks @akosiaris. This change looks fairly simple and reasonable.

Sorry for the delay. LGTM

PR #1672, opened 2022-05-28
QA: fix SSH multiplexing setup

rbott:
[SSH multiplexing](https://blog.scottlowe.org/2015/12/11/using-ssh-multiplexing/) was broken in the QA suite, because the code only used to set up a multiplexing connection to the first node, not for all nodes.

Addtionally, there was a bug that let to the multiplexer not being used in certain scenarios. Multplexed connections are stored in the `qua_utils._MULTIPLEXERS` dictionary. However, since the `qa_utils` module gets imported in two different flavors at the same time (as the entire module but also selected functions directly like `AssertCommand()`). Only one instance had the `_MULTIPLEXERS`  dict populated - SSH calls with e.g. `AssertCommand()` in the callstack would not be able to use the multiplexed connections.
This is fixed bei moving the `_MULTIPLEXERS` dictionary over to the `qa_config` module which only has one instance throughout the entire QA.

This speeds up the QA on a multi-node cluster by a few minutes (depending on the actual QA setup).

### DIFF
--- a/qa/ganeti-qa.py
+++ b/qa/ganeti-qa.py
@@ -1178,12 +1178,11 @@ def main():
 
   qa_config.Load(config_file)
 
-  primary = qa_config.GetMasterNode().primary
-  qa_utils.StartMultiplexer(primary)
-  print("SSH command for primary node: %s" %
-        utils.ShellQuoteArgs(qa_utils.GetSSHCommand(primary, "")))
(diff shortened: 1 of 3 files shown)


saschalucas (approved):

rbott:
Thank you! :-)

PR #1676, opened 2022-06-05
Replace epydoc with pydoctor

newellz2:
I modified Makefile.am to use pydoctor==19.11, which is the version available via apt in Debian 11 Bullseye, to generate the py-apidocs instead of Epydoc. Epydoc's last release was in 2008 and does not have support for Python3. The pydoctor.ini file contains some basic settings for the py-apidoc
generation.

I corrected some epytext in the instance_migration.py file so the `make py-apidocs` command exits successfully.

You can view the new apidocs associated with my commit here:
https://ncr-ng.cse.unr.edu/py/


Makefile.am
```diff
@@ -2892,11 +2892,14 @@ apidoc: py-apidoc
...
 	env - PATH="$$PATH" PYTHONPATH="$$PYTHONPATH" \
-	$(RUN_IN_TEMPDIR) epydoc -v \
-	  --conf $(CURDIR)/epydoc.conf \
-	  --output $(CURDIR)/$(APIDOC_PY_DIR)
+	$(RUN_IN_TEMPDIR) pydoctor -v \
+	  --config $(CURDIR)/pydoctor.ini \
+	  --html-output $(CURDIR)/$(APIDOC_PY_DIR) \
+          $(lint_python_code)
+
+
```
saschalucas, line 2903: Thanks a lot @newellz2.

Do the above two new lines look to much?

However, please also cleanup the `Makefile.am` with occurrence of epydoc: `epydoc.conf` target itself and from `CLEANFILES`, `epydoc.conf.in` from `EXTRA_DIST` (add `pydoctor.ini` therefore).

newellz2: @saschalucas It is too much. 😀 

I cleaned up the references to epydoc in Makefile.am and the py-apidoc target.

saschalucas (changes requested):
Thanks so far. I just discovered one additional cleanup (Sorry for not looking so far in the first review):

Since the `epydoc.conf` target and `epydoc.conf.in` are now removed, the pattern `@MODULES@` is nowhere else in use. Please remove it from `$(REPLACE_VARS_SED)` target, too.

Other than that, LGTM.

newellz2:
@saschalucas, I removed the line responsible for adding the `@MODULES@` variable to the `replace_vars.sed` file. Let me know if anything else needs to be done.

saschalucas (approved):
Thanks @newellz2! LGTM.

PR #1679, opened 2022-06-22
add test data to Makefile

saschalucas:
During current development, new test data were added. The `devel/release` script requires all files included in `Makefile.am`. ATM `py-tests` breaks, because of missing test data inside the temporary build tree.


rbott (approved):

rbott:
Thanks for spotting this! I wonder why I did not stumble upon this earlier :(

LGTM!

PR #1686, opened 2022-11-11
Fix variable names in ganeti-os-interface(7).

mbakke:
The `COMMON ENVIRONMENT` section of [ganeti-os-interface(7)](https://docs.ganeti.org/docs/ganeti/3.0/man/ganeti-os-interface.html) has some stray hyperlinks and lack an underscore.  The reason is that the RsT processor treats a "trailing" underscore (one followed by special symbols such as `\n` and apparently `%`) as hyperlinks: this PR fixes that.


rbott (approved):
LGTM, thank you :-)

PR #1688, opened 2022-12-04
Replace google.com/test.com by rfc 2606 domain example.com

mainboarder:
as users do not have control over google.com or test.com zone it is better to use one of the dedicated example domains.


rbott (approved):
LGTM and thank you for your first contribution to the Ganeti project :-)

I'll go forward and merge this.

PR #1690, opened 2023-02-23
Add a  document for Terrafor

atta:
A Document on how a potential Terraform
Provider for ganeti could be done

### DIFF
--- a/doc/design-terrafor-provider.rst
+++ b/doc/design-terrafor-provider.rst
@@ -1,0 +1,175 @@
+==================================
+Ganeti Terraform Provider Proposal
+==================================
+
+:Created: 2023-Feb-23
+:Status: Draft
+
+.. toctree::
+   :hidden:
+
+1. Introduction
+---------------
+
+This document proposes the creation of a Terraform provider for Ganeti, 
+a cluster virtualization management software. The provider will 
+allow users to manage their Ganeti infrastructure as code, 
+using the Terraform configuration language and tool. 
+The provider will be implemented using the Terraform provider SDK 
+and the Ganeti Remote API.
+
+2. Background
+-------------
+
+Terraform is an infrastructure-as-code tool that enables users to manage
+and provision cloud infrastructure using a declarative configuration 
+language. Terraform is platform-agnostic and supports a wide range of 
+infrastructure providers, including major cloud providers such as AWS, 
+Azure, and Google Cloud. 
+Terraform providers are plugins that extend Terraform's capabilities 
+to support new infrastructure platforms or services.
+
+Ganeti is a cluster virtualization management software that allows users
+to create and manage virtual machines (VMs) on a cluster of 
+physical hosts. Ganeti provides a number of features, such as automatic 
+failover, live migration, and DRBD-based disk replication. 
+Ganeti also has a Remote API that enables users to manage their Ganeti 
+cluster programmatically.
+
+3. Scope
+--------
+
+The proposed Ganeti Terraform provider will support the following Ganeti
+resources:
+
+- `ganeti_instance`: A Ganeti virtual machine instance, running on a 
+  physical host (node).
+
+The provider will support basic create, read, update, and delete (CRUD)
+operations for these resources, as well as importing existing resources 
+into Terraform state. The provider will also support dependencies 
+between resources, such as instances depending on a other instance.
+
+1. Requirements
+---------------
+
+The following requirements must be met in order to create the Ganeti 
+Terraform provider:
+
+- Access to a Ganeti cluster, with the Ganeti Remote API enabled and 
+  configured.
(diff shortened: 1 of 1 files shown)


doc/design-terrafor-provider.rst
```diff
@@ -0,0 +1,175 @@
...
+to support new infrastructure platforms or services.
+
+Ganeti is a cluster virtualization management software that allows users
+to create and manage virtual machines (VMs) on a cluster of 
+physical hosts. Ganeti provides a number of features, such as automatic 
+failover, live migration, and DRBD-based disk replication. 
+Ganeti also has a Remote API that enables users to manage their Ganeti 
+cluster programmatically.
+
+3. Scope
```
rbott, line 39: This section should go a bit more into detail about what should be implemented or can be expected from the provider, so that we end up with something people can use to create code and/or also use to compare against PRs etc.

- supported hypervisors
- supported disk templates
- supported disk schemes (number of disks etc.)
- supported network schemes (numbers of nics, network types etc.)
- other supported features (e.g. automatic/manual node selection, custom hypervisor parameters)

I would suggest to start simple:
- we only support KVM for the first iteration
- all possible disk templates
- only automatic instance allocation
- any number of disks, any number of NICs
- network mode `bridged` with optional parameters `link`, `network` and `vlan` (this should cover most usecases)
- all backend parameters (`maxmem`, `minmem`, `vcpus`)
- all current KVM hvparams (given that we are using a strongly typed language, supporting "any" parameter will be harder to implement. OTOH most people will likely use only few hvparams on instance level as most will usually be set on cluster level)
- set OS type (and possibly also OS parameters?)

We also need to agree on what the provider should do with instance changes:
- will new disks/NICs be hot- or cold-added? Will this be configurable?
- will disks/NICs be deleted? Do we support that at all? Or maybe only in later releases? How should the terraform data be modeled to "detect" that a disk/NIC has been removed from the instance definition?
- it should support disk growing
- what happens when backend/hv parameters are changed? will this trigger an automatic reboot of the instance? should this be configurable?

rbott (commented):
Thanks @atta for your PR :-) I agree that we should provide an upstream terraform provider so that people can easier integrate Ganeti into their workflows/tools.

This should probably live in its own repository under the ganeti organisation.

Since we already have lots of Ganeti RAPI Go code in [gnt-cc](https://github.com/sipgate/gnt-cc) I would strongly suggest to create an official RAPI client in Go as a separate package which both projects can use. This way we avoid duplicating code. 

What do you think?

I have also included some more comments in this review. In general I think we should specify what to expect from the terraform provider (and what not to expect from it) and how it should be used. This will allow others to join the development efforts as it will be much clearer what the scope of this project.

doc/design-terrafor-provider.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Here is an example Terraform configuration that uses the Ganeti provider
+to manage a Ganeti instance:
+
+::
+   provider "ganeti" {
+    endpoint = "https://ganeti-cluster.example.com:5080"
+    username = "admin"
+    password = "s3cr3t"
+  }
```
rbott, line 164: This line seems to be a compile error - building the documentation fails with the following message:
```

Warning, treated as error:
/tmp/gntbuild.daxPPrTS/doc/design-terrafor-provider.rst:164:Block quote ends without a blank line; unexpected unindent.
```

doc/design-terrafor-provider.rst
```diff
@@ -0,0 +1,175 @@
...
+--------------
+
+- Terraform: https://www.terraform.io/
+- Ganeti: https://ganeti.github.io/
+- Ganeti Remote API: https://docs.ganeti.org/docs/ganeti/2.17/html/rapi.html
+- Terraform Provider SDK: https://www.terraform.io/docs/extend/plugin-sdk.html
+- Go programming language: https://golang.org/
+- Terraform schema language: https://www.terraform.io/docs/language/types/schema.html
+
+11. Appendix: Example Configuration
```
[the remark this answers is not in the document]
atta: @rbott my intention was to not overload this document: a lot of features are indeed possible with terraform but not jet a goal as a start.

doc/design-terrafor-provider.rst
atta: @rbott i will incorporate your suggestion, however i still think i should be less as a start to not over-complicate it.

rbott: This document needs to be fairly exhaustive to not over-complicate the development of the terraform provider itself :-) If we expect people to contribute, everyone needs to be able to understand planned features/goals/ideas. Otherwise we scare contributors away be rejecting their contributions because they do not meet whatever we might have in mind.

This document can/should also be used to decide whether the provider has reached a stable version 1.0 or if it is still missing features considered a bare minimum requirement.

PR #1692, opened 2023-03-06
uidpool_unittest: avoid using negative UIDs

### LINKED ISSUE #1691: ganeti-testsuite: autopkgtest failure with procps 2:4.0.3-1
As report in Debian bug [#1031765](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1031765) the handling of user id -1 has been changed current pgrep.

All user ids in Linux are positive number. pgrep does not longer report `pgrep -u -1` as an unused user ID with return code `$?=1` but as an invalid argument with return code `$?=2`.

The test testRequestUnusedUid() in test/py/ganeti.uidpool_unittest.py needs to be adjusted.

apoikos:
The uidpool unittests long relied on the fact that pgrep would accept "-1" as a valid uid and in turn would match nothing. As of propcs 4.0.3 however, pgrep will error out when passed a negative uid, thus breaking the tests.

Work around this by using a valid UID, but one that is unlikely to be used in real systems: on most systems, SUB_UID_MAX which is the highest uid expected to be in regular use, is set to 600100000. Picking 2^30 + 42 looks like a reasonably safe choice.

This fixes #1691.

### DIFF
--- a/test/py/ganeti.uidpool_unittest.py
+++ b/test/py/ganeti.uidpool_unittest.py
@@ -106,23 +106,24 @@ class TestUidPool(testutils.GanetiTestCase):
 
     # Check with a single, known unused user-id
     #
-    # We use "-1" here, which is not a valid user-id, so it's
-    # guaranteed that it's unused.
-    uid = uidpool.RequestUnusedUid(set([-1]))
-    self.assertEqualValues(uid.GetUid(), -1)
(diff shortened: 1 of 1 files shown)


saschalucas (approved):
Thanks @apoikos, please merge your changes.

PR #1694, opened 2023-03-06
Fix building against GHC 9 / Lens 5

apoikos:
It looks like something in the GHC 9 / Lens 5 combination yields unexpected results when using view on Lens', with GHC complaining that it's passed a Lens' instead of a Getting (which should still be compatible).  Commenting out the signatures of Ganeti.Network.poolLens, Ganeti.Network.poolArrayLens and Ganeti.Utils.MultiMap.multiMapL makes Ganeti build again, so I'd keep this ugly hack until we figure out what's actually wrong.

### DIFF
--- a/src/Ganeti/Network.hs
+++ b/src/Ganeti/Network.hs
@@ -87,11 +87,11 @@ data PoolPart = PoolInstances | PoolExt
 addressPoolIso :: Iso' AddressPool BA.BitArray
 addressPoolIso = iso apReservations AddressPool
 
-poolLens :: PoolPart -> Lens' Network (Maybe AddressPool)
+--poolLens :: PoolPart -> Lens' Network (Maybe AddressPool)
 poolLens PoolInstances = networkReservationsL
 poolLens PoolExt = networkExtReservationsL
 
-poolArrayLens :: PoolPart -> Lens' Network (Maybe BA.BitArray)
+--poolArrayLens :: PoolPart -> Lens' Network (Maybe BA.BitArray)
 poolArrayLens part = poolLens part . mapping addressPoolIso
 
 netIpv4NumHosts :: Network -> Integer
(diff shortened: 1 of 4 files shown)


saschalucas (commented):
Dear @apoikos,

does this PR address compile errors like this one?
```
src/Ganeti/Utils/MultiMap.hs:103:17: error:
    • Couldn't match type: forall (f :: * -> *).
                           Functor f =>
                           (S.Set v0 -> f (S.Set v0)) -> MultiMap k v0 -> f (MultiMap k v0)
                     with: (S.Set v -> Const (S.Set v) (S.Set v))
                           -> MultiMap k v -> Const (S.Set v) (MultiMap k v)
      Expected: k -> Getting (S.Set v) (MultiMap k v) (S.Set v)
        Actual: k -> Lens' (MultiMap k v0) (S.Set v0)

```
and what will happen if the type signature is removed? Will Haskell compute the "right one" implicitly? If so, what are the implications of automatic type signatures?

TIA, Sascha.

apoikos:
@saschalucas yes, it fixes these errors. In general, type signatures are optional as long as the compiler can figure them out. For most simple functions, the compiler can rely on type inference to guess the function's signature. If the compiler can infer a function's type *and* a signature has been explicitly defined, then these two *must* match, otherwise - as in this case - the compilation errors out.

Now, it looks like with have an additional error with the tests, one that I didn't catch in Debian because I'm not running the Haskell tests there during build. I will address this shortly.

saschalucas (approved):
Thanks @apoikos for your explanation. Let's go ahead an use automatic type signatures if that helps to run on newer Haskell stacks.

As you mentioned, github CI `hs-tests` fails ATM. Fixing this would be nice, too.

apoikos:
@saschalucas, I fixed the other build issues as well.

anarcat:
is this commit planned to be backported into Debian bookworm somehow? worried about upgrading our clusters to bookworm right now...

Issue #1687, opened 2022-11-29
Support for Python 3.10/3.11 needed


robbat2:
There was a previous PR for Python3.8 support, but it's not clear further Python work has taken place to support Python 3.9/3.10/3.11.

Upstream Python has both 3.8 & 3.9 series in security-only fixes.

Is there an effort to support newer Python yet, before distributions start to remove Python 3.8? (Gentoo is discussing removal of 3.8 already).

jfut:
For your information, at least on Python 3.9.14 on RHEL/AlmaLinux/Rocky Linux 9, all `make py-tests` succeed and works fine.

`make py-tests` results:
https://github.com/jfut/ganeti-rpm/issues/50#issuecomment-1255719822

My test environment:

```bash
[jfut@node01.example.org ~]$ cat /etc/redhat-release
AlmaLinux release 9.1 (Lime Lynx)

[jfut@node01.example.org ~]$ ps aux | grep ganeti
gnt-con+     878  0.0  0.0 1073993796 19756 ?    Ss   Nov21   1:23 /usr/sbin/ganeti-confd -f
root         883  0.0  0.0 1073983660 12708 ?    Ss   Nov21   0:01 /usr/sbin/ganeti-kvmd -f
root         887  0.0  0.8 264416 264304 ?       SLs  Nov21   0:14 /usr/bin/python3 /usr/sbin/ganeti-noded -f
gnt-mas+     892  0.0  0.0 1073998356 27676 ?    Ss   Nov21   0:30 /usr/sbin/ganeti-wconfd -f
gnt-mas+    1100  0.0  0.0 1073998288 30264 ?    Ss   Nov21   3:00 /usr/sbin/ganeti-luxid -f
gnt-rapi    1101  0.0  0.1 256924 34516 ?        Ss   Nov21   0:03 /usr/bin/python3 /usr/sbin/ganeti-rapi -f -b 127.0.0.1 --require-authentication
gnt-met+    5568  0.0  0.0 1073988936 13376 ?    Ss   Nov21   0:00 /usr/sbin/ganeti-metad -f
jfut      227665  0.0  0.0 221796  2368 pts/0    S+   11:58   0:00 grep --color=auto ganeti

[jfut@node01.example.org ~]$ /usr/bin/python3 -V
Python 3.9.14

[jfut@node01.example.org ~]$ rpm -qa | grep python3-3
python3-3.9.14-1.el9_1.1.x86_64
```

[...]

PR #1695, opened 2023-03-07
Replace calls to deprecated threading methods

apoikos:
Python 3.10 has deprecated a bunch of camelCase threading methods in favor of snake_case methods and getter/setter attributes. The new methods/attributes have existed since Python 2.6, so switching does not break backward compatibility.

This is part of #1687.


saschalucas (approved):
Thanks @apoikos. Please merge your changes.

Issue #1696, opened 2023-03-10
move-instance difficult to use and ultimately fails


anarcat:
I'm trying to migrate between two Ganeti clusters. I have found with great anticipation the [move-instance](https://docs.ganeti.org/docs/ganeti/3.0/html/move-instance.html) command, but I'm having a hard time making it work.

At first, it would just crash with a backtrace in Debian bullseye:

```
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

That's due to this code:

https://github.com/ganeti/ganeti/blob/114e59fcc9d4a7c82618569f5d6b7389a0f80123/tools/move-instance#L941

If I pass `--opportunistic-tries=1` it tells me:

```
move-instance: error: Opportunistic instance creation can only be used with an iallocator
```

So, basically, right now, you *must* use:

```
move-instance --opportunistic-tries=1 --iallocator=hail
```

According to @apoikos (on IRC), the `TypeError` is a python2-to-3 leftover...

The next problem I had with `move-instance` was a `ganeti.rapi.client.Error: Password not specified`, but that was me failing at setting up the RAPI users. I also got `ganeti.rapi.client.GanetiApiError: 401 Unauthorized: No permission -- see authorization schemes` on the destination cluster. Maybe the docs could be improved to lead the operator the right way ("check your RAPI users again") in the documentation. Having a way to test the users out of band (say with `curl`) would also be useful here.

Then I had another error which was pretty opaque:

```
ganeti.errors.OpPrereqError: ("Invalid handshake: Hash didn't match, clusters don't share the same domain secret", 'wrong_input')
```

So that might seem obvious but I *did* copy the secret over and ran:

```
gnt-cluster renew-crypto --cluster-domain-secret=cluster-domain-secret
```

So it seems the bug *there* is that the `--cluster-domain-secret=` argument actually fails to replace the secret on the cluster. I had to manually copy the `cluster-domain-secret` file in `/var/lib/ganeti` and restart the server for that to work.

But what completely blocked me is this:

```
ganeti.errors.OpPrereqError: ('If network is given, no mode or link is allowed to be passed', 'wrong_input')
```

It looks like the source node is encoding NIC information in the backup and the target node is somewhat unhappy with it. I'm not sure how to debug this: I'm lost in the stack between the client and server method definitions and I don't actually understand what's going on so much.

Did *anyone* get that thing to work at all? What am I doing wrong?

Should I open separate issues for those things?

anarcat:
[...]

PR #1698, opened 2023-03-13
move-instance: allow users to empty out NIC parameters

anarcat:
In #1696, I ended up in a situation where the instance I'm moving had both "network" and "link" parameters. That seems like a natural configuration in our cluster, but something move-instance really doesn't like, as it's passing this verbatim to create-instance which naturally freaks out on the contradictory arguments.

Instead of just crashing, detect empty parameters (`not v` is a shortcut here) and null them out. This will result in those parameters being passed as empty to create-instance, or to be more accurate, not be passed at all.

I've been able to fix the crash in #1696 by passing `--net 0:ip=pool,network=gnt-dal-01,mode=,link=` to move-instance with this patch.

I'm not sure it's the right approach though. I'd much rather *not* have to pass `--net` at all, as I actually want to move multiple instances across clusters, and that seems incompatible with `--net` for some reason I cannot currently fathom (but which is possibly related to this problem).

### DIFF
--- a/tools/move-instance
+++ b/tools/move-instance
@@ -972,6 +972,12 @@ def _CheckInstanceOptions(parser, options, instance_names):
 
     if options.nics:
       options.nics = cli.ParseNicOption(options.nics)
+      # allow user to override instance parameters by hardcoding the
+      # empty string as "ignore this setting" shortcut
+      for nic in options.nics:
+        for k, v in nic.items():
+          if v == "":
+            nic[k] = None
   else:
     # Moving more than one instance
     if compat.any(options.dest_instance_name, options.dest_primary_node,


tools/move-instance
```diff
@@ -972,6 +972,12 @@ def _CheckInstanceOptions(parser, options, instance_names):
 
     if options.nics:
       options.nics = cli.ParseNicOption(options.nics)
+      # allow user to override instance parameters by hardcoding the
+      # empty string as "ignore this setting" shortcut
+      for nic in options.nics:
+        for k, v in nic.items():
+          if not v:
```
apoikos, line 979: I would prefer to make it explicit that we are looking for an empty string here (`if not v == ""` or `if not len(v)`), because `not v` also matches things like `0` which can be legitimate values.

anarcat: oh yeah of course, fixed.

also, i think you meant `if v == ""` here, right?

anarcat:
@rbott in case you missed it, this is the third PR i need to get move instance flying on my end right now. :)

rbott:
LGTM, thanks @anarcat  :-)

rbott (approved):

Issue #1702, opened 2023-03-21
move-instance optimisation: don't wait for DRBD sync


anarcat:
when we create VMs in the DRBD backend, we pass `--no-wait-for-sync` to speed up the instance creation. the downside is the VM is sitting there syncing the DRBD disks for a while, but at least it's responsive quickly and anyway we typically have post-creation bootstrapping to do before the machine is actually in production.

when i *move* an instance with the `move-instance` command, however, it doesn't specify that option and spends a *long* time doing that silly transfer.

let's optimize this out or at least make it configurable.

jcharaoui:
There is also a way to simply bypass the initial sync, which is (often) useless for new volumes:

>     # drbdadm create-md <resource>  -  (do this on both nodes)
>     # drbdadm up <resource>  -  (do this on both nodes)
>     # drbdadm -- --clear-bitmap new-current-uuid <resource>  -  (do this on secondary node)
>     # drbdadm primary <resource>  -  (do this on primary node)

This is what one of the core developpers of DRBD recommend.
See https://www.mail-archive.com/drbd-user@lists.linbit.com/msg12212.html

anarcat:
On 2023-03-21 11:23:45, Jerome Charaoui wrote:
> There is also a way to simply bypass the initial sync, which is (often) useless for new volumes:
>
>>     # drbdadm create-md <resource>  -  (do this on both nodes)
>>     # drbdadm up <resource>  -  (do this on both nodes)
>>     # drbdadm -- --clear-bitmap new-current-uuid <resource>  -  (do this on secondary node)
>>     # drbdadm primary <resource>  -  (do this on primary node)
>
> This is what one of the core developpers of DRBD recommend.
> See ***@***.***/msg12212.html

While this is a great idea, this is something that could be done for
instance creation, in general. We already have a tweak that allows us to
create the DRBD pair without waiting for it, I'm suggesting we just
enable that bypass for this operation here.

But yes, that would be a nice optimization to make in create-instance
itself, maybe we even already do this! I haven't checked.

zen2:
> But yes, that would be a nice optimization to make in create-instance itself, maybe we even already do this! I haven't checked.

That could be used too when adding a new disk to an existing instance too.

PR #1703, opened 2023-03-21
optimize move-instance: do not wait for DRBD sync

anarcat:
This is debatable and a change from the default in create-instance, but I think it makes sense. We're already creating a copy of the VM in move-instance, if `keep_source_instance` is set, of course, but also *while* the migration is happening, in any case. So we shouldn't worry too much about the lack of DRBD consistency during the move.

Otherwise this takes twice the time it could normally take, as the move needs to wait for the entire disk to sync. In my tests, it turns a 17 minutes migration into a ~7 minutes migration, improving performance by a twofold.

Naturally, this could also be made configurable, but I am not sure we want to exactly copy all of the instance creation parameters here, and this one particularly feels like a default that could be improved.

Just like `keep_source_instance` used to be hardcoded, let's start with a hardcoded value and switch it to configurable if people actually need to.

Closes: #1702

### DIFF
--- a/tools/move-instance
+++ b/tools/move-instance
@@ -690,7 +690,8 @@ class MoveDestExecutor(object):
                              osparams=objects.FillDict(inst_osparams,
                                                        override_osparams),
                              opportunistic_locking=is_attempt_opportunistic,
-                             tags=instance["tags"]
+                             tags=instance["tags"],
+                             wait_for_sync=False,
                              )
 
 


rbott:
Hi @anarcat,

I think this is worth a try. Usually ganeti focuses on maximum availability/safety but as you stated; in this case there already is a copy available on the source side.

We will see how this evolves in the future :-)

rbott (approved):

PR #1704, opened 2023-04-14
cherry-pick watcher enhancements from attic branch

saschalucas:
I'm abusing `ganeti-watcher` for upgrades of HV parameters (`cpu_type`, `machine_version`, ...) or backend parameters (`vcpus`, `memory`):
* just shut down instances from within the guest
* wait until next 5 minute slot (watcher cronjob)
* and let the watcher startup the instance again (with new parameters set)

This works good, as long as the watcher can run. However long running jobs like replace-disks build the following chain:
* the fist watcher run, after long-running-job has started, will submit a `verify-disks` job
* `verify-disks` waits in the queue for long-running-job to finish
* further watcher runs immediately return doing nothing, because a file lock for the watcher process is held open until `verify-disks` job finishes.

Commit beff5977e will close the file lock before submitting the `verify-disks` job. And commit 765063882 will prevent now to fill the queue with further `verify-disks` jobs.

Commit 3a366d647 further enhances `verify-disks` not to always wait for the long-running-job to complete.

### DIFF
--- a/doc/examples/ganeti.cron.in
+++ b/doc/examples/ganeti.cron.in
@@ -3,8 +3,11 @@ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin
 # On reboot, continue a Ganeti upgrade, if one was in progress
 @reboot root [ -x @SBINDIR@/gnt-cluster ] && @SBINDIR@/gnt-cluster upgrade --resume
 
-# Restart failed instances (every 5 minutes)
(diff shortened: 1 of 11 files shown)


rbott (approved):
LGTM, thanks for finding this @saschalucas :-)

We should make sure this gets properly mentioned in the NEWS section with the 3.1 release as it will also touch the ganeti cronjob example file and probably needs to be picked up separately by package maintainers.

Debian Bullseye KVM QA run went fine, Bookworm currently does not work due to problems with qemu + hotplugging (but not related to this PR). A xen run is currently still ongoing. Results can be checked here: https://stargazer.ganeti.org

PR #1706, opened 2023-05-25
Add warning about setting the PVID as instance vlan

### LINKED ISSUE #1670: Adding nic to a vm on vLan 1 causes network issues on host node
### Scenario
- Management interface is in xen-br0, with untagged traffic
- The default vLan on the network is vLan 1

### Issue
If you add a nic interface to an instance (type=bridged, vlan=1) the host network becomes inaccessible. 
`
gnt-instance modify --net add:mode=bridged,vlan=1 server.name
`

I feel this should be raised as a critical issue as this will cause network issues with the node.

rbott:
As discussed in #1670 this commit adds a warning to not set the PVID (port vlan ID) as an instance VLAN ID. Network equipment might drop frames which explicitly have the VLAN ID set which is configured as the default VLAN ID/port VLAN ID on the network side.

This fixes #1670.

### DIFF
--- a/man/gnt-instance.rst
+++ b/man/gnt-instance.rst
@@ -181,11 +181,13 @@ name
 vlan
    in bridged and openvswitch mode specifies the VLANs that the NIC
    will be connected to. To connect as an access port use ``n`` or 
-   ``.n`` with **n** being the VLAN ID. To connect as an trunk port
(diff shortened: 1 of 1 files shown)


atta (approved):

atta:
LGTM; Thanks for adding the Clarification

PR #1711, opened 2023-07-07
KVM: handle migrate-set-capabilities with qemu upgrades

robbat2:
Newer version of QEMU only support the string `rdma-pin-all`, not the older `x-rdma-pin-all`. The migration works fine if the string is just updated during a migration.

The error preventing migration is:
```
Failed to accept instance: kvm: error executing the
migrate-set-capabilities command: Parameter 'capability' does not accept
value 'x-rdma-pin-all' (GenericError):
````

### DIFF
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -774,6 +774,13 @@ class QmpConnection(MonitorSocket):
       "capabilities": []
     }
     for capability in capabilities:
+      # Fix the old capability names
+      # TODO: how do we verify that the new name is supported?
+      # Fixes "error executing the migrate-set-capabilities command:
+      # Parameter 'capability' does not accept value 'x-rdma-pin-all'"
+      if capability == 'x-rdma-pin-all':
+        capability = 'rdma-pin-all'
+      # Setup the caps.
       arguments["capabilities"].append({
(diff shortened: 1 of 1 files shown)


robbat2:
@rbott  @atta  could you approve/merge please?

saschalucas (changes requested):
The KVM parameter `migration_caps` is just a string split at `:`. You can set it to what ever your qemu version supports.

The solution to your problem is just to change your cluster config to the right `migration_caps` value of `rdma-pin-all`.

The only needed code change inside Ganeti is the man page of `gnt-instance`, no longer mentioning the x-variant of `rdma-pin-all`.

https://github.com/ganeti/ganeti/blob/68dc7eddc277f3e84ac3d63b80a151169ff03a2d/man/gnt-instance.rst?plain=1#L896-L898

A quick search shows that `rdma-pin-all` [seems no longer experimental since qemu-2.0](https://wiki.qemu.org/Features/RDMALiveMigration). The question that remains open is, when was the experimental name removed? However it looks save to just switch to `rdma-pin-all` because <qemu-2.0 is out of scope for current Ganeti.

rbott:
I'll close this PR as it is not a Ganeti problem per se. Although we need to update the documentation/man page to not list the experimental param name any more :-)

PR #1712, opened 2023-07-08
lib/runtime: improve not-master error

robbat2:
For unclear reasons, there was an environment where the master was running, but the clients did not believe it was the master.

kernel.hostname returned the unqualified hostname, as expected, but the master check was looking for the FQDN.

Improve the error message as a start on making it easier to debug.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

### DIFF
--- a/lib/runtime.py
+++ b/lib/runtime.py
@@ -286,8 +286,11 @@ def GetClient():
 
     master, myself = ssconf.GetMasterAndMyself(ss=ss)
     if master != myself:
-      raise errors.OpPrereqError("This is not the master node, please connect"
-                                 " to node '%s' and rerun the command" %
-                                 master, errors.ECODE_INVAL)
+      # print our own name to aid debugging, esp. re FQDN
+      # common error is: hostname != hostname.domain.tld
+      raise errors.OpPrereqError("This node, '%s', is not the master node,
+                                 " please connect to node '%s' and rerun "
+                                 " the command" %
+                                 % (myself, master), errors.ECODE_INVAL)
     raise
   return client


robbat2:
@rbott  @atta review please?

lib/runtime.py
```diff
@@ -286,8 +286,11 @@ def GetClient():
...
     if master != myself:
-      raise errors.OpPrereqError("This is not the master node, please connect"
-                                 " to node '%s' and rerun the command" %
-                                 master, errors.ECODE_INVAL)
+      # print our own name to aid debugging, esp. re FQDN
+      # common error is: hostname != hostname.domain.tld
+      raise errors.OpPrereqError("This node, '%s', is not the master node,
+                                 " please connect to node '%s' and rerun "
+                                 " the command" %
+                                 % (myself, master), errors.ECODE_INVAL)
```
mbakke, line 294: Looks like there is an extra `%` at the start of the line.  Otherwise LGTM.

lib/runtime.py
```diff
@@ -286,8 +286,11 @@ def GetClient():
 
     master, myself = ssconf.GetMasterAndMyself(ss=ss)
     if master != myself:
-      raise errors.OpPrereqError("This is not the master node, please connect"
-                                 " to node '%s' and rerun the command" %
-                                 master, errors.ECODE_INVAL)
+      # print our own name to aid debugging, esp. re FQDN
+      # common error is: hostname != hostname.domain.tld
+      raise errors.OpPrereqError("This node, '%s', is not the master node,
```
saschalucas, line 291: Also there seems a missing double quote at the end:

```
Exception occurred:
  File "/tmp/gntbuild.kZnFQSGG/ganeti/cli.py", line 62, in <module>
    from ganeti.runtime import (GetClient)
  File "/tmp/gntbuild.kZnFQSGG/ganeti/runtime.py", line 291
    raise errors.OpPrereqError("This node, '%s', is not the master node,
                                                                       ^
SyntaxError: EOL while scanning string literal
```

lib/runtime.py
codefritzel, line 294: In such cases I personally prefer f-strings
```suggestion
      raise errors.OpPrereqError(f"This node, {myself}, is not the master node, "
                                 f"please connect to node {master} and rerun "
                                 "the command",
                                 errors.ECODE_INVAL)
```
Perhaps the old %-style should be replaced by f-strings, which are supported in version 3.6?

saschalucas: Thanks for your suggestion. f-strings look pretty readable. However for the moment we stay with the old `%` style to align with existing code. Maybe in future we can adopt to follow the [language authorities standards](https://peps.python.org/pep-0498/).

PR #1716, opened 2023-10-05
Fix building with Template Haskell 2.18 / GHC 9.2.

mbakke:
Hello,

Here is another hack similar to #1694 that fixes building with GHC 9.2 simply by commenting type signatures.  Also introduce a `myConP` for backwards compatibility with Template Haskell < 2.18.

CC @apoikos .

### DIFF
--- a/src/Ganeti/BasicTypes.hs
+++ b/src/Ganeti/BasicTypes.hs
@@ -206,12 +206,12 @@ instance MonadTrans (ResultT a) where
 instance (MonadIO m, Error a) => MonadIO (ResultT a m) where
   liftIO = ResultT . liftIO
                    . liftM (either (failError . show) return)
-                   . (try :: IO a -> IO (Either IOError a))
+                   . (try :: IO α -> IO (Either IOError α))
 
 instance (MonadBase IO m, Error a) => MonadBase IO (ResultT a m) where
   liftBase = ResultT . liftBase
                    . liftM (either (failError . show) return)
-                   . (try :: IO a -> IO (Either IOError a))
+                   . (try :: IO α -> IO (Either IOError α))
 
 instance (Error a) => MonadTransControl (ResultT a) where
 #if MIN_VERSION_monad_control(1,0,0)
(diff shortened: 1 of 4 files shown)


src/Ganeti/BasicTypes.hs
```diff
@@ -206,12 +206,12 @@ instance MonadTrans (ResultT a) where
 instance (MonadIO m, Error a) => MonadIO (ResultT a m) where
   liftIO = ResultT . liftIO
                    . liftM (either (failError . show) return)
-                   . (try :: IO a -> IO (Either IOError a))
+                   . (try :: IO α -> IO (Either IOError α))
```
saschalucas, line 209: Sorry for being late.

What has changed here? Latin letter `a` to Greek letter `α` (alpha)? Is this intentional? And if so, why?

apoikos: This is presumably to decouple the inner (`IO a`) from the outer (`Error a`) type, letting type inference do its magic. @mbakke Greek letters are great (obviously, I'm Greek :-) but it's rather unusual to encounter them in source code. I would stick with ASCII characters if possible, e.g. we could try using `a'` here.

src/Ganeti/BasicTypes.hs
```diff
@@ -206,12 +206,12 @@ instance MonadTrans (ResultT a) where
...
 
 instance (MonadBase IO m, Error a) => MonadBase IO (ResultT a m) where
   liftBase = ResultT . liftBase
                    . liftM (either (failError . show) return)
-                   . (try :: IO a -> IO (Either IOError a))
+                   . (try :: IO α -> IO (Either IOError α))
```
apoikos, line 214: Ditto

apoikos (changes requested):
@mbakke thanks for the PR and apologies for the extremely long delay. 

Other than `α`, I would just like to see a brief commit message describing the two changes.

saschalucas:
This PR is obsolete now by:

* commit e16148325e / PR #1740
* commit 9c51d6f1610c78d15e3edfc947813bc6456933e1 / PR #1759
* commit bf4e713466 / PR #1739

Thanks @mbakke for your work.

PR #1718, opened 2023-12-17
Fixes for running ganeti with --primary-ip-version=6

aurel32:


saschalucas (approved):
Thanks @aurel32, LGTM.

PR #1719, opened 2023-12-18
Fix flaky GetCmdline() tests in Github Action containers

rbott:
The test for utils.GetCmdline() sometimes returns an error because the process started by `subprocess.popen()` does not _yet_ return a valid command line via `/proc/$pid/cmdline` (just an empty string).

This does only seem to happen in containers running in the Github Actions environment as I was not able to reproduce this anywhere else. The "fix" is to wait for one second so things can settle.


saschalucas (approved):
LGTM

rbott:
Thanks @saschalucas !

PR #1721, opened 2024-03-14
Add devcontainer for developing Ganeti

codefritzel:
This PR adds a basic devcontainer config.
With the [Devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) standard, a dev environment can easily be created in the vscode editor or gh codespaces.

So you can develop and test within the donatiner without having to install any Haskell stuff on your own host. 
The ci-image from jammy is used as base image.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+{
+    "name": "Ganeti Dev Container",
+    "image": "ganeti/ci:jammy-py3",
+
+    "customizations":{
+        "vscode": {
+            "extensions": [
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "Ganeti Dev Container",
+    "image": "ganeti/ci:jammy-py3",
```
saschalucas, line 3: Would you mind to switch to Debian based defaults: bookworm-py3?

saschalucas (approved):
Thanks @codefritzel.

Issue #1710, opened 2023-07-03
Python 3.12 support


apoikos:
Python 3.12 is around the corner with [a lot of breaking changes](https://docs.python.org/3.12/whatsnew/3.12.html#removed). Some changes have already been dealt with, but we still need to tackle at least the following removals:

 - `distutils`: we have a single usecase in test/py/ganeti.utils.x509_unittest.py that should be easy to fix.
 - `asyncore`: despite #1701, asyncore is still in use by the http server (`noded`, `rapi`). Both of these are forking servers and asyncore was used only for signal handling. I've started work to replace the whole implementation using standard library primitives (socketserver, BaseHTTPRequestHandler).

saschalucas:
There is one to add #1764 : module imp is removed

saschalucas:
>     * `asyncore`: despite [Drop some asyncore-related code #1701](https://github.com/ganeti/ganeti/pull/1701), asyncore is still in use by the http server (`noded`, `rapi`). Both of these are forking servers and asyncore was used only for signal handling. I've started work to replace the whole implementation using standard library primitives (socketserver, BaseHTTPRequestHandler).

I don't know if this are good news, but there is now [pyasyncore](https://pypi.org/project/pyasyncore/) in [Debian Trixie/Sid](https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=asyncore) and also in [Ubuntu-24.04](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=asyncore).

For the moment Ganeti seems to work on Ubuntu-24.04 with python-3.12.

rbott:
asyncore dependency has been successfully removed, so we can close this issue.

PR #1722, opened 2024-03-28
drop distutils

codefritzel:
This PR removes distutils because it is deprecated and will be removed in 3.12.
See #1710

distutils was only used to check if the pyopenssl is greater than 0.7 in test/py/ganeti.utils.x509_unittest.py. 
version.parse from packaging is now used for the version comparison.

### DIFF
--- a/test/py/ganeti.utils.x509_unittest.py
+++ b/test/py/ganeti.utils.x509_unittest.py
@@ -34,9 +34,9 @@ import os
 import tempfile
 import unittest
 import shutil
-import OpenSSL
-import distutils.version
 import string
+import OpenSSL
+import packaging.version
 
 from ganeti import constants
 from ganeti import utils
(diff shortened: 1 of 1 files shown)


test/py/ganeti.utils.x509_unittest.py
```diff
@@ -34,9 +34,9 @@
 import tempfile
 import unittest
 import shutil
-import OpenSSL
-import distutils.version
 import string
+import OpenSSL
+from packaging import version
```
saschalucas, line 39: would you mind keep import namespace as it was with distutils: `import packaging.version`? and reference it later as `packaging.version.parse`? Being explicit seems better than implicit.

test/py/ganeti.utils.x509_unittest.py
```diff
@@ -78,10 +78,11 @@ class TestGetX509CertValidity(testutils.GanetiTestCase):
   def setUp(self):
     testutils.GanetiTestCase.setUp(self)
 
-    pyopenssl_version = distutils.version.LooseVersion(OpenSSL.__version__)
+    pyopenssl_version = version.parse(OpenSSL.__version__)
+    desired_pyopenssl_version = version.parse("0.7")
```
saschalucas, line 82: maybe switch the name from `desired` to `minimal`, this would better represent the need `>=`?

test/py/ganeti.utils.x509_unittest.py
codefritzel: I see no advantage with `import packaging.version` , except longer code.
It would only be better if more than one modules with the name "version" were imported.

test/py/ganeti.utils.x509_unittest.py
codefritzel: Thanks for the hint.
minimal_pyopenssl_version is a more meaningful variable name.

saschalucas (approved):
Thanks @codefritzel. LGTM

PR #1723, opened 2024-04-01
replace simplejson by the python build-in json module

codefritzel:
This PR replaces simplejson by the python build-in json module.
In the newer Python versions starting from 3.2, there is no reason to prefer simplejson to the build in
See https://bugs.python.org/issue7451

This removes simplejson as external dependency.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -36,7 +36,6 @@ Before installing, please verify that you have the following programs:
 - `ndisc6 <http://www.remlab.net/ndisc6/>`_ (if using IPv6)
 - `Python <http://www.python.org/>`_, version 3.6 or above
 - `Python OpenSSL bindings <https://www.pyopenssl.org/>`_
-- `simplejson Python module <https://simplejson.readthedocs.io/>`_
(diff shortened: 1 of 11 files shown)


rbott:
Hey @codefritzel, thanks for this contribution. This has been discussed for years, but nobody ever stepped up to implement it :-)

For the most part we can ignore the broken Test on Debian Testing (at least for _this_ PR as the problem is related to Haskell).

I have one minor point: would it be possible to re-implement the whole PR as a single (signed) commit, containing a meaningful headline and some extra description? That will make it way easier to craft the NEWS section for the next release so that distribution package maintainers can pick up this important change.

I will do some [QA runs](https://stargazer.ganeti.org/) against this PR so we can see if this change as any unwanted side effects. Unfortunately, hot-adding disks is currently broken with KVM on Debian Bookworm, so _this_ QA run will be somewhat expected to fail.

rbott:
Thanks a lot @codefritzel :-)

rbott (approved):

PR #1725, opened 2024-05-06
fix MNode in gnt-node list for >= Ubuntu-22.04

saschalucas:
Ganeti has used `Active` memory from `/proc/meminfo` to get how much memory a hypervisor is using (shown in `gnt-node list` fild `MNode`). As it turns out by upgrading from Ubuntu-20.04 to 22.04 the included Linux kernel does no longer satisfy this assumption. There is a big difference:

```
root@node:~# grep Active: /proc/meminfo
Active:          2019624 kB

root@node:~# free -k
               total        used        free      shared  buff/cache   available
Mem:      1055942572   164543996   890356300       10728     1042276   887191112
```

Active memory shows approx. 1.9 GiB, while in reality 157 GiB is used. And while at it, switch free memory calculation to newer metric `MemAvailable`[1], introduced in Linux-3.14.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773


codefritzel (approved):
The code looks good to me!

rbott:
LGTM in theory - I have started two QA runs to confirm (xen-pvm and KVM): https://stargazer.ganeti.org/

The KVM one will most likely fail as there are known issues with hot-adding disks on Debian Bookworm (which would be fixed by #1667, hint hint :smile:)

rbott (approved):

PR #1727, opened 2024-05-16
Make sure QA does not fail on empty hypervisor parameters

rbott:
The fake HV does not come with any hypervisor parameters, which breaks the QA suite when parsing the YAML output of `gnt-instance info $instance`.

This commit adds a safety check to catch this corner case.


rbott:
QA is running and looks good so far: https://stargazer.ganeti.org/0bcaed6d720aabcb0d2d579b6951d7886097e4b5/

saschalucas (approved):
Thanks @rbott . Please merge your changes.

PR #1729, opened 2024-05-18
Fix minor hotplug documentation issue

### LINKED ISSUE #823: wrong hotplug documentaton for 2.10
Originally reported of Google Code with ID 767.

```
i guess the second example should be something like this:

gnt-instance modify --disk add:size=1G --hotplug test
gnt-instance modify --disk 1:remove --hotplug test

instead of:

gnt-instance modify --disk add:size=1G --hotplug test
gnt-instance modify --net 1:remove --hotplug test

http://docs.ganeti.org/ganeti/2.10/html/design-hotplug.html#disk-hotplug


```
Originally added on 2014-03-20 09:43:38 +0000 UTC.

rbott:
This fixes a minor docs / example issue. Fixes #823

### DIFF
--- a/doc/design-hotplug.rst
+++ b/doc/design-hotplug.rst
@@ -222,7 +222,7 @@ support only disk addition/deletion.
 ::
 
  gnt-instance modify --disk add:size=1G --hotplug test
- gnt-instance modify --net 1:remove --hotplug test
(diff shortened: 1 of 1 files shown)


saschalucas (approved):
Thanks @rbott, please merge your changes.

Issue #1549, opened 2020-12-28
Notify KVM instances after growing disks


apoikos:
Disks are currently able to grow online, but KVM instances must be explicitly notified to allow the guest OS to use the extra capacity without a hypervisor restart. This is relatively simple to implement and should be enabled by default.

saschalucas:
this seems a duplicate of issue #28.

codefritzel:
closed by #1731

Issue #28, opened 2017-02-03
online disk grow/resizing with qemu/KVM


zitterbacke:
Hi,

I like to reactivate an old goole-code-issue-258 I've found.

I found online grow more useful then adding a disk with --hotplug. The main reason is, that I can extend an existing guest filesystem without the usage of LVM (BTW: and even without the usage of partitions). Without LVM you have a 1:1 relation of a disk and a filesystem (in all layers). This makes a transfer of a disk between VMs (disk de-/attach functionality) more handy for me.

My workflow I use often is (pseudo code):

if disk_template in ['plain','drbd']:
  gnt-instance grow-disk ...
elif disk_template in ['blockdev']:
  grow-externally
  
if disk_template in ['plain','drbd','blockdev']:
  run_qemu_monitor_command: info block
  run_qemu_monitor_command: block_resize hotdisk-7d8a2ae9-pci-7 <any_size_smaller_then_current>G
elif disk_template in ['sharedfile','file']:
  run_qemu_monitor_command: info block
  run_qemu_monitor_command: block_resize hotdisk-7d8a2ae9-pci-7 <new_size_>G

if disk_template in ['sharedfile','file','blockdev']:
  gnt-cluster repair-disk-sizes

Sounds like a simple task to add this into Ganeti? Will have a look at it, hopefully not touching Haskell code ;-)

Thanks, Alf.

kasimon:
A <pre>gnt-instance grow-disk --hotplug ...</pre> issuing the qemu commands above would indeed be a very useful addition to ganeti.

apoikos:
This is certainly something that is useful, regardless of hotplugging support, and relatively easy to implement.

anarcat:
> run_qemu_monitor_command: info block
> run_qemu_monitor_command: block_resize hotdisk-7d8a2ae9-pci-7 <any_size_smaller_then_current>G

i'm curious to hear more about this... everything i read online does this through `virsh` which, i suspect won't work with ganeti. what is that `run_qemu_monitor_command` command anyways?

Ganneff:
Actually, resizing works nicely. Even online, and for /. That is, for
us, in our setup. :)

Needed: The gnt-instance grow-disk must work, without impact on the
VM. It does so here with extstorage, and that is what this is tested
on. I do not use drbd or plain or blockdev or whatnot.

Also, my system disks are set as one big partition.

But then it is:

`gnt-instance grow-disk to whatever new size`

Now on the host the VM runs on:

[...]

PR #1731, opened 2024-06-10
kvm online disk resize via qmp

codefritzel:
This commit adds the ability to grow a disk of a kvm instance online without hook scripts. When the 'gnt-instance grow-disk' command is completed, an rpc call is called, which notifies the hypervisor of a disk change via qmp-socket.

This was realised with the new ‘abstracted‘ method 'ResizeDisk' in lib/hypervisor/hv_base.py, which was currently only implemented for the kvm hypervisor.

See #1549 & #28

### DIFF
--- a/lib/rpc_defs.py
+++ b/lib/rpc_defs.py
@@ -304,6 +304,11 @@ _INSTANCE_CALLS = [
     ("extra", None, "Extra info for device (dev_path for disk)"),
     ("seq", None, "Device seq"),
     ], None, None, "Hotplug a device to a running instance"),
+  ("resize_disk", SINGLE, None, constants.RPC_TMO_NORMAL, [
+    ("instance", ED_INST_DICT, "Instance object"),
+    ("disk", None, "The disk to be resized"),
+    ("new_size", None, "The new disk size in bytes"),
+    ], None, None, "Notify the HV about a disk resize"),
   ("hotplug_supported", SINGLE, None, constants.RPC_TMO_NORMAL, [
     ("instance", ED_INST_DICT, "Instance object"),
     ], None, None, "Check if hotplug is supported"),
--- a/lib/backend.py
+++ b/lib/backend.py
@@ -3180,6 +3180,20 @@ def GetMigrationStatus(instance):
     _Fail("Failed to get migration status: %s", err, exc=True)
 
 
+def ResizeDisk(instance, disk, new_size):
+  """Notify the hypervisor about a disk change.
+
+    @type disk: L{objects.Disk}
+    @param disk: the disk to be changed
+    @type new_size: int
+    @param new_size: the new size in bytes
+    @raise errors.HotplugError: if disk resize is not supported
+  """
+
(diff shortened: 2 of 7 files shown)


lib/rpc_defs.py
```diff
@@ -304,6 +304,11 @@ def _TestDelayTimeout(duration):
     ("extra", None, "Extra info for device (dev_path for disk)"),
     ("seq", None, "Device seq"),
     ], None, None, "Hoplug a device to a running instance"),
+  ("resize_disk", SINGLE, None, constants.RPC_TMO_NORMAL, [
+    ("instance", ED_INST_DICT, "Instance object"),
+    ("disk", None, "The dist to be resized"),
+    ("new_size", None, "The new disk size in bytes"),
+    ], None, None, "Notify the HV about a disk resize"),
```
saschalucas, line 311: For growing a disk, there is already `blockdev_grow` RPC. Wouldn't it make sense to also QMP-notify the hypervisor directly from there, without a new RPC?

lib/rpc_defs.py
```diff
@@ -304,6 +304,11 @@ def _TestDelayTimeout(duration):
     ("extra", None, "Extra info for device (dev_path for disk)"),
     ("seq", None, "Device seq"),
     ], None, None, "Hoplug a device to a running instance"),
+  ("resize_disk", SINGLE, None, constants.RPC_TMO_NORMAL, [
+    ("instance", ED_INST_DICT, "Instance object"),
+    ("disk", None, "The dist to be resized"),
```
saschalucas, line 309: typo dist -> disk

codefritzel:
@rbott can you please run this against the QA suite?

rbott:
@codefritzel started, you can find the logs/output here once it's done: https://stargazer.ganeti.org/ (might take ~4-5 hours unless it breaks)

lib/rpc_defs.py
saschalucas: Just for the record: it has been discussed somewhere else, that the `blockdev_grow` RPC is called several times (dry-run, primary and in case of DRBD secondary node). Technically it deals with storage only an not hypervisor stuff. So the extra RPC seems fine here.

lib/rpc_defs.py
```diff
@@ -304,6 +304,11 @@ def _TestDelayTimeout(duration):
     ("extra", None, "Extra info for device (dev_path for disk)"),
     ("seq", None, "Device seq"),
     ], None, None, "Hotplug a device to a running instance"),
+  ("resize_disk", SINGLE, None, constants.RPC_TMO_NORMAL, [
+    ("instance", ED_INST_DICT, "Instance object"),
+    ("disk", None, "The dist to be resized"),
```
saschalucas, line 309: there is still the little typo dist -> disk in here

saschalucas (approved):
Ganeti QA Run was successful, just fix the typo, after that LGTM.

codefritzel:
Thanks for testing and the review!

PR #1734, opened 2024-06-18
Network.poolLens, poolArrayLens: reactivate Lens' type signature

thielema:
readPool, readPoolE: avoid use of function composition (.) since its type does not match the forall quantifier hidden in the `Lens'` type synonym.

This reverts 5e30bad1bba63c9f6c782003ef2560f107a0ba24

### DIFF
--- a/src/Ganeti/Network.hs
+++ b/src/Ganeti/Network.hs
@@ -87,11 +87,11 @@ data PoolPart = PoolInstances | PoolExt
 addressPoolIso :: Iso' AddressPool BA.BitArray
 addressPoolIso = iso apReservations AddressPool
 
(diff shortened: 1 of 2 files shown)


thielema:
Alternatively we could use these type signatures:
~~~~
poolLens :: (Functor f) => PoolPart -> LensLike' f Network (Maybe AddressPool)
poolArrayLens ::
  (Functor f) => PoolPart -> LensLike' f Network (Maybe BA.BitArray)
~~~~
and leave the function composition (.) at the call sites as is.

saschalucas (approved):
Thanks @thielema. I'm fine with fixing this on the caller's side.

Just a minor note for future commits: please use `git commit --signoff`. Ganeti does not strictly depend on this, but its seems a common practice, maybe regarding US law.

PR #1735, opened 2024-06-19
reduce redundant Haskell module imports

thielema:


src/Ganeti/JSON.hs
```diff
@@ -109,7 +109,7 @@ import Text.JSON.Pretty (pp_value)
 -- itself to build many other modules.
 
 import Ganeti.BasicTypes
-import Ganeti.Compat
+import Ganeti.Compat ()
```
saschalucas, line 112: just for curiosity and for non Haskell programmers: what's the difference between `import something` and `import something ()`

src/Ganeti/Metad/ConfigServer.hs
```diff
@@ -35,7 +35,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 module Ganeti.Metad.ConfigServer where
 
 import Control.Exception (finally)
-import Control.Monad.Reader
+import Control.Monad (forever)
```
saschalucas, line 38: Also for curiosity, what happens here: `Reader` -> `(forever)`?

saschalucas (approved):
Thanks @thielema. If you have the time, please explain very short my two comments. Otherwise LGTM.

thielema:
On Wed, 19 Jun 2024, saschalucas wrote:

> @saschalucas commented on this pull request.
> 
> ______________________________________________________________________________________________________________________________
> 
> In src/Ganeti/JSON.hs:
> 
> > @@ -109,7 +109,7 @@ import Text.JSON.Pretty (pp_value)
>  -- itself to build many other modules.
>
>  import Ganeti.BasicTypes
> -import Ganeti.Compat
> +import Ganeti.Compat ()
> 
> just for curiosity and for non Haskell programmers: what's the difference between import something and import something ()

We need to import the instance MonadFail JSON.Result that is conditionally 
defined in Ganeti.Compat for json>=0.10. However, we do not need any 
visible identifier from Ganeti.Compat. Thus GHC warns about an apparently 
unused import. To silence that, I added the parentheses. It means: We do 
not need any identifier, but maybe some instances.

So this means to other programmers: Do not remove that import, even if the 
compilation works. It might fail in other circumstances, since the 
instance is orphan and some reordering of imports might bring it into 
scope or not depending on how depending modules import. Ok, I am afraid 
the complete explanation would require an excursus on orphan instances ...

But your question is very valid and gives me the idea, that we should add 
a comment like this one:


import Ganeti.Compat () -- instance MonadFail JSON.Result

thielema:
On Wed, 19 Jun 2024, saschalucas wrote:

> @saschalucas commented on this pull request.
> 
> ______________________________________________________________________________________________________________________________
> 
> In src/Ganeti/Metad/ConfigServer.hs:
> 
> > @@ -35,7 +35,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
>  module Ganeti.Metad.ConfigServer where
>
>  import Control.Exception (finally)
> -import Control.Monad.Reader
> +import Control.Monad (forever)
> 
> Also for curiosity, what happens here: Reader -> (forever)?

The modules from MTL (Monad Template Library) used to re-export also 
generic helper functions from Control.Monad. In this case, no Reader monad 
was used so I replaced it by the import of just the generic helper 
function 'forever'.

PR #1738, opened 2024-06-19
src/Ganeti: replace LANGUAGE RankNTypes -> Rank2Types

thielema:
Rank2Types is the less demanding Haskell type extension.


saschalucas (approved):
Thanks @thielema.

PR #1740, opened 2024-06-19
Control.Monad.Error -> Control.Monad.Except

thielema:
for compatibility with transformers-0.6

### DIFF
--- a/cabal/ganeti.template.cabal
+++ b/cabal/ganeti.template.cabal
@@ -52,7 +52,7 @@ library
     , deepseq                       >= 1.3.0.0
     , directory                     >= 1.1.0.2
     , filepath                      >= 1.3.0.0
-    , mtl                           >= 2.1.1
+    , mtl                           >= 2.2.1
     , old-time                      >= 1.1.0.0
     , pretty                        >= 1.1.1.0
     , process                       >= 1.1.0.1
--- a/src/Ganeti/BasicTypes.hs
+++ b/src/Ganeti/BasicTypes.hs
@@ -48,7 +48,7 @@ module Ganeti.BasicTypes
   , toErrorBase
   , toErrorStr
   , tryError
-  , Error(..) -- re-export from Control.Monad.Error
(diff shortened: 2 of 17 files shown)


saschalucas:
Thanks @apoikos for self assigning review.

@thielema: it seems we have a merge conflict now, probably due to PR #1735 now being merged into master. Conflicts only appear around `import` and should be easily resolvable.

thielema:
Explanation for the patch:
`Control.Monad.Error` was a way to add exceptions with explicit types to monads. It wanted to support the `Monad.fail :: String -> m a` method, that takes a `String` parameter as an exception. To this end `Monad.Error` required the `strMsg :: String -> a` method to convert an anonymous `String` to your custom exception type `e`.
Over the years it became clear, that the classic `Monad` type class wanted too much at once. So `Applicative` class was introduced as an addition and `Monad` became subclass of `Applicative` and `Functor`. Eventually, `fail` has been moved from the classic `Monad` type class to the new `MonadFail` type class. (Next issue on the Haskeller's ToDo list is the "Monad of no Return" proposal.) There is no longer a need to implement a `fail` method for every monad and thus there is no strict need for `strMsg` anymore. So the packages `transformers` and `mtl` added `Control.Monad.Except` as a `fail`- and `strMsg`-free alternative to `Control.Monad.Error`.
Finally, with `transformers-0.6` the `Control.Monad.Error` module has been removed. However, Ganeti used that module and it also used the `strMsg` method. Thus I added a custom definition of the `Error` class with its `strMsg` method to `BasicTypes`.

saschalucas:
Thanks @apoikos for taking review. Please note, that due to vacation period an answer could take some time. We've planned to continue around July the 15th. Maybe a view days earlier...

saschalucas (approved):
I'm going to approve this now. We have "In principle it looks good" and "raise the lower limit for mtl ... to 2.2.1" satisfied, both as reviewed by @apoikos.

PR #1743, opened 2024-06-19
avoid warnings on use of the partial function `head`

thielema:
Those warnings are issued by GHC-9.10 and in principle also by HLint, if they were not silenced.


thielema:
re-run tests

saschalucas (approved):
Thanks @thielema. In absence of true understanding for all changes here I'll merge this to proceed further.

PR #1745, opened 2024-06-20
implement 'return' using 'pure' to silence warnings in GHC-9.10

thielema:
in anticipation of the "Monad of no Return" proposal


apoikos (approved):
LGTM, thanks!

PR #1756, opened 2024-07-19
Replace query-cpus with query-cpus-fast

codefritzel:
This replaces the query-cpus qmp command with query-cpus-fast. The query-cpus command can interrupt running cpus, which can cause performances problems.

See https://patchwork.kernel.org/project/qemu-devel/patch/20210318092512.250725-6-berrange@redhat.com


saschalucas (approved):
Thanks @codefritzel

Issue #1755, opened 2024-07-18
Makefile.ghc generation broken in Ubuntu Noble (24.04 / GHC-9.4)?


saschalucas:
As noted in #1748 there is a Type error exposed using `make` and not `Cabal`. With this workaround:

```diff
diff --git a/src/Ganeti/Lens.hs b/src/Ganeti/Lens.hs
index c6184ea..32fcc2f 100644
--- a/src/Ganeti/Lens.hs
+++ b/src/Ganeti/Lens.hs
@@ -93,14 +93,14 @@ makeCustomLenses' name lst = makeCustomLensesFiltered f name
 -- Most often the @g@ functor is @(,) r@ and 'traverseOf2' is used to
 -- traverse an effectful computation that also returns an additional output
 -- value.
-traverseOf2 :: Over (->) (Compose f g) s t a b
-            -> (a -> f (g b)) -> s -> f (g t)
+-- traverseOf2 :: Over (->) (Compose f g) s t a b
+--             -> (a -> f (g b)) -> s -> f (g t)
 traverseOf2 k f = getCompose . traverseOf k (Compose . f)
 
 -- | Traverses over a composition of a monad and a functor.
 -- See 'traverseOf2'.
-mapMOf2 :: Over (->) (Compose (WrappedMonad m) g) s t a b
-        -> (a -> m (g b)) -> s -> m (g t)
+-- mapMOf2 :: Over (->) (Compose (WrappedMonad m) g) s t a b
+--         -> (a -> m (g b)) -> s -> m (g t)
 mapMOf2 k f = unwrapMonad . traverseOf2 k (WrapMonad . f)
 
 -- | A helper lens over sets.
```
compilation goes further until:
```
[GHC]: src/mon-collector.o <- cabal_macros.h

src/mon-collector.hs:38:1: error:
    Could not find module ‘Ganeti.DataCollectors.CLI’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
38 | import Ganeti.DataCollectors.CLI (genericOptions, defaultOptions)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/mon-collector.hs:39:1: error:
    Could not find module ‘Ganeti.DataCollectors.Program’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
[...]

PR #1763, opened 2024-07-27
generate Makefile.ghc by running `ghc -M` individually for every executable

thielema:
This is a work-around for https://gitlab.haskell.org/ghc/ghc/-/issues/25113.

It should fix https://github.com/ganeti/ganeti/issues/1755.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1348,9 +1348,19 @@ HS_MAKEFILE_GHC_SRCS += $(HS_TEST_PROGS:%=%.hs)
 endif
 Makefile.ghc: $(HS_MAKEFILE_GHC_SRCS) Makefile $(HASKELL_PACKAGE_VERSIONS_FILE) \
               | $(built_base_sources) $(HS_BUILT_SRCS)
-	$(GHC) -M -dep-makefile $@ $(DEP_SUFFIXES) $(HFLAGS) $(HFLAGS_DYNAMIC) \
-		-itest/hs \
-		$(HEXTRA_COMBINED) $(HS_MAKEFILE_GHC_SRCS)
+# Run dependency generation for every executable.
+# Up to GHC-9.2 we could perform this in one run of 'ghc -M'.
+# Due to https://gitlab.haskell.org/ghc/ghc/-/issues/25113
+# we have to process every executable manually now.
+# This way, many dependencies in Makefile.ghc are duplicate now
+# but 'make' should cope with this.
+	-mv $@ $@.bak
+	for prog_hs in $(HEXTRA_COMBINED) $(HS_MAKEFILE_GHC_SRCS); do \
+		$(GHC) -M -dep-makefile $@.part $(DEP_SUFFIXES) $(HFLAGS) $(HFLAGS_DYNAMIC) \
+			-itest/hs $$prog_hs ; \
+		cat $@.part >>$@ ; \
+		rm $@.part ; \
+	done
 # Since ghc -M does not generate dependency line for object files, dependencies
 # from a target executable seed object (e.g. src/hluxid.o) to objects which
 # finally will be linked to the target object (e.g. src/Ganeti/Daemon.o) are


Makefile.am
```diff
@@ -1348,9 +1348,18 @@ HS_MAKEFILE_GHC_SRCS += $(HS_TEST_PROGS:%=%.hs)
...
-	$(GHC) -M -dep-makefile $@ $(DEP_SUFFIXES) $(HFLAGS) $(HFLAGS_DYNAMIC) \
-		-itest/hs \
-		$(HEXTRA_COMBINED) $(HS_MAKEFILE_GHC_SRCS)
+# Run dependency generation for every executable.
+# Up to GHC-9.2 we could perform this in one run of 'ghc -M'.
+# Due to https://gitlab.haskell.org/ghc/ghc/-/issues/25113
+# we have to process every executable manually now.
+# This way, many dependencies in Makefile.ghc are duplicate now
+# but 'make' should cope with this.
+	-mv $@ $@.bak
```
saschalucas, line 1357: for what purpose is the `.bak` file here?

Makefile.am
```diff
@@ -1348,9 +1348,18 @@ HS_MAKEFILE_GHC_SRCS += $(HS_TEST_PROGS:%=%.hs)
...
+	for prog_hs in $(HEXTRA_COMBINED) $(HS_MAKEFILE_GHC_SRCS); do \
+		$(GHC) -M -dep-makefile $@.part $(DEP_SUFFIXES) $(HFLAGS) $(HFLAGS_DYNAMIC) \
+			-itest/hs $$prog_hs ; \
+		cat $@.part >>$@ ; \
```
saschalucas, line 1361: Would it hurt to replace `cat` by `sort -u` as you suggested in #1755?

And is `>>` the automake version of normal shell output redirection `>`, or is this a shell append? May be related to why `mv` is used above, if this a append?

thielema:
On Sat, 27 Jul 2024, saschalucas wrote:

> for what purpose is the .bak file here?

I simulate what `ghc -M` does. It does create a .bak file as a first step. 
You also already find Makefile.ghc.bak in the list of temporary files.

thielema:
On Sat, 27 Jul 2024, saschalucas wrote:

> In Makefile.am:
> 
> +	for prog_hs in $(HEXTRA_COMBINED) $(HS_MAKEFILE_GHC_SRCS); do \
> +		$(GHC) -M -dep-makefile ***@***.*** $(DEP_SUFFIXES) $(HFLAGS) $(HFLAGS_DYNAMIC) \
> +			-itest/hs $$prog_hs ; \
> +		cat ***@***.*** >>$@ ; \
> 
> Would it hurt to replace cat by sort -u as you suggested in #1755?

Sorting individual dependency lists makes no difference. I would have to 
sort all dependency blocks together.

> And is >> the automake version of normal shell output redirection >, or is this a shell append?

It is plain shell append. I append the outputs of every run of `ghc -M`.

If I manage to write the dependencies to stdout I could also put the 
for-loop in parentheses and pipe everything in one go to Makefile.ghc, and 
I could also apply `sort --unique` as a last step.

saschalucas:
> I simulate what `ghc -M` does. It does create a .bak file as a first step. You also already find Makefile.ghc.bak in the list of temporary files.

Thanks for clarification. This explains why there is also `Makefile.ghc.part.bak` now. I'm not sure if we need to add this also to temporary files? 

> Sorting individual dependency lists makes no difference.

Of course, my fault not realizing that `cat` is inside the loop.

> If I manage to write the dependencies to stdout I could also put the for-loop in parentheses and pipe everything in one go to Makefile.ghc, and I could also apply `sort --unique` as a last step.

That would be a very elegant way. However don't investigate to much time in that. It's more important to have a workaround as it is.

thielema:
On Sat, 27 Jul 2024, saschalucas wrote:

> Thanks for clarification. This explains why there is also 
> Makefile.ghc.part.bak now. I'm not sure if we need to add this also to 
> temporary files?

Right. Either making it a temporary file or, I think better, deleting it 
after appending.

saschalucas (approved):
Thanks @thielema.

PR #1767, opened 2024-07-29
Makefile.am: remove warning on Monomorphism restriction

thielema:
These warnings flood compiler output with no apparent value.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -812,7 +812,6 @@ HS_BUILT_TEST_HELPERS = $(HS_BIN_ROLES:%=test/hs/%) test/hs/hail
 
 HFLAGS = \
 	-O -Wall -isrc \
-	-fwarn-monomorphism-restriction \
 	-fwarn-tabs \
 	-optP-include -optP$(HASKELL_PACKAGE_VERSIONS_FILE) \
 	-hide-all-packages \


saschalucas (approved):
thanks @thielema. LGTM.

Just BTW: why has this warning no value? Sorry for asking :-)

thielema:
On Tue, 30 Jul 2024, saschalucas wrote:

> @saschalucas approved this pull request.
> 
> thanks @thielema. LGTM.
> 
> Just BTW: why has this warning no value? Sorry for asking :-)

Short answer: If it would have a value, the original authors would have 
added type signatures to satisfy the warner.

Medium size answer: Adding the type signatures does not make the code 
safer, maybe more comprehensible, but sometimes for the price of requiring 
another language extension (ScopedTypeVariables). The omitted type 
signatures do not allow for additional type errors. I don't consider the 
indicated source locations as written in bad style. However, the memory 
consumption can differ between monomorph and polymorph. But so far the 
users seemed to be happy with the memory allocation.

Long answer: In Haskell a type can be monomorph or polymorph. Example:

Monomorph:
~~~~
2 :: Integer
~~~~

Polymorph:
~~~~
2 :: Num a => a
~~~~
The polymorph type signature means: 
"Two" is a value of every numeric type "a".

The polymorphically typed value is of course more flexible in its use. But 
the drawback is that polymorph values are not stored once evaluated, 
because next time it might be evaluated with a different type.

E.g. compare
~~~~
primesMono :: [Integer]
~~~~
vs.
~~~~
primesPoly :: (Integral a) => [a]
~~~~

The list of primes is infinite, you may evaluate some prefix. You may want 
to share the computed primes between different evaluations, because you 
may find the computation of primes to be expensive. In this case you would 
choose primesMono. But it might also be that you want to traverse the list 
of primes only once, in which case storing the evaluated prefix would 
cause ever growing memory consumption. If you want to traverse the list 
of primes multiple times but with different types (bad example, because 
other integral types like Int have limited value range), then you would 
choose primesPoly.

The Monomorphism Restriction applies to expressions where you do not give 
the type signature explicitly. The Haskell compiler has to choose 
automatically between monomorph and polymorph type then and it might do so 
differently from what you expected. You could try to disambiguate by 
adding an explicit type signature, but this is not alway possible using 
only the tools of standard Haskell 98. E.g. a local function in a 
top-level function might require types, that are not monomorphic but also 
not fully polymorphic, that is, independent from the outer function. Many 
local functions "inherit" types from their parent function. In these cases 
you could use the ScopedTypeVariables extension to specify which type 
variables to inherit and which ones shall be fully polymorphic.

Issue #1747, opened 2024-06-20
ganeti.cabal: static vs. generated


thielema:
Currently `ganeti.cabal` is constructed from `cabal/ganeti.template.cabal` by `cabal/cabal-from-modules.py`. But for construction `make` seems to already need to run Haskell code, using the incomplete `cabal/ganeti.template.cabal`. In the end the Haskell code seems to be build without `Cabal` but with `Makefile.ghc`. This looks all pretty weird. I'd prefer to have a static `ganeti.cabal` file. This would allow for better structuring of the package description. E.g. `htest` could be a `Test-Suite` instead of a plain `Executable`. The test modules could be part of the Test-Suite and would be hidden from the public Ganeti Haskell API.
I guess, constructing `ganeti.cabal` by a script intended to reduce redundancies that are in the final `ganeti.cabal` file. But the redundancies are acceptable. If not, we have common stanzas since Cabal-2.2. [1]
More advanced would be to generate `ganeti.cabal` from a Yaml file using `hpack`. This is an established solution [2], supported by tools like `stack` and `cabal2nix`. I guess, `cabal/ganeti.template.cabal` predates both Cabal-2.2 and `hpack`.

[1] https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#common-stanzas
[2] https://packages.debian.org/bookworm/hpack

thielema:
The Python code could go in a Pip package or the Python source could be included in the Cabal package as `Extra-Source-Files`.

thielema:
The Cabal file could look like this:
~~~~
cabal-version:       2.4
name:                ganeti
version:             3.0
homepage:            http://www.ganeti.org
license:             BSD-2-Clause
license-file:        COPYING
author:              Google Inc.
maintainer:          ganeti-devel@googlegroups.com
copyright:           2006-2015 Google Inc.
category:            System
build-type:          Simple
extra-source-files:  README
synopsis:            Cluster-based virtualization management software
description:
  Cluster-based virtualization management software
  .
  See <http://www.ganeti.org>


flag mond
  description: enable the ganeti monitoring daemon
  default:     True

flag metad
  description: enable the ganeti metadata daemon
  default:     True

-- superseeded by --enable-tests
flag htest
  description: enable tests
  default:     True

flag mcio_xformers
  description: use MonadCatchIO-transformers (deprecated)
  default:     True

flag network_bsd
  description: use the split network-bsd package
  default:     False

library
  exposed-modules:
[...]

PR #1768, opened 2024-07-29
ganeti.cabal: make a static file

thielema:
This addresses https://github.com/ganeti/ganeti/issues/1747.

Can we run this commit against an automated build of the generated source tarball?


thielema:
Ok, this cannot work. CabalDependenciesMacros would have to respect the flags that enable particular executables in order to determine the set of dependencies accurately. So I am afraid I must take two steps at once, that is, static ganeti.cabal and building with Cabal.

PR #1770, opened 2024-07-29
use full image uri for devcontainers

codefritzel:
With this change, container-engines other than Docker (e.g. Podman) can also be used.


saschalucas (approved):
Thanks @codefritzel

PR #1772, opened 2024-07-30
hluxid -> ganeti-luxid, hconfd -> ganeti-confd

thielema:
hluxid and hconfd were build and then copied over to ganeti-luxid and ganeti-confd.
Seems to be a historical leftover.


saschalucas (approved):
Thanks @thielema. I have no glue why this was copied ...

PR #1787, opened 2024-08-13
Test.Ganeti.Utils: correct instance Arbitrary ClockTime

thielema:
There was one digit missing to 10^12-1.


thielema:
please squash commits, the second one only fulfills 80-character-width constraint

Issue #1778, opened 2024-08-02
Test failure timediffAdd


thielema:
I got the following test failure:
~~~~
Utils:
  ...
  timediffAdd: [Failed]
*** Failed! Falsified (after 115 tests):
Sat Dec 29 00:04:07 CET 2018
Tue Apr 13 09:06:29 CEST 2021
Sun Mar 27 04:22:37 CEST 2022
Expected equality, but got mismatch
expected: Sun Mar 27 04:22:37 CEST 2022
 but got: Sun Mar 27 05:22:37 CEST 2022
(used seed -3298211784409845898)
~~~~

Without further investigation it looks like a problem with daylight saving time.

saschalucas:
I've seen this error only once. It seems not to happen often.

At the same time where this error occurred, the test py_compat_types from issue #1753 didn't fail. Any other run py_compat_types has failed.

thielema:
The error is strange. It happens at 2022-03-27, this is the day were daylight saving time started in 2022. If the bug is about daylight saving time this would explain why test failure is so rare. However time zones switch from 2:00 CET to 3:00 CEST, but 4:22 and 5:22 are after that point. Surprisingly the problem arises with ClockTime, which carries no time zone, at all, but measures seconds and picoseconds. So maybe daylight saving time is a red herring.

According to the comments the test wants to check a work-around for `old-time:diffClockTimes`. Original `diffClockTimes` seems to have a bug, where difference between two `ClockTimes` can have a negative fractional parts of the seconds. GHC bug tracker says, it will not be fixed, because `old-time` is abandoned.
https://gitlab.haskell.org/ghc/ghc/-/issues/2519

They suggest to move to the new `time` library. This one does not provide `ClockTime` anymore, but measures system time with type `UTCTime`. `ClockTime` is used in many places in Ganeti. A clean solution would certainly be to replace use of `old-time` entirely by `time`. There are still chances that all used `old-time` functions have counterparts in `time`.

thielema:
Ok, with full picosecond range I get failures like this one:
~~~~
  timediffAdd: [Failed]
*** Failed! Falsified (after 285 tests):
Tue Mar  4 16:41:38 CET 2008
Sat Nov 21 19:08:16 CET 2009
Wed Jul 11 01:46:52 CEST 2035
Expected equality, but got mismatch
expected: Wed Jul 11 01:46:52 CEST 2035
 but got: Wed Jul 11 02:46:52 CEST 2035
[...]

PR #1790, opened 2024-08-15
Improve tests on ClockTime

thielema:
Orphan instances are discouraged because they may conflict with instances of same type class and type defined by others and you cannot control what and whether instances are imported.

Please leave the individual commits intact.


saschalucas (approved):
Thanks @thielema

saschalucas:
This also improves but does not solve #1778, right?

thielema:
On Thu, 15 Aug 2024, saschalucas wrote:

> This also improves but does not solve #1778, right?

yes, but the fix is already prepared, I just waited for approval of the 
former steps

PR #1792, opened 2024-08-15
Utils.TimeDiff: simplified custom definition

thielema:
that avoids bug in old-time:System.Time.addToClockTime

solves https://github.com/ganeti/ganeti/issues/1778


saschalucas (approved):
Thanks @thielema.

PR #1793, opened 2024-08-15
Utils.Time: extracted from Utils

thielema:
I think the number of time related functions in `Ganeti.Utils` justifies a dedicated module.


saschalucas (approved):
Thanks @thielema.

PR #1786, opened 2024-08-10
allow three alternative RegEx implementations: regex-pcre, regex-pcre-builtin, regex-tdfa

thielema:
regex-pcre remains the default
Query.RegEx: module providing a uniform interface across RegEx implementations

I still have to find out, how to make a `configure` option for those alternatives.

PR #1794, opened 2024-08-18
Allow three alternative RegEx implementations: regex-pcre, regex-pcre-builtin, regex-tdfa

thielema:
I cannot re-open https://github.com/ganeti/ganeti/pull/1786, again. :-(


saschalucas (approved):
Thanks @thielema.

Issue #1753, opened 2024-07-11
test failure: py_compat_types


attila-lendvai:
i'm trying to revive the Guix package of Ganeti, but the `py_compat_types` test fails. notice the discrepancy between the two float formats at the very beginning.

```
  py_compat_types: [Failed]
Different result after encoding/decoding
[...]

PR #1795, opened 2024-08-18
OpCodes.opSummaryVal.opDelayDuration: format using printf not 'show'

thielema:
This might solve #1753.
Tests should be run multiple times in order to see, whether the problem actually disappears.

### DIFF
--- a/src/Ganeti/OpCodes.hs
+++ b/src/Ganeti/OpCodes.hs
@@ -63,6 +63,7 @@ import Data.List (intercalate)
 import Data.Map (Map)
 import qualified Text.JSON
 import Text.JSON (readJSON, JSObject, JSON, JSValue(..), fromJSObject)
+import Text.Printf (printf)
 
 import qualified Ganeti.Constants as C
 import qualified Ganeti.Hs2Py.OpDoc as OpDoc
@@ -1046,7 +1047,7 @@ opSummaryVal OpBackupExport { opInstanceName = s } = Just s
 opSummaryVal OpBackupRemove { opInstanceName = s } = Just s
 opSummaryVal OpTagsGet { opKind = s } = Just (show s)
 opSummaryVal OpTagsSearch { opTagSearchPattern = s } = Just (fromNonEmpty s)
-opSummaryVal OpTestDelay { opDelayDuration = d } = Just (show d)
+opSummaryVal OpTestDelay { opDelayDuration = d } = Just (printf "%f" d)
 opSummaryVal OpTestAllocator { opIallocator = s } =
   -- FIXME: Python doesn't handle None fields well, so we have behave the same
   Just $ maybe "None" fromNonEmpty s


thielema:
You can run that specific test:
~~~~
dist/build/htest/htest --select-tests=py_compat_types
~~~~
or with `cabal` if it is not compiled and you have internet access to https://hackage.haskell.org/:
~~~~
cabal run -- htest --select-tests=py_compat_types
~~~~

thielema:
The proposed fix uses `printf` with format `"%f"`, thus it will always format with fixed point representation. However, this fix point representation may still not always match the formatting of the same number done by Python. In order to format a double precision binary IEEE floating point number exactly you may need 53 decimal places or more. Instead, what decimal formatting routines usually do: Format with as many decimal places needed to uniquely identify the original binary fraction. Different implementation may choose different number of decimal places for this purpose.
To make the test robust, we would need to decide on a number of decimal places for both Haskell and Python formatting implementations.

thielema:
On Sun, 18 Aug 2024, thielema wrote:

> To make the test robust, we would need to decide on a number of decimal 
> places for both Haskell and Python formatting implementations.

Alternatively we could show the binary representation of the IEEE float.

thielema:
On Mon, 19 Aug 2024, thielema wrote:

> Alternatively we could show the binary representation of the IEEE float.

Like so:

Prelude> decodeFloat (3.338214845516507e-2::Double)
(4810874602992256,-57)

This means 4810874602992256 * 2^(-57) and could be formatted this way.

However, this representation must also be normalized. E.g. the value could 
also be written as: 37584957835877 * 2^(-50)

thielema:
On Sun, 18 Aug 2024, thielema wrote:

> To make the test robust, we would need to decide on a number of decimal 
> places for both Haskell and Python formatting implementations.

The summary TestDelay duration is formatted in opcodes.py/class 
OpCode/Summary with this code:

       txt = "%s(%s)" % (txt, field_value)

Thus, like the Haskell formatting, it relies on some default formatting 
behavior. The comments say that you may change the default formatter to a 
programmer defined one with the class variable OP_DSC_FORMATTER. However, 
I do not see any use of it in opcodes.py.

thielema:
Ok, after running htest 1000  times, I got this deviation, this time with Python and Haskell's role swapped:

Haskell:
~~~~
*Main> Text.Printf.printf "%f" (6.507186432525014e-05::Double)
0.00006507186432525014
~~~~
Python:
~~~~
>>> 6.507186432525014e-05
6.507186432525014e-05
~~~~

In my next attempt I will use a fixed precision, say `"%.10f"`.

thielema:
The Python classes for handling OpCodes are generated in `Ganeti.Hs2Py.GenOpCodes.showPyClass`. However, there is simply no code to generate an `OP_DSC_FORMATTER`.

thielema:
@iustin Hi Iustin, how was the `OP_DSC_FORMATTER` field supposed to be set? I do not see `OP_DSC_FORMATTER` in `Ganeti.Hs2Py.GenOpCodes.showPyClass`, at all.

thielema:
I see, when `OP_DSC_FORMATTER` was introduced, `opcodes.py` was included in the repository (maybe not even autogenerated?) and contained a custom implementation of `OpTestDelay.OP_DSC_FORMATTER`. This was still problematic, since it was a default formatting of a floating point value, but at this stage it would have been easy to use a specific formatting string like "%.9f".

thielema:
Withdrawn in favor of https://github.com/ganeti/ganeti/pull/1798.

PR #1797, opened 2024-08-20
Test delay fixed precision

thielema:
I am now using formatted printing with format "%.9f" (nanoseconds) both in Haskell and Python.


thielema:
Withdrawn in favor of https://github.com/ganeti/ganeti/pull/1798.

PR #1798, opened 2024-08-20
Test delay nanoseconds

thielema:
try with integer numbers

still relies on the same `round` functionality in Haskell and Python


thielema:
please leave the two commits intact

thielema:
I am still running the according test for some time, then I will release the request.

saschalucas (approved):
Thanks @thielema.

PR #1800, opened 2024-08-21
add .github/workflows to makelfile exclude

codefritzel:
.github/workflows was added to DIRCHECK_EXCLUDE in the makefile.

Now the actions can be executed locally (e.g. with act[1]) without losing the .github directory in the container If the actions failed in check-local, the directory has been lost with act.

[1] https://github.com/nektos/act


saschalucas (approved):
Thanks @codefritzel . Please merge your changes.

PR #1801, opened 2024-08-22
add Haskel lib regex-tdfa to CI matrix

saschalucas:
Recently regex-pcre is no longer available in Debian testing (Trixi) and an alternative regex-tdfa has been implemented, but at the coast of losing PCRE regex in favor of POSIX. Adding regex-tdfa to the CI matrix. This means that pcre and tdfa are tested both (where possible) for a transitional period.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,11 +34,11 @@ jobs:
         run: ./autogen.sh
 
       - name: configure
-        if: ${{ matrix.regex = 'pcre' }}
+        if: ${{ matrix.regex == 'pcre' }}
         run: ./configure --enable-haskell-tests
 
       - name: configure
-        if: ${{ matrix.regex = 'tdfa' }}
+        if: ${{ matrix.regex == 'tdfa' }}
         run: ./configure --enable-haskell-tests --enable-regex-tdfa
 
       - name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ permissions: read-all
 
 jobs:
   build:
-    name: Build on ${{ matrix.os }}
+    name: Build on ${{ matrix.os }}-${{ matrix.regex }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


saschalucas:
This is a draft. I have no glue if the syntax is correct. I'm still looking for a way to test this before asking for review.

thielema:
On Thu, 22 Aug 2024, saschalucas wrote:

> Recently regex-pcre is no longer available in Debian testing (Trixi) and 
> an alternative regex-tdfa has been implemented, but at the coast of 
> losing PCRE regex in favor of POSIX. Adding regex-tdfa to the CI matrix. 
> This means that pcre and tdfa are tested both (where possible) for a 
> transitional period.

Where can we lobby for adding regex-pcre-builtin to Debian? This would 
give us Perl Regexes in newer Debian releases.

rbott:
Paging @apoikos or @anarcat on this one :-)

anarcat:
i have zero idea, not very familiar with the haskell packaging, but i can ask around.

anarcat:
so the word from the world of irc.oftc.net (#debian-haskell FTW), is that regex-pcre is missing from debian because it depends on pcre3 which we're trying to remove from Debian (see [1000067](https://bugs.debian.org/1000067)).

According to a person in that channel, it might also be the case with regex-pcre-builtin -- it just vendors the library, which makes it altogether unsuitable for Debian -- but that should be double-checked.

thielema:
On Thu, 22 Aug 2024, anarcat wrote:

> According to a person in that channel, it might also be the case with 
> regex-pcre-builtin -- it just vendors the library, which makes it 
> altogether unsuitable for Debian -- but that should be double-checked.

If using regex-pcre-builtin poses a security risk, then I withdraw my 
proposal.

codefritzel:
Hi @saschalucas,
how about this?
```
      - name: configure
        run: |
          if [[ $REGEX == 'tdfa' ]]; then
            ./configure --enable-haskell-tests --enable-regex-tdfa
          else
            ./configure --enable-haskell-tests
          fi
        env: 
          REGEX: ${{ matrix.regex }}
```
So the configure task would work with a single job. The only disadvantage here is that you cannot see the exact command in the log.

codefritzel (approved):
In my local act environment, the actions have run successfully.
Regardless of whether the bash variant or the 2 jobs are used
LGTM :+1:

saschalucas:
> So the configure task would work with a single job. The only disadvantage here is that you cannot see the exact command in the log.

Thanks for the hint, I'll keep that in mind, but for the moment I leave it as is.

PR #1776, opened 2024-07-31
redesign kvm qmp monitor

codefritzel:
This commit has changed the substructure of QMP communication.  The code has been modernised with typehints codestyle and better logging.

The inheritance hierarchy has been adapted so that communication with the QGA agent is now possible.

Furthermore, the QmpConnection class can now wait for certain events (see QmpConnection#wait_for_qmp_event).  This can also solve the issue #1749.

To minimise code changes, the old method 'Execute' refers to the new 'execute_qmp'.

Issue #1388, opened 2019-06-19
Move away from i440fx as the default machine model on x86


apoikos:
Most of our KVM hypervisor is designed around QEMU's default machine model, i440fx, which emulates a Pentium Pro class system with a [https://en.wikipedia.org/wiki/Intel_440FX](440FX northbridge) and a PIIX3 southbridge. This system has a single 32-slot PCI host bus, an ISA bridge and a couple of IDE buses. The IDE and ISA buses are of little use today; the limited space available on the PCI bus has implications on the hotplugging system, which unfortunately cannot be amended using additional PCI-to-PCI bridges (which do not support hotplugging).

Apart from i440fx, QEMU also emulates q35, an ICH9-class system with native PCIe support. PCIe is more flexible than PCI, offering the ability to add as many root ports as we like, with proper hotplugging support throughout the board. We should look into using Q35 as the base model, or build our own model starting with `-M none` and then adding devices as needed.

pboguslawski:
Tried to create VM with `-machine q35,smm=on` in Debian 11 (ganeti 3.0.2-1~deb11u1) to test [secure booting](https://wiki.debian.org/SecureBoot/VirtualMachine):

```
# gnt-instance add [...] --hypervisor-parameters kvm:machine_version='q35\,smm=on' [...] myvm.mydomain
[...]
Failure: command execution error:
Hypervisor parameter validation failed on node mynode.mydomain: Unsupported machine version: q35,smm=on

```

Tried to create VM with just `-machine q35` in the same environment:
```
# gnt-instance [...] --hypervisor-parameters kvm:machine_version='q35' [...] myvm.mydomain

# gnt-instance start myvm.mydomain
Waiting for job 123 for myvm.mydomain ...
Job 123 for myvm.mydomain has failed: Failure: command execution error:
Could not start instance 'myvm.mydomain': Hypervisor error: Failed to start instance myvm.mydomain: exited with exit code 1 ()

# cat /var/log/ganeti/kvm/myvm.mydomain.log
kvm: -device virtio-net-pci,id=nic-44444444-5555-2222,bus=pci.0,addr=0xd,netdev=nic-44444444-5555-2222,mac=12:12:12:11:11:11: Bus 'pci.0' not found
```

Would be nice to have q35 with machine properties supported in ganeti (i.e. for guest secure booting).

lbteoth:
[...]

PR #1803, opened 2024-08-30
replace time.sleep with wait_for_event

### LINKED ISSUE #1749: Implement QMP Event Handling
With #1667 we got entirely rid of using the HMP for Qemu communication. [Hot-removing disks consists of calling `device_del`, waiting for the event to confirm the removal of the device and then calling `blockdev-del`](https://github.com/ganeti/ganeti/blob/b87714e68d5413b08584d0df8d571124952ef15e/lib/hypervisor/hv_kvm/monitor.py#L584-L587).

Interacting with HMP was slow enough that it has never been noticed it was actually required to wait for the event to fire. However, with QMP the call to `blockdev-del` usually fails because the device has not been removed yet.

As a quick workaround, a `sleep(1)` has been added which mostly resembles the old timing. The real solution would be to implement some sort of "wait-for-QMP-event"-handling on the QMP implementation.

codefritzel:
This replaces the time.sleep from #1667 with wait_for_qmp_event from #1776.

closes #1749

### DIFF
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -553,10 +553,11 @@ class QmpConnection(QemuMonitorSocket):
     """
     self.execute_qmp("device_del", {"id": devid})
 
-    # TODO: implement receiving of QMP events
-    # We need to wait for the DEVICE_DELETED event via QMP before proceeding.
-    # The old implementation using HMP only worked because it is so slow.
-    time.sleep(1)
+    # wait for the DEVICE_DELETED event with five seconds timeout
+    event = self.wait_for_qmp_event("DEVICE_DELETED", 5)
+    if event is None:
+      raise errors.HypervisorError("DEVICE_DELETED event has not arrived")
+
     self.execute_qmp("blockdev-del", {"node-name": devid})
 
   def _GetPCIDevices(self):


codefritzel:
@rbott can you please run this against the QA suite?

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -553,10 +553,8 @@ def HotDelDisk(self, devid):
     """
     self.execute_qmp("device_del", {"id": devid})
 
-    # TODO: implement receiving of QMP events
-    # We need to wait for the DEVICE_DELETED event via QMP before proceeding.
-    # The old implementation using HMP only worked because it is so slow.
-    time.sleep(1)
+    # wait for the DEVICE_DELETED event with one minute timeout
```
saschalucas, line 556: I assume this should be one second not minute?

rbott: I think a minute is a good choice here. [Qemu docs](https://qemu.readthedocs.io/en/v8.1.5/interop/qemu-qmp-ref.html#qapidoc-2461) state:

> Hot removal is an operation that requires guest cooperation. This command merely requests that the guest begin the hot removal process. Completion of the device removal process is signaled with a DEVICE_DELETED event.

It does not really say what is going on inside the guest and how long it might take to complete but I would assume that a single second _could_ cause problems.

While our experience tells us that a second has been enough in the past (either because the `socat` / HMP implementation was too slow or because of the explicit `time.sleep(1)`) I would not bet on it to be accurate in _all_ cases :-)

rbott:
> @rbott can you please run this against the QA suite?

QA Run started: https://stargazer.ganeti.org/bd82b965f768d727168aaed5ac9d66a50bc52c75/

I will also send you an email and set up access to the QA system for you so that you can trigger runs by yourself :-)

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -553,10 +553,8 @@ def HotDelDisk(self, devid):
...
+    self.wait_for_qmp_event("DEVICE_DELETED", 1)
```
rbott, line 557: The docstring of wait_for_qmp_event() says:

> If the timeout is reached, None returns.

Shouldn't we check for `None` value to see if the timeout has been reached and error out accordingly (e.g. throw `HypervisorError` )? Otherwise we would just wait one minute and then continue anyways (and also I assume the subsequent QMP call to `blockdev-del` would fail).

codefritzel: You're right about the None test. I will add this.

lib/hypervisor/hv_kvm/monitor.py
codefritzel: A one-minute timeout is quite a lot for me. If something unexpected happens, the execution is blocked for a whole minute. I would suggest 5-10 seconds as timeout.

codefritzel:
@rbott what do you think about a timeout of 5 seconds?

lib/hypervisor/hv_kvm/monitor.py
rbott: I have asked on the OFTC #qemu IRC channel. Let's see if they have any advice on this. If this is about (potentially) flushing buffers/caches or whatnot, we should probably give the guest some time to deal with that.

If it is just about "has the guest OS support for hot removal or not", a short timeout should be sufficient I guess.

rbott:
@codefritzel from working on another pull request I learned that depending on the hotplug/hotremove mech used inside the guest, 15-20 seconds may be required. Especially switching to the q35 machine model with legacy PCI devices (instead of native PCIe devices with native PCIe hotplug) triggers this behaviour.

See also #1388 and the related PR for details.

Other than that, it looks good to go :-)

codefritzel:
According to qemu, a DEVICE_UNPLUG_GUEST_ERROR event can also fired if the hot unplug fails. It would be better to modify the wait_for_event method so that it can listen to several events. In this way, unnecessary waiting time can be avoided in the event of an error.

See https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#qapidoc-2336

rbott:
That sounds good, but could also be implemented in a separate PR.
I wouldn't mind merging this one as it is (with the correct timeout value) so that we can get rid of the time.sleep().

saschalucas (approved):
I think we are replacing an unconditional 1s sleep with a 5s wait for an event, right? So let's later figure out, if really longer times are needed.

So thanks @codefritzel. LGTM.

PR #1804, opened 2024-09-07
QmpConnection as pytest unittest

codefritzel:
This commits reimplement the QmpConnection as unittest for the pytest framework. Existing from
test/py/legacy/ganeti.hypervisor.hv_kvm_unittest.py has been removed.

In the unit test, a QMP socket is simulated via a thread. In this, qmp commands can be defined by creating a method according to the scheme 'simulate_<command>' in test/py/unit/hypervisor/hv_kvm/test_monitor.py. For the time being, only the most important methods such as connect, execute_qmp and wait_for_qmp_event are tested.


cilq (changes requested):
Thanks for your contribution.
I like the approach of having an explicit mapping between QMP messages and their responses.
Please implement the test cases for messages with chunked replies and multiple messages and then we can merge.

cilq (approved):
Thanks for the rework. The changes are fine now.

PR #1807, opened 2024-09-22
kvm_runtime as seperate class

codefritzel:
This commit abstracts the KVMRuntime as a separate class, which
encapsulates the structure behind it (list or dict).  Previously, the
runtime was displayed as a list or tuple. It can be assumed that this
structure is Haskell will continue to be used. For this reason, the KVM
runtime actually represents a list. In the future, this can be converted
into a dict without having to adapt the code outside KVMRuntime.

This is required for the implementation of vCPU and RAM hotplug.

Depends on https://github.com/ganeti/ganeti/pull/1804

Signed-off-by: codefritzel <pascal-pf@hotmail.de>


codefritzel:
The qa-run was successfull:
https://stargazer.ganeti.org/4ad0758bd1a631ad14430d9191b36d644fa03fda/

cilq (approved):
Thanks for improving the codebase and making it easier to understand.
I will merge this right away.

Issue #1374, opened 2019-05-27
UEFI support


paravoid:
In order to support UEFI VMs with QEMU, the following steps are needed, basically mimicking what libvirt does:
- `/usr/share/OVMF/OVMF_CODE.fd` and `/usr/share/OVMF/OVMF_VARS.fd` need to exist. In Debian, that's just `apt install ovmf`. (OVMF is the UEFI firmware for VMs)
- On every instance creation, `/usr/share/OVMF/OVMF_VARS.fd` needs to be copied $somewhere available on the nodes. This is writeable state and each VM needs its own file, and can (and usually will) be modified by the VM. An example modification would be the Debian installer adding a "debian" menu entry. This is really can be thought as a disk, and perhaps should use the disk abstractions and be DRBDed etc.?
- Ganeti should pass QEMU these arguments:
  - `-drive if=pflash,format=raw,readonly,file=/usr/share/OVMF/OVMF_CODE.fd`
  - `-drive if=pflash,format=raw,file=instancevarstore.fd`

The [KVM OVMF whitepaper](http://www.linux-kvm.org/downloads/lersek/ovmf-whitepaper-c770f8c.txt) has the details at length, but there is a pretty good summary at Debian's [#764918](https://bugs.debian.org/764918).

paravoid:
Also see this informative [mailing list post](https://lists.01.org/pipermail/edk2-devel/2018-March/022185.html) on the topic of OVMF code updates & live migrations. 

Maybe that could inform some decisions -- I think an important one would be to whether each VM should keep its own copy of the code as well (making it more stable across hosts/versions/live migrations, but also harder to update if e.g. vulnerabilities crop up), or using a shared copy.

paravoid:
One complication discovered today: Ganeti's `boot_order` argument will have no effect with OVMF. This would need to be reworked to fully support UEFI. Specifically, Ganeti would need to stop using QEMU's `-boot order=N` and instead use `-device …,bootindex=N` instead.

[RedHat's #1323085 against libvirt](https://bugzilla.redhat.com/show_bug.cgi?id=1323085#c10) has a more detailed explanation but in short:
- QEMU's `-boot order` argument sets a legacy BIOS order in CMOS (registers 0x3d and 0x38).
- OVMF does not and will not support reading those registers, as they're too simplistic ("hard drive", "CD" etc.) for what UEFI supports
[...]

PR #1808, opened 2024-09-27
switch to bootindex property

codefritzel:
This commit switches the "-boot ordner=" to the "bootindex" of from the -device parameter. The previous behavior when booting will remain unchanged.

This is preliminary work to support uefi Boot in the future. (See #1374)


codefritzel:
The qa-run was successful.
https://stargazer.ganeti.org/2bfa15105f58f2fad3771067e874fc156de75d9b/

codefritzel:
Friendly ping to @saschalucas

The qa run was deleted during this time...

saschalucas:
Thanks @codefritzel. As discussed elsewhere we tend to merge this before #1875.

This PR does less than #1572 and #1573 but switches to bootindex property without changing or adding HV parameters.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1221,21 +1217,30 @@ def _GenerateKVMBlockDevicesOptions(self, up_hvp, kvm_disks,
...
                                                  kvm_devid)
 
+      dev_val = ""
+
       if disk_type == constants.HT_DISK_IDE:
-        dev_opts.extend(["-device", "ide-hd,drive=%s,write-cache=%s" %
-                         (kvm_devid,
-                          kvm_utils.TranslateBoolToOnOff(writeback))])
+        dev_val += "ide-hd,drive={},write-cache={}".format(
+          kvm_devid, kvm_utils.TranslateBoolToOnOff(writeback)
```
saschalucas, line 1224: Can `drive={},write-cache={}` be removed and added outside the if clause?

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1221,21 +1217,30 @@ def _GenerateKVMBlockDevicesOptions(self, up_hvp, kvm_disks,
...
         # _GenerateKVMRuntime() for new instances.
 
         # Add driver, id, bus, and addr or channel, scsi-id, lun if any.
-        dev_val = _GenerateDeviceHVInfoStr(cfdev.hvinfo)
-        dev_val += ",drive=%s,write-cache=%s" % (kvm_devid,
-                                                 kvm_utils.TranslateBoolToOnOff(
-                                                   writeback))
-        dev_opts.extend(["-device", dev_val])
+        dev_val += _GenerateDeviceHVInfoStr(cfdev.hvinfo)
+        dev_val += ",drive={},write-cache={}".format(
```
saschalucas, line 1233: and `drive={},write-cache={}` is appended for both cases?

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1964,17 +1969,27 @@ def _ExecuteKVMRuntime(self, instance, kvm_runtime, kvmhelp, incoming=None):
...
             netdev = "netdev%d" % nic_seq
+
           nic_val += (",netdev=%s,mac=%s%s" % (netdev, nic.mac, nic_extra))
           tap_val = ("type=tap,id=%s,%s%s%s" %
                      (netdev, tapfd, vhostfd, tap_extra))
-          kvm_cmd.extend(["-netdev", tap_val, "-device", nic_val])
+          dev_ops.append(nic_val)
+
+          # add bootindex property to the first nic if network boot is enabled
+          if boot_network and incoming is None:
```
saschalucas, line 1988: what's the idea behind "not when `incoming`"? will then a network booted instance not be able to reboot after live migration? maybe "not when `kernel_path` was meant?

rbott:
Just a quick pointer/thing to consider: what happens during or after a cluster upgrade? Instances which have been started pre-upgrade will still have the old parameter stored in their KVM runtime (and vice versa in the rare case of a cluster downgrade). Do we need some compability or up/downgrade code here?

lib/hypervisor/hv_kvm/__init__.py
saschalucas: Caused by commit b87714e68d. Maybe resolve later ...

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1850,17 +1855,27 @@ def _ExecuteKVMRuntime(self, instance, kvm_runtime: KVMRuntime,
...
             netdev = "netdev%d" % nic_seq
+
           nic_val += (",netdev=%s,mac=%s%s" % (netdev, nic.mac, nic_extra))
           tap_val = ("type=tap,id=%s,%s%s%s" %
                      (netdev, tapfd, vhostfd, tap_extra))
-          kvm_cmd.extend(["-netdev", tap_val, "-device", nic_val])
+          dev_ops.append(nic_val)
+
+          # add bootindex property to the first nic if network boot is enabled
+          if boot_network:
```
saschalucas, line 1874: For completeness, please check if `kernel_path` is unset, else the network boot can be skipped.

saschalucas:
current observations:

Scenario | version started | version migrated to | boot=disk | boot=cd | boot=nic
-- | -- | -- | -- | -- | --
Upgrade | 3.1 | 3.2 | ✳️ migrated to bootindex | ✅ keeps -boot order=d from runtime | ✅ keeps -boot order=n from runtime
Downgrade | 3.2 | 3.1 | ✳️ migrated to -boot order=c | ✅ keeps bootindex from runtime | ❌ migrates to nothing

Legend:
* ✅ Parameter kept stable (part of runtime kvm_cmd)
* ✳️ Parameter migrated, but functional (nothing in runtime)
* ❌ Parameter and functionality lost. Instance can migrate and runs stable, but lost ability to boot from network -> must be rebooted from outside (via Ganeti) to gain back network boot. May be tolerable?

codefritzel:
Friendly ping to @saschalucas

PR #1809, opened 2024-11-04
KVM: make Ganeti compatible to the q35 machine type

rbott:
This PR picks up the conversation in #1388 (thanks @alfonso-escribano and @lbteoth) and makes some slight changes to the KVM hypervisor code to support the `q35` machine type. It does so by implementing a PCIe-to-PCI bridge which then holds all the (legacy) PCI devices (e.g. NICs).

This way we do not need to change any of the existing code (which still assumes network/disk devices reside on the PCI bus `pci.0`). In short this PR will

- add a PCIe-to-PCI bridge emulating a `pci.0` bus on `q35` machines
- modify the hotplug (verification) code to retrieve the devices located on the nested PCI bridges instead of the top-level PCI bus, otherwise Ganeti will falsely claim that hotplugging failed because it can't locate the newly added devices
- increase the timeout for adding/removing devices as the PCIe-to-PCI bridge sadly relies on the older SHPC mechanism for hotplugging/hotremoving which is slow

I have done some successful local testing with this code. There is also currently a QA run underway using `q35` machine types and one using `pc` machine types to avoid regressions. I'll update this PR once the results are in.

While this is not a perfect solution (ideally we would be using PCIe devices on the `q35` machine type with native PCIe hotplug support), this PR will at least allow users to set `machine_version` to `q35` and receive a bootable/working instance.

Since (according to the discussion in #1388) the `pc` type will be deprecated in upcoming Qemu/KVM versions and Ganeti automatically chooses the default machine type as indicated by Qemu/KVM (`kvm -machine help|grep default`) we should be prepared for a transparent switch to `q35` unless users explicitly configure the `machine_version` parameter in Ganeti.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1292,6 +1292,16 @@ class KVMHypervisor(hv_base.BaseHypervisor):
         specprop = ""
       machinespec = "%s%s" % (mversion, specprop)
       kvm_cmd.extend(["-machine", machinespec])
+      # if q35 is used, we will initialize a PCIe bus and add
+      # a pcie-to-pci bridge to provide a legacy pci.0 bus for
+      # ganeti devices
+      # this is a first step toward supporting q35 properly
+      if mversion.startswith("q35"):
+        kvm_cmd.extend(["-device",
+                        "pcie-pci-bridge,id=pcie-pci-bridge1,bus=pcie.0"])
+        kvm_cmd.extend(["-device",
+                        "pci-bridge,id=pci.0,bus=pcie-pci-bridge1,"
+                        "chassis_nr=1,addr=02"])
     else:
       kvm_cmd.extend(["-machine", mversion])
       if (hvp[constants.HV_KVM_FLAG] == constants.HT_KVM_ENABLED and
@@ -2107,7 +2117,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     @raise errors.HypervisorError: if result is not the expected one
 
     """
-    for i in range(5):
+    for i in range(15):
       found = self.qmp.HasDevice(kvm_devid)
       logging.info("Verifying hotplug command (retry %s): %s", i, found)
       if found and should_exist:
--- a/lib/hypervisor/hv_kvm/monitor.py
+++ b/lib/hypervisor/hv_kvm/monitor.py
@@ -567,6 +567,19 @@ class QmpConnection(QemuMonitorSocket):
     time.sleep(1)
     self.execute_qmp("blockdev-del", {"node-name": devid})
 
+  def _GetNestedPCIDevicesOnQ35(self, devices):
+    """Return the PCI devices (sub-) tree on nested PCI structures
+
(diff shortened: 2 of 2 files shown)


codefritzel:
Hi @rbott,
I have already started to implement pcie support.
See https://github.com/codefritzel/ganeti/tree/pcie

I had started pcie with a flat hiercharchie with a fixed number of pcie-root-ports to which the devices are (hot) plugged.
Unfortunately, as so often in this project, testing had become too complex. I thought about outsourcing the logic of which bus and address should be used to a separate class like PCIManager.

I think we should take some more time for the q35 and pcie topic and rebuild the code properly.

rbott:
Hi @codefritzel,thanks for taking on this matter. I think unbreaking q35 should be a priority here, as user currently can configure the machine type, but not boot the instance if is set to q35. I definitely agree that PCI for q35 needs a proper re-implementation, but that will take a while and should not go into 3.1 / postpone 3.1 any further.

We should inform users in the release notes that device names (e.g. predictable network interface names) might change in the future (due to bus topology changes) on q35. But at least it will be usable, which it is currently not.

rbott:
Hi @codefritzel, we are trying to get a 3.1 beta release out (to target a 3.1-stable for Debian Trixie) and there won't be enough time to fully re-implement the PCI stuff. I would propose to move a full re-implementation of the PCI handling to 3.2, so that we have enough time to do extensive testing and have proper documentation of the expected changes.

E.g. currently additional disks "move" VirtIO net devices further back on the bus, leading to problems with "predictable network interface names" which could be considered a bug, but changing _anything_ about this behaviour needs to be thoroughly tested. Either we need to fully mimic the current behaviour in all aspects (which is hard, considering the amount of knobs and settings) or we end up shuffling around devices for users between reboots/failovers (or worse, during live migrations).

With that in mind I would still vote to "unblock" q35 usage for Ganeti 3.1 with this simple fix, but add to the release notes that 
- qemu will change the default model and people _should_ set the model inside Ganeti to `pc` to avoid any surprises (unless they specifically _want_ to use/test `q35`)
- the current `q35` support is experimental at best and the PCI structure is bound to change with 3.2

After that, we can discuss/implement/test/document a new PCI handling which covers both `pc` and `q35` properly, uses proper PCIe hotplugging and either fixes current Ganeti PCI hotplugging glitches _or_ specifically re-implements them.

codefritzel:
Hi @rbott,
I'm struggling with this “small quick” code change. Here again something is inserted that can only be tested live on the cluster without unit test scenarios.  A function is extended with a new function for a single use case.

I think it would make more sense to make the HasDevice function recursive and to create unit tests that cover this case and the other cases. In this way, only one function is processed and one test is written for it.

rbott:
After a useful discussion on IRC I will ditch this PR for now and instead open another PR which adds some upgrade code to switch machine_type from "unconfigured" to "pc". Configurations which already have set. Specific value will be left untouched as those users probably know what they are doing.

That way, there will be no surprising changes during qemu upgrades.

We can come up with a better solution to the actual problem in time for the 3.2 release.

PR #1810, opened 2024-12-05
speedup CI cabal build

saschalucas:
... by utilizing all available CPUs. Unlike `make -j`, which runs many things in parallel, this still results in only one invocation of GHC compiler. Parallelism seems internal. Increase in memory usage on 8 vCPU VM stays within acceptable limit of less than 2 GiB:

without `-j`:
```
    297.38user 15.19system 5:11.03elapsed 100%CPU (0avgtext+0avgdata 1082668maxresident)k
    0inputs+2338560outputs (4367major+3696086minor)pagefaults 0swaps
```
with `-j`:
```
    530.65user 177.55system 3:19.71elapsed 354%CPU (0avgtext+0avgdata 1410948maxresident)k
    0inputs+2286024outputs (4453major+3776267minor)pagefaults 0swaps
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,8 +41,11 @@ jobs:
         if: ${{ matrix.regex == 'tdfa' }}
         run: ./configure --enable-haskell-tests --enable-regex-tdfa
 
+      - name: pimp cabal parallelism
+        run: sed -e "s/\t\$(CABAL_SETUP) build/\t\$(CABAL_SETUP) build -j$(nproc)/" -i Makefile.am
+
       - name: Build
-        run: make -j 2
+        run: make -j $(nproc)
 
       - name: Check Local
         run: LC_ALL=C make check-local
--- a/Makefile.am
+++ b/Makefile.am
@@ -1485,7 +1485,7 @@ HS_SRCS = $(HS_LIBTESTBUILT_SRCS)
 $(HS_SRC_PROGS) exe/htest: dist/app.stamp
 
 dist/app.stamp: exe/ dist/setup-config
-	$(CABAL_SETUP) build -j
+	$(CABAL_SETUP) build
 	-(cd exe; \
 	 for name in ganeti-kvmd ganeti-wconfd ganeti-confd ganeti-luxid \
 		rpc-test ganeti-mond ganeti-metad \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1485,7 +1485,7 @@ HS_SRCS = $(HS_LIBTESTBUILT_SRCS)
 $(HS_SRC_PROGS) exe/htest: dist/app.stamp
 
 dist/app.stamp: exe/ dist/setup-config
-	$(CABAL_SETUP) build
+	$(CABAL_SETUP) build -j
 	-(cd exe; \
 	 for name in ganeti-kvmd ganeti-wconfd ganeti-confd ganeti-luxid \
 		rpc-test ganeti-mond ganeti-metad \


thielema:
On Wed, 4 Dec 2024, saschalucas wrote:

> ... by utilizing all available CPUs. Unlike make -j, which runs many 
> things in parallel, this still results in only one invocation of GHC 
> compiler. Parallelism seems internal. Increase in memory usage on 8 vCPU 
> VM stays within acceptable limit of less than 2 GiB:
> 
> without -j:
>
>     297.38user 15.19system 5:11.03elapsed 100%CPU (0avgtext+0avgdata 1082668maxresident)k
>     0inputs+2338560outputs (4367major+3696086minor)pagefaults 0swaps
> 
> with -j:
>
>     530.65user 177.55system 3:19.71elapsed 354%CPU (0avgtext+0avgdata 1410948maxresident)k
>     0inputs+2286024outputs (4453major+3776267minor)pagefaults 0swaps


There are two levels of parallelization, one provided by Cabal and one by 
GHC. "cabal -j3" means: "compile up to 3 packages in parallel". But GHC 
itself can also parallelize. That is, "cabal --ghc-option=-j3" means 
"compile up to three modules within package in parallel".

It would be nice to have only one "-j" option that sets a maximum of 
parallelization across all modules of all packages, but this would break 
the separation between Cabal and GHC. This feature is not impossible but 
much more difficult to implement, because it needs to run GHC in a server 
mode. It seems the problem was solved recently, but I have not used the 
feature myself so far:

https://well-typed.com/blog/2023/08/reducing-haskell-parallel-build-times/


To stay compatible with older Cabal and GHC versions I'd suggest to use 
"cabal -jN --ghc-option=-jM" where N*M slightly exceeds the number of 
available CPU cores and test which combination works best.

Of course, using the maximum number of CPU cores does not always maximize 
throughput because for a large number of modules available RAM can become 
the limitting factor.

However, in Debian the imported GHC packages are compiled separately by 
Debian anyway. Thus, only --ghc-option=-jN has an effect.

thielema:
Btw, I would never hardcode parallelization in build scripts because the user may not want to employ all CPU cores with building. Maybe he wants to build in background with only half of the available cores, or even only one core in order to save memory.
You can configure default parallelization in `$HOME/.cabal/config` like so:
~~~~
jobs: $ncpus
~~~~
You can also put this in a project directory in `$PROJECT_DIR/cabal.config`.
However, `jobs` only applies to Cabal, thus it won't have an effect for single package builds as in Debian.

For parallel module builds, you may put
~~~~
package ganeti
  ghc-option: -j
~~~~
in `$PROJECT_DIR/cabal.project` or run cabal with option `--ghc-option=-j`.

saschalucas:
> Btw, I would never hardcode parallelization in build scripts because the user may not want to employ all CPU cores with building. Maybe he wants to build in background with only half of the available cores, or even only one core in order to save memory. 

This is true, but I've suppressed that in favor to get CI-test faster. Ideally I would like to pick up the users wanted parallelism supplied at `make -jN`, but I didn't find a way to query this inside the `Makefile`???

What would happen, if we ship a `cabal.config` in Ganeti? Will this enforce the user to use the dictated parallelism?

Or are there other ideas to add this to https://github.com/ganeti/ganeti/blob/377cfd5840476ee72e91a60b2c53a42e8b7a1546/.github/workflows/ci.yml#L44-L45 Maybe inline a `cabal.config` generated by an `echo TEXT > cabal.config`?

thielema:
On Fri, 6 Dec 2024, thielema wrote:

> Indeed, putting cabal.config in the repository would have the same 
> effect as hardcoding parallelism in the build script. As you propose, I 
> would put a echo TEXT > cabal.config in the CI script.

I have just checked: cabal.config does not accept ghc-option or 
ghc-options field.

Instead you must include the lines:

program-default-options
   ghc-options: -j8


When you run 'cabal init' in a fresh (project) directory as a fresh user, 
cabal will generate a $HOME/.cabal/config containing all options and their 
default values as comments.



With 'cabal v2-build' (Nix-style builds) we could initialize the file 
cabal.project.local with this content:

program-options
   ghc-options: -j8

But for Debian compatibility we use plain Cabal (the library) builds.

rbott:
Sorry for the delay. This looks good to me. We should merge it after the `3.1.0-rc1` release, thanks @saschalucas and @thielema!

rbott (approved):

PR #1812, opened 2025-01-15
qemu-9.0 compat: user and chroot are now combined

saschalucas:
With qemu>=9.0 the two independent parameters `-runas` and `-chroot` are merged into the new `-run-with` parameter. In preparation move related parts of the code together.

Introduce a separate generator function for `-run-with`, which combines user and chroot parameters. Call this in every combination of chroot and security_model.


rbott (approved):
LGTM, please merge

PR #1813, opened 2025-01-17
Fix multiple incompabilities with Qemu 7 and 9

rbott:
This PR addresses multiple issues which have been uncovered by running the QA suite against Qemu 9.2 from Debian Testing:

- Qemu >= 9.0 requires O_DIRECT set on storage backends/fd's
- typing in QMP messages has become more strict ("on/off" settings need to be a literal string, previously a boolean was also accepted)
- Qemu >= 9.0 has removed `-no-acpi` and replaced it with a `-machine` parameter (that also uncovered some other, older bugs)
- Qemu >= 7.1 has removed `-soundhw` and is now replaced by `-audiodev`

I'll keep this as draft for now, as the QA suite might uncover more problems. The `soundhw` fix is also still WIP.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -502,6 +502,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     re.compile(r"^-blockdev\s([^-]|(?<!^)-)*,auto-read-only=on\|off",
                re.M | re.S)
   _RUNWITH_RE = re.compile(r"^-run-with\s.*chroot=.*user=", re.M)
+  _ACPI_RE = re.compile(r"^-no-acpi\s", re.M)
 
   # Slot 0 for Host bridge, Slot 1 for ISA bridge, Slot 2 for VGA controller
   # and the rest up to slot 11 will be used by QEMU implicitly.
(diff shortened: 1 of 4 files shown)


lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1255,7 +1257,12 @@ def _GenerateKVMRuntime(self, instance, block_devices, startup_paused,
     # As requested by music lovers
     if hvp[constants.HV_SOUNDHW]:
       soundhw = hvp[constants.HV_SOUNDHW]
-      kvm_cmd.extend(["-soundhw", soundhw])
+      if self._SOUND_RE.search(kvmhelp):
+        # this parameter has been replaced starting with Qemu 7.1
+        kvm_cmd.extend(["-soundhw", soundhw])
+      else:
+        kvm_cmd.extend(["-audio", "driver=none,model=%s,id=soundhw"
+                        % soundhw])
```
saschalucas, line 1265: It seems that `driver=none` does not create a related device to bring the sound outside of the instance. at least in a test with spice no sound is played. I've spied on libvirt, they use something like
```
+      kvm_cmd.extend(["-device", "ich9-intel-hda,id=sound0",
+                      "-device", "hda-duplex,id=sound0-codec0,bus=sound0.0"]
+      )
```
This magically seems to try alsa, when no spice is configured and spice, when spice is used.

I can't think of other cases than spice, so an option may be to conditionally only enable sound, when also `spice_bind` is set.

An other option would be to turn `HV_SOUNDHW` into a boolean that only enables this one ich9/hda-duplex. I have somehow tried this way in commit c2a8e28cd09d5afc34c3e35d2052678d42f53803, which works fine, expect in the case of new cluster init...

rbott: Yes, it basically creates a sound card without any audio backend on the host (which only seems useful if you run a software that fails without a soundcard being present but don't care about the output at all).

I actually did not know that playing audio via spice would be an option. I'll take this into account and try to rewrite that part.

man/gnt-instance.rst
```diff
@@ -700,9 +700,10 @@ security\_domain
 kvm\_flag
     Valid for the KVM hypervisor.
 
-    If *enabled* the -enable-kvm flag is passed to kvm. If *disabled*
-    -disable-kvm is passed. If unset no flag is passed, and the
-    default running mode for your kvm binary will be used.
+    If *enabled* accel=kvm is appended to the -machine parameter
+    *Disabling* (or not setting this flag at all) currently does
+    nothing. Since KVM is enabled by default anways, this flag is
+    only kept for backwards compability of Ganeti.
```
saschalucas, line 706: I'm not sure about enabled by default. Maybe this is only the case when the `/usr/bin/kvm` executable is used. Setting `kvm_path` to `qemu-system-x86_64` could result in [tcg](https://manpages.debian.org/jessie/qemu-system-x86/qemu-system-x86_64.1.en.html#OPTIONS).

rbott: That is true, I'll give that another spin.

At least we can say for sure there is no `-disable-kvm` parameter (and according to my research never has been). But since the code always checked the KVM help output for the existand of said parameter first, it never caused any errors. But also no notifications that disabling KVM did not succeed.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1529,6 +1521,23 @@ def _GenerateKVMRuntime(self, instance, block_devices, startup_paused,
...
+        # also, we need to pick a host backend/driver. we'll use
+        # spice if that is configured, otherwise there will be
+        # no sound output
+        if spice_bind:
+          driver = "spice"
+        else:
+          driver = "none"
+        kvm_cmd.extend(["-audio", "driver=%s,model=%s,id=soundhw"
+                        % (driver, soundhw)])
+
```
codefritzel, line 1540: If you are already editing something here, please use format strings right away:
```suggestion
        kvm_cmd.extend(["-audio", "driver={},model={},id=soundhw"
                               .format(driver, soundhw)])
```

rbott:
So technically, this PR gets the QA suite into a state where it passes on bookworm (~6 hours). On trixie however it has become painfully slow. My last attempt was: 19 hours of (successful) QA, after that it failed because the disk of the master node was full due to extensive logging. I added another 5GB of diskspace to the nodes and it is currently running since last night.

I also started some work to convert the QA output into something that https://www.speedscope.app/ understands to graph the flow/timing of the QA. This might give some insight on where things go wrong. However, this is not done yet (and maybe speedscope is not the right tool for the job). I will update this PR as soon as I have figured out what slows down Ganeti. Since this is the first time we have been getting this far with the QA suite on trixie, there is really nothing to compare against (besides Bookworm results). Previously, all the Haskell and Qemu changes prevented the QA from running at all or at least from running more than 30 minutes on Trixie.

man/gnt-instance.rst
rbott: I have updated the documentation to reflect the current state.

lib/hypervisor/hv_kvm/__init__.py
rbott: done

lib/hypervisor/hv_kvm/__init__.py
rbott: I have updated to code to use spice for the host side if it is configured, otherwise fall back to no audio output.

codefritzel (approved):
The code look good to me.

saschalucas (changes requested):
Thanks @rbott. Just a minor change request regarding sound and the man page gnt-instance(8). Would it be possible to document how the user can find information in the qemu man page regarding the sound card model? Since `soundhw` is not present in newer qemu man pages, a hint to `-audio` or `-audio model=help` or typical models `ac97` / `hda` would be helpful.

rbott:
Thanks for the hint @saschalucas, I have tried to adress this in the man page in the latest update.

saschalucas (approved):
Thanks @rbott. Commit 66d9b56220fc45c260538d21c56552c160c836f2 introduces leftovers from a merge: `<<<<<<< HEAD` and `=======`, which the next commit removes. Regarding this, it may be cleaner to squash this two. I'm also fine to squash all 4 commits, but please deiced on your own, whether to squash at all.

LGTM.

rbott:
I have fixed the commit (through rebasing), good catch @saschalucas! I will now rebase and merge, thanks everyone :-)

PR #1814, opened 2025-01-24
Classify DRBD devices in StandAlone mode as faulty

### LINKED ISSUE #1218: gnt-cluster doesn't detect "unknown" disks as drbd failed
Originally reported of Google Code with ID 1165.

```
What software version are you running? Please provide the output of "gnt-
cluster --version", "gnt-cluster version", and "hspace --version".

Software version: 2.12.4
Internode protocol: 2120000
Configuration format: 2120000
OS api version: 20
Export interface: 0
VCS version: (ganeti) version v2.12.4

gnt-cluster (ganeti v2.12.4) 2.12.4

hspace (ganeti) version v2.12.4
compiled with ghc 7.6
running on linux x86_64


<b>What distribution are you using?</b>

debian jessie

<b>What steps will reproduce the problem?</b>
1. server1 with instance a primary, server2 with instance a secondary
2. reboot server2
3. after reboot sometimes on server2 - drbd state for disk of instance a is UpToDate/UpToDate, but on server1 state is:


 9: cs:WFConnection ro:Primary/Unknown ds:UpToDate/DUnknown C r---b-

4. gnt-cluster verify-disk doesn't detect that as a problem, and doesn't fix that. gnt-instance activate-disk a - fix problem correctly.

<b>What is the expected output? What do you see instead?</b>

gnt-cluster activate-disk should detect that problem and fix it.

<b>Please provide any additional information below.</b>

I think problem is that unknown state is not recognized as problem. in:
lib/backend.py problematic are states: stats.is_standalone or stats.is_diskless.

I think there should be added additional state stats.is_unknown which should point problem when state is Unknown
```
Originally added on 2016-01-11 10:46:42 +0000 UTC.

rbott:
DRBD devices should be reported faulty not only when in diskless mode, but also in StandAlone mode (as local data will most likely be outdated). This fixes the issue that degraded connections will be re-classified from ERROR to WARNING as long as they are not diskless on any of the nodes.

Broken/disconnected DRBD devices will now generate a proper ERROR and corresponding return code from `gnt-cluster verify`.

Fixes #1218

This is currently running through QA:
- [Bookworm](https://stargazer.ganeti.org/c0bd4ec49ed7c91aad1e05ec2682ff4fad2ae3e0/)
- [Trixie](https://stargazer.ganeti.org/2c99e54004702beccfde84e675decfb585b6a46b/)

*Update*: both finished successfully

### DIFF
--- a/lib/storage/drbd.py
+++ b/lib/storage/drbd.py
@@ -674,10 +674,10 @@ class DRBD8Dev(base.BlockDev):
     stats = self.GetProcStatus()
     is_degraded = not stats.is_connected or not stats.is_disk_uptodate
 
-    if stats.is_disk_uptodate:
-      ldisk_status = constants.LDS_OKAY
-    elif stats.is_diskless:
+    if stats.is_diskless or stats.is_standalone:
       ldisk_status = constants.LDS_FAULTY
+    elif stats.is_disk_uptodate:
+      ldisk_status = constants.LDS_OKAY
     elif stats.is_in_resync:
       ldisk_status = constants.LDS_SYNC
     else:


saschalucas (approved):

rbott:
Thanks for your quick review @saschalucas!

Issue #1820, opened 2025-01-31
downgrade 3.1 -> 3.0: instances startet by 3.1 are ERROR_down in 3.0


saschalucas:
Instances started/rebooted with Ganeti-3.1 are in `ERROR_down` when downgrade to 3.0:

```
root@master:~# gnt-cluster upgrade --to 3.0
...
Verifying cluster.
Command ['gnt-cluster', 'verify'] failed: exited with exit code 1; Output Submitted jobs 72, 73
...
Waiting for job 73 ...
Fri Jan 31 16:44:20 2025 * Verifying group 'default'
Fri Jan 31 16:44:20 2025 * Gathering data (3 nodes)
Fri Jan 31 16:44:20 2025 * Gathering information about nodes (3 nodes)
Fri Jan 31 16:44:21 2025 * Gathering disk information (3 nodes)
Fri Jan 31 16:44:21 2025 * Verifying configuration file consistency
Fri Jan 31 16:44:22 2025 * Verifying node status
Fri Jan 31 16:44:22 2025 * Verifying instance status
Fri Jan 31 16:44:22 2025   - ERROR: instance test.vm: instance not running on its primary node node03.344c50b6b6ce146871ed935b6928df518cb6da6a.gnt.test
Fri Jan 31 16:44:22 2025 * Verifying orphan volumes
Fri Jan 31 16:44:22 2025 * Verifying N+1 Memory redundancy
Fri Jan 31 16:44:22 2025 * Other Notes
Fri Jan 31 16:44:22 2025 * Hooks Results
```
```
root@master:~# gnt-instance list
Instance Hypervisor OS                                 Primary_node                                             Status     Memory
test.vm  kvm        instance-guestfish+debian-bookworm node03.344c50b6b6ce146871ed935b6928df518cb6da6a.gnt.test ERROR_down      -
```

Luckily double running by ganeti-watcher is circumvented by qemu:
```
root@node03:~# cat /var/log/ganeti/kvm/test.vm.log
kvm: cannot create PID file: Cannot lock pid file: Resource temporarily unavailable  
```

Can be related to 459fdb7415832caa3926c8ddd2e4cbafe49b399b?

rbott:
> Can be related to [459fdb7](https://github.com/ganeti/ganeti/commit/459fdb7415832caa3926c8ddd2e4cbafe49b399b)?

That commit has already been in all 3.0.x releases. I wonder how that could be an issue now. But I have not yet taken a look into the code.

[...]

PR #1821, opened 2025-02-03
Up-/Downgrade Code for 3.0 <-> 3.1

rbott:
This PR addresses the following things:
- add missing downgrade code for the `debug_threads`  KVM hvparam
- set `machine_version` KVM hvparam explicitly to `pc` if hvparam was unset prior to upgrade (we assume that users who have set that param to something specific know what they are doing)
- initialize new clusters with `machine_version` set to `pc`

### DIFF
--- a/lib/tools/cfgupgrade.py
+++ b/lib/tools/cfgupgrade.py
@@ -762,11 +762,31 @@ class CfgUpgrade(object):
       if variant in hvparams:
         hvparams[variant]["xen_cmd"] = "xl"
 
+  @OrFail("Revert KVM parameter")
+  def DowngradeKVMSettings(self):
+    """Downgrades KVM Settings
+
+    """
+    # pylint: disable=E1103
+    # Because config_data is a dictionary which has the get method.
+    cluster = self.config_data.get("cluster", None)
+    if cluster is None:
+      raise Error("Can't find the cluster entry in the configuration")
+
+    hvparams = cluster.get("hvparams", None)
+    if hvparams is None:
+      return
+
+    if DOWNGRADE_MAJOR == 3 and DOWNGRADE_MINOR < 1:
+      if "debug_threads" in hvparams[constants.HT_KVM]:
+        del hvparams[constants.HT_KVM]["debug_threads"]
+
   def DowngradeAll(self):
     self.config_data["version"] = version.BuildVersion(DOWNGRADE_MAJOR,
                                                        DOWNGRADE_MINOR, 0)
 
     self.DowngradeXenSettings()
+    self.DowngradeKVMSettings()
     return not self.errors
 
   def _ComposePaths(self):
--- a/lib/tools/cfgupgrade.py
+++ b/lib/tools/cfgupgrade.py
@@ -365,6 +365,15 @@ class CfgUpgrade(object):
         cluster["hvparams"][constants.HT_KVM][constants.HV_DISK_DISCARD] = \
           constants.HT_DISCARD_IGNORE
 
+      # if machine_version is unset, it will be set to 'pc' with 3.1.0
+      # this way we avoid the soon-to-be QEMU default of 'q35' which is
+      # currently broken in Ganeti
+      if cluster["hvparams"][constants.HT_KVM]\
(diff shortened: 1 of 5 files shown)


lib/tools/cfgupgrade.py
```diff
@@ -365,6 +365,15 @@ def UpgradeCluster(self):
...
 
+      # if machine_version is unset, it will be set to 'pc' with 3.1.0
+      # this way we avoid the soon-to-be QEMU default of 'q35' which is
+      # currently broken in Ganeti
+      if cluster["hvparams"][constants.HT_KVM]\
+              [constants.HV_KVM_MACHINE_VERSION] == "":
+        cluster["hvparams"][constants.HT_KVM]\
+          [constants.HV_KVM_MACHINE_VERSION] = \
+          constants.HT_KVM_MACHINE_VERSION_PC
+
```
saschalucas, line 376: I'm fine with default to `pc`on new cluster init (in src/Ganeti/Constants.hs). But I need to think a moment on changing this default on upgrade. At least it can not be reversed on downgrade :-).

rbott: Sure. An alternative way could also be to just print a warning during cluster upgrade informing users that they _might_ run into future problems, because machine_version is currently unset. That would leave it up to them to fix this.

lib/tools/cfgupgrade.py
```diff
@@ -762,11 +771,31 @@ def DowngradeXenSettings(self):
...
+      raise Error("Can't find the cluster entry in the configuration")
+
+    hvparams = cluster.get("hvparams", None)
+    if hvparams is None:
+      return
+
+    if DOWNGRADE_MAJOR == 3 and DOWNGRADE_MINOR < 1:
+      if "debug_threads" in hvparams[constants.HT_KVM]:
+        del hvparams[constants.HT_KVM]["debug_threads"]
+
```
saschalucas, line 792: `debug_threads` can also be set individually at instance level, right? Shouldn't the downgrade code also check and remove on every instance?

rbott: Yes you are right, I guess it was a bit to late :-) I'll test for this scenario and add code if required.

lib/tools/cfgupgrade.py
saschalucas: > But I need to think a moment on changing this default on upgrade.

Here is my first obstacle: Ubuntu decided to have their own (`default`) machine versions:
```
# qemu-system-x86_64 -machine help
Supported machines are:
...
ubuntu               Ubuntu 24.04 PC (i440FX + PIIX, 1996) (alias of pc-i440fx-noble)
pc-i440fx-noble      Ubuntu 24.04 PC (i440FX + PIIX, 1996) (default)
...
pc                   Standard PC (i440FX + PIIX, 1996) (alias of pc-i440fx-8.2)
```
While Debian does not, where `default` matches what `pc` stands for:
```
$ qemu-system-x86_64 -machine help
Supported machines are:
...
pc                   Standard PC (i440FX + PIIX, 1996) (alias of pc-i440fx-7.2)
pc-i440fx-7.2        Standard PC (i440FX + PIIX, 1996) (default)
```
Defaulting now to `pc`on Ubuntu will switch from `default=pc-i440fx-noble` to `pc=pc-i440fx-8.2` which is not the same (at least by comparing the name, but I don't know if there are technically differences).

saschalucas: I have following suggestions:
* don't touch `machine_version` on upgrade
* warn and point to man page on empty `machine_version`in `gnt-cluster verify`, this is even the last step of `gnt-cluster upgrade`, so every user should see it
* add all thoughts from this PR regarding `machine_version` into the man page `gnt-instance(8)`

rbott:
I will close this PR. Discussion (#1820 and this PR) has shown that `debug_threads` should not be present in 3.1 and instead aim for 3.2.

I'll follow up with a new PR to clean up this mess (remove debug_threads, keep upgrade/downgrade tests, take care of `machine_version` with documentation) :-)

PR #1528, opened 2020-10-04
Add KVM parameter to toggle QEMU's debug-threads=on|off feature

rbott:
This PR adds support for the `debug-threads` sub-parameter of the `qemu -name` parameter, according to #1526. This sub-parameter has been introduced in QEMU 2.0 and hence should not impose a problem on our currently supported platforms.

Albeit containing the phrase "debug", this change does not imply performance penalties. It will make QEMU name its threads in a more meaningful way, making it easier to debug. You can observe this e.g. with `top` (using the `H` key to view threads instead of tasks).

I have labeled this change for Ganeti 3.1, as it introduces new KVM instance parameters. It defaults to "false/off", which should reflect the current behaviour.

PR #1822, opened 2025-02-06
Postpone `debug_threads` to 3.2 release, add hvparam assessment & up/downgrade tests

### LINKED ISSUE #1818: downgrade 3.1 -> 3.0 Unknown parameter `debug_threads`
Commit 0acdcecdeb318f15dfcccc4a0610f20107ce541a introduced a new KVM parameter `debug_threads`, but misses downgrade code to remove this:

A downgrade to 3.0 from a 3.1 cluster:
```
# gnt-cluster upgrade --to 3.0
Verifying 3.0 present on all nodes
Draining queue
Pausing the watcher for one hour.
Stopping daemons on master node.
Stopping daemons everywhere.
Backing up configuration as /var/lib/ganeti1738331384.tar
Downgrading configuration
Performing version-specific downgrade tasks.
Switching to version 3.0 on all nodes
Starting daemons everywhere.
Redistributing the configuration.
Restarting daemons everywhere.
Undraining the queue.
Running post-upgrade hooks
Unpausing the watcher.
Verifying cluster.
Command ['gnt-cluster', 'verify'] failed: exited with exit code 1; Output Submitted jobs 23, 24
Waiting for job 23 ...
Fri Jan 31 14:50:04 2025 * Verifying cluster config
Fri Jan 31 14:50:04 2025 * Verifying cluster certificate files
Fri Jan 31 14:50:04 2025 * Verifying hypervisor parameters
Fri Jan 31 14:50:04 2025   - ERROR: cluster : hypervisor cluster parameters syntax check (source kvm): Unknown parameter 'debug_threads'
```
A simple solution would be to not introduce a new HV parameter and implicitly set `debug-threads=on`on the command line (like libvirt does). Otherwise we need some downgrade code here.

### LINKED ISSUE #1819: machine_version should be set explicetly during 3.0 -> 3.1 upgrade
Currently if unset `machine_version` will be detected from Qemu's `-M ?` output. As of now, `pc` is the default, but that might change in the future (also see also comments in #1388). The other sensible option `q35` is currently not usable with Ganeti (at least not without some hacks using `kvm_extra`). 

To spare Ganeti users a possibly bad experiene, we should add upgrade code from 3.0 to 3.1 to explicetly set `machine_version` to `pc` _if_ it is currently unset. If users have configured to anything specific, they probably have a good reason to do so and `machine_version` should not be touched.

`pc` should then also become the default of `machine_version` on new clusters, until `q35` has been implemented properly.

### LINKED ISSUE #1820: downgrade 3.1 -> 3.0: instances startet by 3.1 are ERROR_down in 3.0

rbott:
In #1820 we learned that the implementation of `debug_threads` (#1528) causes more problems than it solves. To not further postpone the 3.1 release, this PR does the following:

- revert #1528 
- prepare Ganeti to deal with optional comma separated parameters, which are appended to the instance name (`kvm -name <instance-name>,<param1=value>,<param2=value>`). This way, re-introducing `debug_threads` (or any other parameter that gets attached to the instance name) with Ganeti 3.2 will not break Ganeti 3.1 after a cluster-downgrade (chances of users downgrading clusters are rare, but they should expect them to not break havoc :-) )
- to address #1819, we re-introduce #1607 in a slightly updated manner, to warn users about the implications of an unset/empty `machine_verison` hvparam
- this also integrates missing cluster up/downgrade unit tests as present in #1821 (but without the up/downgrade code related to `machine_version` and `debug_threads`)
- new clusters will now default to `pc` as `machine_version` because `q35` is not supported by Ganeti (yet)

Fixes #1818 
Fixes #1819 
Fixes #1820

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -408,7 +408,6 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     constants.HV_KVM_SPICE_USE_TLS: hv_base.NO_CHECK,
     constants.HV_KVM_SPICE_TLS_CIPHERS: hv_base.NO_CHECK,
     constants.HV_KVM_SPICE_USE_VDAGENT: hv_base.NO_CHECK,
-    constants.HV_KVM_DEBUG_THREADS: hv_base.NO_CHECK,
     constants.HV_KVM_FLOPPY_IMAGE_PATH: hv_base.OPT_FILE_CHECK,
     constants.HV_CDROM_IMAGE_PATH: hv_base.OPT_FILE_OR_URL_CHECK,
     constants.HV_KVM_CDROM2_IMAGE_PATH: hv_base.OPT_FILE_OR_URL_CHECK,
@@ -611,7 +610,7 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     while arg_list:
       arg = arg_list.pop(0)
       if arg == "-name":
-        instance = arg_list.pop(0).split(",")[0]
+        instance = arg_list.pop(0)
       elif arg == "-m":
         memory = int(arg_list.pop(0))
       elif arg == "-smp":
@@ -1233,11 +1232,8 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     pidfile = self._InstancePidFile(instance.name)
     kvm = hvp[constants.HV_KVM_PATH]
     kvm_cmd = [kvm]
-    if hvp[constants.HV_KVM_DEBUG_THREADS]:
-      name_parameter = "%s,debug-threads=on" % (instance.name)
-    else:
-      name_parameter = "%s,debug-threads=off" % (instance.name)
-    kvm_cmd.extend(["-name", name_parameter])
+    # used just by the vnc server, if enabled
+    kvm_cmd.extend(["-name", instance.name])
     kvm_cmd.extend(["-m", instance.beparams[constants.BE_MAXMEM]])
 
     smp_list = ["%s" % instance.beparams[constants.BE_VCPUS]]
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -610,7 +610,11 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     while arg_list:
       arg = arg_list.pop(0)
       if arg == "-name":
-        instance = arg_list.pop(0)
+        # Qemu supports additional parameters which are appended
+        # comma-separated to the -name parameter. While Ganeti
+        # does not use this _now_, we need to be aware of future
+        # Ganeti versions which might (see #1820 for the whole story)
+        instance = arg_list.pop(0).split(",")[0]
       elif arg == "-m":
         memory = int(arg_list.pop(0))
       elif arg == "-smp":
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -2714,6 +2714,87 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     validate_machine_version(hvparams, kvm_output)
 
   @classmethod
+  def AssessParameters(cls, hvparams):
+    """Check the given parameters for uncommon/suboptimal values
+
+    This should check the passed set of parameters for suboptimal
+    values.
+
+    @type hvparams: dict
+    @param hvparams: dictionary with parameter names/value
+
+    """
+
+    warnings = []
+
+    cpu_type = hvparams[constants.HV_CPU_TYPE]
+    if not cpu_type:
+      warnings.append("cpu_type is currently unset and defaults to 'qemu64'"
+                      ", please read the gnt-instance man page on the security"
+                      " and performance implications of this parameter.")
+    elif cpu_type in ("qemu32", "qemu64", "kvm32", "kvm64"):
+      warnings.append("cpu_type is currently set to '%s'"
+                      ", please read the gnt-instance man page on the security"
+                      " and performance implications of this parameter." %
+                      cpu_type)
+    elif cpu_type == "host":
+      warnings.append("cpu_type is currently set to 'host', please make"
+                      " sure all your cluster nodes have the exact same"
+                      " CPU type to allow live migrations.")
+
+    vhost_net = hvparams[constants.HV_VHOST_NET]
+    if not vhost_net:
+      warnings.append("vhost_net should be enabled for paravirtual NICs to"
+                      " improve network latency and throughput.")
+
+    migration_bandwidth = hvparams[constants.HV_MIGRATION_BANDWIDTH]
+    default_migration_bw = constants.HVC_DEFAULTS[constants.HT_KVM]\
+      [constants.HV_MIGRATION_BANDWIDTH]
+    if migration_bandwidth == default_migration_bw:
+      warnings.append("migration_bandwidth is still set to its default value"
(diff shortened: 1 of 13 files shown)


rbott:
QA against Debian Trixie is currently running: https://stargazer.ganeti.org/df443c9b38d4f3ae12a7e95c7fc388885ba30e25/

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2713,6 +2713,89 @@ def ValidateParameters(cls, hvparams):
...
+    elif cpu_type == "host":
+      warnings.append("cpu_type is currently set to 'host', please make"
+                      " sure all your cluster nodes have the exact same"
+                      " CPU type to allow live migrations.")
+
+    disk_aio = hvparams[constants.HV_KVM_DISK_AIO]
+    if disk_aio == "threads":
+      warnings.append("disk_aio is set to 'threads' and should be set"
+                      " to 'native' for better disk performance.")
+
```
saschalucas, line 2749: On which basis is `aio=native` suggested here? Other sources[1] seem to recommend `=threads`? Maybe we can avoid suggesting non critical (and sometimes not easy to explain) parts?

[1] https://kb.blockbridge.com/technote/proxmox-aio-vs-iouring/

apoikos: @saschalucas I think you are (understandably) mixing up `disk_aio=threads` and IOThreads which is the backend implementation at the controller level. See http://kvmonz.blogspot.com/p/knowledge-disk-performance-hints-tips.html (esp. "Choosing the right AIO mode") for details, but the summary is `disk_aio` should be set to `native` and the IOThreads mentioned by Proxmox should be configured at the disk controller level.

lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -2713,6 +2713,89 @@ def ValidateParameters(cls, hvparams):
...
+    vhost_net = hvparams[constants.HV_VHOST_NET]
+    if not vhost_net:
+      warnings.append("vhost_net should be enabled for paravirtual NICs to"
+                      " improve network latency and throughput.")
+
+    migration_bandwidth = hvparams[constants.HV_MIGRATION_BANDWIDTH]
+    if migration_bandwidth < 125:
+      warnings.append("migration_bandwidth is set to less than 125 MiB/s"
+                      ", please consider adjusting it to a higher value"
+                      " that can leverage todays network speeds.")
```
saschalucas, line 2759: Would it be possible to compare against cluster-init-defaults here (`constants.HVC_DEFAULTS.kvm.migration_bandwidth=32`)? And if the cluster has still defaults, remember the admin to tune `migration_bandwidth` and `migration_downtime`? I find it difficult to generally recommended >1Gb/s.

lib/hypervisor/hv_kvm/__init__.py
saschalucas: Thanks @apoikos for clarification. Indeed I mixed things which do not belong together. However the article you have linked, recommends `native` for disk templates `plain`, `drbd` and `blockdev`, but `threads` for `file`, `sharedfile` and maybe `rbd`?

My point here is, whether we recommend only critical and generally valid configuration settings or tuning/best practices, that sometimes depends ...

saschalucas (approved):
Thanks @rbott, good work! From my point of view LGTM.

Regarding the discussion on `disk_aio` and `migration_bandwidth`, please decide on your own whether to keep it as or to make some adaptations.

lib/hypervisor/hv_kvm/__init__.py
rbott: I see your point here @saschalucas. Since this is an assessment of global/cluster-level settings, individual instances and their disk backend/template can not be taken into account. However, it would be possible to combine this check with the list of enabled disk templates. This would allow the message to be a bit more specific and reduce the chance of false alerts.

lib/hypervisor/hv_kvm/__init__.py
rbott: sure, that sounds reasonable. Ill add this!

lib/hypervisor/hv_kvm/__init__.py
rbott: Actually I re-read the article posted by @apoikos again. Since Ganeti uses what the article calls "raw image files", it also recommends using `native` there. It only recommends `threads` in the case of non-pre-allocated/sparse qcow2 images, which Ganeti does not use/support.

In that case I think it is pretty safe to assume `native` is the best choice here.

saschalucas: > When space efficient image files are used (QCOW2 without pre-allocation, or sparse raw images) the default of io='threads' may be better suited.

The later (sparse raw image) is used by `file`, `sharedfile`and maybe in analogy by `rbd`.

rbott: Oh, I did not even know that 🤦‍♂️

But wouldn't `rbd` be either a blockdevice or userspace access, depending on configuration?

saschalucas: In analogy `rbd` images are sparse (in block terminology thin). Even LVM can be thin today (thinpool). The article just mentioned the first write of an unallocated storage block. It may be argued, that on heavy workloads (where tuning make sens) a block is written more than once and hence aio=native could be better for all writes after the first. But who knows...

rbott: I tried to come up with a quick solution that is a "bit more intelligent" (with regards to the enabled disk templates). However, I eventually just dropped this part from the PR.

It adds the groundwork to assess hvparams and everything we require _right now_ is included (`machine_version` checks). 

We can improve from here and add more assessments in future PRs.

PR #1823, opened 2025-02-07
wait_for_qmp_event extended to multiple events

codefritzel:
This PR is an improvement of #1803.
The wait_for_qmp_event method has been modified so that it can now wait for multiple event types.
This is now used in HotDelDisk(), where a DEVICE_DELETED event is waited for once on a successful unplug or a DEVICE_UNPLUG_GUEST_ERROR on a failed unplug.


saschalucas (approved):
Thanks @codefritzel. LGTM.

PR #1826, opened 2025-02-09
Release 3.1.0-rc1

rbott:
This prepares the release of the 3.1.0-rc1 version of Ganeti. I tried to come up with a valid summary of the changes and include everything that is relevant to end users and/or distro maintainers.

Do you think anything is missing (or not accurate)?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,166 @@ News
 ====
 
 
+Version 3.1.0 rc1
+-----------------
+
+*(Released Tue, 11 Feb 2025)*
+
+Upgrade notes
+~~~~~~~~~~~~~
+
+Upgrading is supported from 3.0.x versions.
+
+If you are using the KVM hypervisor, you should consider setting the parameter
+`machine_version` to the latest `pc`-version supported by your version of kvm.
+You can query that by running `kvm -M ?`. Do not set this to `pc`, but rather
+to a specific version like `pc-i440fx-9.2`. Otherwise you might risk instance
+crashes if you live-migrate between different kvm versions (e.g. during a
+rolling cluster upgrade). The `q35` versions are not yet supported by Ganeti
+and will fail your instances to boot once kvm changes from the current default
+of `pc`.
+
+Ganeti now uses `-blockdev` instead of `-drive` parameters for Qemu/KVM. This
+has been tested with all built-in storage options. If you are using the
+`extstorage` backend, please make sure to test this in a non-production
+environment before upgrading.
+
+Important changes
+~~~~~~~~~~~~~~~~~
+
+Adapt to current Python versions
+++++++++++++++++++++++++++++++++
+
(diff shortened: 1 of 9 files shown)


dannyman (commented):
LGTM!

saschalucas:
thanks @rbott. It seams that `3.1` should have been in `ganeti.cabal` from the beginning. Can you please include this in your preparation?

saschalucas:
Maybe we can stress the point, that `--enable-regex-tdfa` is the only option from Debian Trixie onward, the only one tested in CI/QA **and** that `tdfa` does not provide PCRE type regexps. The place where this has an impact is unknown. Best guess seems to point to `gnt-filter(8)`.

rbott:
> Maybe we can stress the point, that `--enable-regex-tdfa` is the only option from Debian Trixie onward, the only one tested in CI/QA **and** that `tdfa` does not provide PCRE type regexps. The place where this has an impact is unknown. Best guess seems to point to `gnt-filter(8)`.

I have updated the NEWS file accordingly and also added a stance about `extstorage` (which might break if people use a custom storage provider that is somehow incompatible with the `-blockdev` parameter). Only situation I can think of would be a provider that does return a userspace URL instead of a blockdevice (if that even exists).

NEWS
```diff
@@ -2,6 +2,166 @@ News
 ====
 
 
+Version 3.1.0 rc1
```
saschalucas, line 5: Just for completeness, I assume skipping a beta release is intentional here? If yes, I'm fine with no beta.

NEWS
```diff
@@ -2,6 +2,166 @@ News
...
+
+*(Released Mon, 10 Feb 2025)*
+
+Upgrade notes
+~~~~~~~~~~~~~
+
+Upgrading is supported from 3.0.x versions.
+
+If you are using the KVM hypervisor, you should consider setting the parameter
+`machine_version` to `pc`. If it is unset, Ganeti will auto-detect the machine
```
saschalucas, line 16: I generally would not recommend setting `machine_version` to `pc`. Instead the admin should chose an explicit machine version (i.e. `pc-i440fx-7.2`). This guaranties safe live migration across qemu upgrades!
Only for the absentminded admin Ganeti defaults to `pc` on new clusters, which represents the _current_ qemu version and is not stable for live migration between different qemu versions. It just prevents a possible breaking switch to `q35`in future qemu versions.

rbott:
After some discussion with @saschalucas we decided to revert the changes regarding `machine_version` for 3.1

This has been reflected in the NEWS file and also in #1827

saschalucas (approved):
Thanks @rbott. LGTM.

apoikos (approved):
LGTM, thank you all for making this release possible!

apoikos:
I will perform a couple of tests before actually tagging the release, please bear with me for a little bit :-)

PR #1827, opened 2025-02-11
Do not use `pc` as the new default kvm `machine_version

rbott:
This commit partly reverts 86a41cecb24487f78ae867cefb0cadcd80c0df63 / #1822 

After some discussions with @saschalucas we concluded that the new default value of `pc` for the `machine_version` will only cause trouble in the long run. It might break instances during live-migration if the receiving end is a newer kvm release (e.g. during rolling cluster upgrades), since that end has a different understanding of what `pc` actually is. It would be more sane to configure an exact version, like `pc-i440fx-9.2`.

Documentation has been updated and also the cluster parameter assessment now warns if `machine_version` is set to unsupported values or to values that might cause trouble.


saschalucas (approved):
Thanks @rbott. LGTM.

PR #1829, opened 2025-03-04
QA: change default SSH key type to RSA

apoikos:
This doesn't even need a justification, but here it goes: DSA has long been deprecated and as of OpenSSH 9.8 completely unsupported (at least in Debian). Switch over to RSA instead.

This fixes QA runs in Debian trixie/unstable.


rbott:
LGTM. I think I "fixed" this on-the-fly in the QA system and totally forgot to fix it upstream 🤦‍♂️

PR #1831, opened 2025-03-06
Make builds reproducible

apoikos:
According to [1], ganeti does not build reproducibly due to some randomness in the generated docs. In particular, auto-generated documentation using utils.CommaJoin() on frozensets does not have stable output. Fix this by sorting the input passed to utils.CommaJoin in the two places that seem to introduce randomness.

[1] https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/ganeti.html


rbott (approved):
Nice, thanks for taking on this matter. And also I am kinda reliefed that it's only these two minor things blocking us from being reproducible :-)

PR #1833, opened 2025-03-06
Various build fixes to get `make distcheck` running

apoikos:
This PR includes several fixes to get `make distcheck` (almost) running again:

1. **Copy htest binary to the runner tempdir**: 
   - We now explicitly copy the **htest** binary to the temporary directory used for testing. This was necessary because of the Haskell build system changes that moved it to a different location (from `src/` to `exe/`).

2. **Fix Python test script paths**:
   - Adjust paths in Python test scripts to reflect the new structure after moving files to `test/py/legacy`.

3. **Update Haskell test dependencies**:
   - Correct the dependencies for Haskell tests in the Makefile to point to the new locations of `htools` and `mon-collector`.

4. **Call Cabal with full path**:
   - Modify the Makefile to use full paths for Cabal files during builds to ensure they are located correctly, especially during an out-of-source-tree ("[VPATH](https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html)") build, like the one performed by `make distcheck-release`.

5. **Remove .ghci from distribution**:
   - Remove the outdated `.ghci` reference from the distribution list in the Makefile since the file itself was removed previously from the source tree.

Note that `make distcheck-release` is still broken because of some integration test failures, which will be dealt with on a different PR.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -903,7 +903,7 @@ if ENABLE_METADATA
 HS_COMPILE_PROGS += exe/ganeti-metad
 endif
 
-CABAL_SETUP = runhaskell Setup
+CABAL_SETUP = runhaskell $(top_srcdir)/Setup
 
 # All Haskell non-test programs to be compiled but not automatically installed
 HS_PROGS = $(HS_BIN_PROGS) $(HS_MYEXECLIB_PROGS)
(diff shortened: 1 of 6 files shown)


rbott (approved):
LGTM, thanks for the cleanup/fixes @apoikos

PR #1835, opened 2025-03-11
Release 3.1.0-rc2

rbott:
The recent `-rc1` tag contained several issues regarding tests and `make dist[check]`, hence there never was an official release or even just a tarball. Thanks to @apoikos all problems have been addressed and we are now ready for `-rc2`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,26 @@ News
 ====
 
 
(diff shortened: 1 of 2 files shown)


rbott:
@apoikos done - I also added the PR IDs for reference (forgot that initially)

PR #1836, opened 2025-03-13
daemon-util: remove pidfile(s) on stop

apoikos:
The daemon pidfiles are created by start-stop-daemon and are not handled by the daemons themselves. Make sure we clean them up on stop. This fixes an autopkgtest failure on s390x.

Note: this patch is part of the Debian package.


rbott (approved):
LGTM, thanks :-)

PR #1837, opened 2025-03-16
doc/rapi.rst: remove randomness

apoikos:
I was too lazy to actually check that #1831 made builds reproducible - and reproducible.org indicates that it didn't.

Remove the last remaining bits of randomness that make ganeti builds unreproducible. This patch is already in unstable, so I did a second build and verified that it is indeed reproducible.


apoikos:
image

apoikos:
I'll take the YOLO path and merge sans review :smile:

apoikos:
image

PR #1849, opened 2025-05-16
Print saved and loaded ClusterData for debugging purposes

thielema:
Please do not merge!

It prints written `ClusterData` in `hscan` after `>>>>> writeDataInner >>>>>` and in other tools it prints received `ClusterData` after `>>>>> loadExternalData >>>>>`. Data from `loadExternalData` can come from any source, e.g. Luxi or text file.

`ClusterData` is written in Haskell syntax. You may indent and format it nicely using the [hindent](https://hackage.haskell.org/package/hindent) tool.


thielema:
After loading plain data, `loadExternalData` modifies the read data. Thus I print data both before and after postprocessing.

thielema:
The debug output may help resolving https://github.com/ganeti/ganeti/issues/1530.

PR #1853, opened 2025-06-03
Upgrade KVM runtime WRT to removal of `-chroot`

saschalucas:
Commit e607423b57521fec5a9a1d289ca736ec90266b18 has removed `-chroot` from the KVM runtime, but did not take into account, that instances started by <Qemu-9.0 needs to upgrade their runtime, too. Without that fix, a rolling upgrade to >=Qemu-9.0 wont be possible, if instances are using chroot.

Simple tests proved that chroot can be changed during live migration and doesn't need to be in the KVM runtime. However, no further knowledge exists why `-chroot` was in the runtime before.

### DIFF
--- a/lib/hypervisor/hv_kvm/kvm_runtime.py
+++ b/lib/hypervisor/hv_kvm/kvm_runtime.py
@@ -224,11 +224,17 @@ def _upgrade_serialized_runtime(loaded_runtime: List) -> List:
       hvparams[constants.HV_DISK_DISCARD] = constants.HT_DISCARD_IGNORE
 
     # remove chroot from the runtime
+    # commit e607423 addresses that the -chroot parameter is now combined into
+    # -run-with. This happens in a way that -chroot is no longer written into
+    # the KVM runtime.  When live migrating, the old runtime (containing
+    # -chroot) is read and therefore has to be filtered out. Otherwise kvm_cmd
+    # will contain old -chroot and new -run-with. This makes live
+    # migration to >=Qemu-9.0 possible.
     try:
       idx = kvm_cmd.index("-chroot")
     except ValueError:
       pass
     else:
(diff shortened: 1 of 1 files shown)


saschalucas:
> > However, no further knowledge exists why `-chroot` was in the runtime before.
> 
> What do you mean with this? Do you mean the code has no further information / context why it "finds" `-chroot` in the stored runtime (aka it does not take the actual hvparams into consideration)? I'd say that is on purpose since the whole point of the runtime information on disk is to make sure that you know how a particular instance was created/started, regardless of the _current_ hvparams stored in the configuration, right?

From my perspective the kvm runtime contains parts that can not change during live migration, otherwise the instance will crash. 

For curiosity `-chroot` lived in `_GenerateKVMRuntime()` while `-runas` in `_ExecuteKVMRuntime()`. For the later the code had a comment: https://github.com/ganeti/ganeti/blob/0decf59277e5a18c90d867a837cb377364305c6c/lib/hypervisor/hv_kvm/__init__.py#L1904-L1905  

While I can't find any hint that `-chroot` can not be changed during live migration. Now with >=Qemu-9.0 and `-run-with` both functionalities (chroot/user) are at the same parameter, which would complicate keeping just chroot in the runtime. So I tested and decided, that chroot does not need to be in the runtime any more.

saschalucas:
> Also I'd say we should release 3.1 after this has been merged, what do you say @apoikos @saschalucas?
> 
> Or do you think the recent changes to htools require `-rc3`?

The fixes after rc2 seems important for Trixie release. Weather rc3 or 3.1 looks just like a number to me :-). We still can release patch level versions if important things pop up.

codefritzel (changes requested):
Please comment more on why this is done.

PR #1862, opened 2025-06-17
QMP: call connect only if not connected

saschalucas:
The pure socket connection is already conditional in the `UnixFileSocketConnection` class. So it does not hurt to call connect() many times. However the subclass `QmpConnection` additionally handles QMP greeting, capabilities and supported commands. Since commit 5fa52d3630813001b4d4b797bcd79685dcc827ce the QMP connection is correctly marked as connected, the decorator `_ensure_connection` won't close it and further instances of `QmpConnection` will block / timeout. As a work around call connect() in the decorator only when not connected.

And while at it, close the connection if the greeting does not match a known key, instead of just setting the _connected flag.


codefritzel (approved):
Thanks @saschalucas 
LGMT :+1:

PR #1866, opened 2025-08-14
close non-decorator-opened QMP connection

saschalucas:
`GetInstanceInfo()` can't use the decorator `_with_qmp()`, because it has no instance object in its args, just a instance name. Since it opens a connection itself, it should also close it.

@rbott please test QA if this fixes #1865

PR #1869, opened 2025-08-21
use Debian Trixie and no Testing for the moment

saschalucas:
In the hope to fix CI tests WRT debian-testing in #1866.


codefritzel (approved):
Thanks @saschalucas

PR #1871, opened 2025-08-22
make: allow C.utf8 alongside C.UTF-8

saschalucas:
Since Debian glibc 2.34-1, the C.UTF-8 locale has been renamed to C.utf8 to match upstream naming. Update the grep command used to figure out the locale name to match against both names.

This patch is in [Debian](https://sources.debian.org/patches/ganeti/3.1.0~rc2-3/0003-make-allow-C.utf8-alongside-C.UTF-8.patch/), but missing upstream -> `./devel/release` fails without that.


rbott (approved):
Thanks for spotting this!

PR #1872, opened 2025-08-22
allow non contraction of can't

saschalucas:
Currently `troff` fails with variant: "troff: ... cannot break line". Allow both versions with and without contraction.


rbott (approved):
Oh boy :(

PR #1877, opened 2025-09-19
py-tests-legacy: fix invalid escape sequence

saschalucas:
Some legacy python tests include strings which later are used in re.search(), where special chars needs escaping. Currently a "SyntaxWarning: invalid escape sequence" is emitted. Fix this by double escape the backslash.


rbott:
I think the more pythonic way would be to use raw strings to define regexes:

regex = 'blah \(a|b\)'
vs
regex = r'blah (a|b)'

What do you think @codefritzel?

codefritzel (approved):
Thanks @saschalucas
LGTM

PR #1883, opened 2025-11-29
feat: add timeout param to rapi instance reboot & shutdown endpoint

codefritzel:
The timeout during shutdown can now be passed to the RAPI, so the default can be overwritten and it is no longer necessary to wait 120 seconds.

The endpoint `/2/instances/[instance_name]/shutdown` gets `timeout` as new parameter and `/2/instances/[instance_name]/`reboot gets `shutdown_timeout`.

### DIFF
--- a/doc/rapi.rst
+++ b/doc/rapi.rst
@@ -1374,8 +1374,8 @@ Reboots URI for an instance.
 
 Reboots the instance.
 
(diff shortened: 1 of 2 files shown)


saschalucas:
> However, this would mean that the RAPI documentation is incomplete in some places, such as instance-reboot.

True. PRs welcome :-).

PR #1838, opened 2025-03-30
Python 3.6 compatibility

jfut:
Hi, I am trying to build RPM packages for Ganeti 3.1.0-rc2. (https://github.com/jfut/ganeti-rpm/issues/53)

This patch is necessary because `RHEL/AlmaLinux/Rocky Linux/others 8` ships with `Python 3.6.8` as the default Python version.

The [INSTALL](https://github.com/ganeti/ganeti/blob/v3.1.0rc2/INSTALL) documentation still lists "Python, version 3.6 or above" in the Software Requirements section. But if you decide to drop support for Python 3.6, it's fine not to merge this patch as we can maintain it in the ganeti-rpm repository.

PR #1884, opened 2025-12-04
python-compatibiliy: for bitarray >=3 (part 2)

saschalucas:
As it turns out, the fix in commit
65574a9c08415d48af9e2e42c37f4fae517c7319 was incomplete. It handles only the return value of the signature change, but does not take care of the callers parameters. Commit ec714797092409db4200dfe07e4c66ad1b66284b introduces a new bus allocation abstraction layer, which hides the mismatch on the caller's parameters (by making relevant code paths obsolete).

An attempt to backport the bitarray >=3 compatibiliy fix to the stable-3.1 branch (i.e. for Debian Forky) unveils this problem. And while at it, make it Python 3.6 happy where Ganeti-3.1 should run, too.

### DIFF
--- a/lib/utils/bitarray_compat.py
+++ b/lib/utils/bitarray_compat.py
@@ -35,23 +35,24 @@ With bitarray-3 the signature of search() has changed:
   new: search(sub_bitarray, start=0, stop=<end>, /, right=False)
          -> iterator
 """
-from __future__ import annotations
 from typing import Union, List
 from bitarray import bitarray as _BaseBitarray
 from ganeti import errors
 
 class CompatBitarray(_BaseBitarray):
   # constructor forwarding for immutable C-Extension types
-  def __new__(cls, *args, **kwargs):
+  def __new__(cls: "type[CompatBitarray]", *args, **kwargs) -> "CompatBitarray":
     return _BaseBitarray.__new__(cls, *args, **kwargs)
 
-  def search(self, sub_bitarray: Union[bitarray, int], *args) -> List[int]:
+  def search(self, sub_bitarray: Union[_BaseBitarray, int], *args) -> List[int]:
     # compatibility: accept at most one optional int (old 'limit')
     if len(args) > 1:
       raise errors.GenericError("The bitarray.search() compatibility only "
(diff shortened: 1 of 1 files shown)


lib/utils/bitarray_compat.py
```diff
@@ -27,35 +27,39 @@
...
-    """
-    res = super().search(*args, **kwargs)
-    return res if isinstance(res, list) else list(res)
+  def search(self, sub_bitarray: Union[bitarray, int], *args) -> List[int]:
+    # compatibility: accept at most one optional int (old 'limit')
+    if len(args) > 1:
+      raise errors.GenericError("The bitarray.search() compatibility only "
+                                "supports one optional argument")
+
+    limit = args[0] if args and isinstance(args[0], int) else None
```
codefritzel: Here, different spellings of ifs are mixed together, long and short (var = ab if .. else).

At this point, I would turn the short if into a long one:
```
limit = None
if args and isinstance(args[0], int):
  limit = args[0] 
```

codefritzel:
Regarding Python 3.6, is it specified anywhere that this will still be supported? I thought the minimum version for supported Debian versions was Python 3.9?

If we do want to support Python 3.6 after all, then it should
1. be included in CI/CD (gh-actions) and 2. be documented somewhere (e.g. readme.md).

saschalucas:
> Regarding Python 3.6, is it specified anywhere that this will still be supported?

Until now only in my head :-/.

For reference: With PR #1838 / commit 3720eb845 the decision for Python-3.6 and Ganeti-3.1 was made.

> If we do want to support Python 3.6 after all, then it should
> 
>     1. be included in CI/CD (gh-actions) and 2. be documented somewhere (e.g. readme.md).

Since Ganeti-3.1 is now "stable", we have no CI therefore. And if we had, we don't have CI-images ready that would run Python-3.6. It's just our development model of doing backports into stable branches. Looking ahead here simplifies backporting, without losing much/any.

I've started documentation in a [Wikipage](https://github.com/ganeti/ganeti/wiki/Supported-Distributions).

PR #1903, opened 2026-02-11
Add design document for removal of asyncore dependency

rbott:
Ganeti depends on the deprecated Python asyncore module:

- HTTP server implementation used by RAPI and node daemons
- daemon eventloop
- inotify and process signalling mechanism used by the RAPI daemon to hot-reload the users file upon change
- UDP socket (client) implementation used by confd

This commit adds a design document explaining how to get rid of it and sets it as a goal for the 3.2 release.

I already have an implementation of the HTTP server part ready, but I would like to discuss / merge the design document first and update the implementation if required.

For Debian Trixie we worked around the removal of asyncore by installing the package `python3-pyasyncore`. However, its description states:

> This package contains the asyncore module as found in Python versions prior to 3.12. It is provided so that existing code relying on "import asyncore" is able to continue being used without significant refactoring.
>
> The module's source code is taken directly from the Python standard library. The specific version of asyncore.py used is  the last update before the addition of removal warnings at import time, and is essentially equivalent to the version provided with Python 3.9.
>
> Please note that new projects should prefer asyncio. 

Instead of relying on the outdated asyncore code (which might break with future Python releases?), we should migrate away from it ASAP. As you can see in the design document I think we do not need the complexities of asynchronous implementations and can safely fall back to e.g. threaded approaches. At least for my implementation of HTTP server this greatly reduced the amount of code and complexity (while still maintaining good performance). I have not done any proof-of-concept implementation for the other usecases of asyncore.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -797,6 +797,7 @@ docinput = \
 	doc/design-2.16.rst \
 	doc/design-3.0.rst \
 	doc/design-3.1.rst \
+	doc/design-3.2.rst \
 	doc/design-allocation-efficiency.rst \
 	doc/design-autorepair.rst \
 	doc/design-bulk-create.rst \
@@ -852,6 +853,7 @@ docinput = \
 	doc/design-qemu-blockdev.rst \
 	doc/design-reason-trail.rst \
 	doc/design-repaird.rst \
+	doc/design-replace-asyncore.rst \
 	doc/design-reservations.rst \
 	doc/design-resource-model.rst \
 	doc/design-restricted-commands.rst \
--- a/doc/design-3.2.rst
+++ b/doc/design-3.2.rst
@@ -1,0 +1,7 @@
+=================
+Ganeti 3.2 design
+=================
+
+The following designs have been implemented in Ganeti 3.2
+
+- :doc:`design-replace-asyncore`
--- a/doc/design-replace-asyncore.rst
+++ b/doc/design-replace-asyncore.rst
@@ -1,0 +1,232 @@
+=======================================================
+Replace Usage Of asyncore With Lightweight Alternatives
+=======================================================
+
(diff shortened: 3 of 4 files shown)


saschalucas:
Thanks @rbott for looking into this important step for Ganeti's future. Your design document sounds excellent to me, unfortunately, my skills are not sufficient to assess the feasibility and possibilities.

I'll wait a moment to give other the possibility for comments. Otherwise you got my LGTM.

rbott:
Hi @codefritzel,

thanks for your review/comments!

> Performance could suffer in Python with threads due to the Global Interpreter Lock (GIL). But maybe this isn't so critical in the Ganeti context?

On the one hand it really does suffer (as my draft implementation in #1905 shows). But on the other I do think its negligible as neither noded nor rapid are handling _that_ many requests per second in real-life scenarios. Although I have not yet figured out whether the slowdown comes from the threaded approach or if it is stdlib's `ssl` module vs `pyopenssl` (which might also be GIL-related). But then again Python 3.14 might speed up things considerably.

> 
> I had imagined that getting started with async.io would be more difficult, but I had the impression that it was very easy to learn. One argument against async.io would be that Python itself has implemented very few asynchronous functions internally and therefore does not offer any advantages when accessing filesystems, networks, etc. External libraries would have to be integrated in order to really benefit from it, and that would increase dependency management.

Looking at HTTP specifically it would be wise to build upon `aiohttp`, but as you said, that increases external dependencies. Also it most likely requires changes to RAPI-specific code because the interfaces do not match.

> 
> In my opinion, this is the best solution to replace asyncore for the time being.

Thanks!

> 
> Please reconsider pyinotify, otherwise LGTM

done :-)

rbott:
Thanks for your feedback, I'll move forward with this :-)

PR #1908, opened 2026-02-18
Add explicit IDE bus/unit allocation for KVM instances

ksanislo:
Recent QEMU versions have changed the default IDE bus assignment, placing IDE devices on the secondary bus instead of the primary. This causes very old FreeBSD VMs (plus anything else that doesn't support modern UUID based disk detection) to fail to boot properly. Add deterministic bus/unit assignment to all IDE devices (hard drives and CDROMs) so placement is consistent across all QEMU versions: primary-master, primary-slave, secondary-master, secondary-slave.

 ## Changes
   - `lib/hypervisor/hv_kvm/__init__.py`: Add `_GetIDEBusUnit()` helper, update `_GenerateKVMBlockDevicesOptions()`,
   `_CdromOption()`, and `_GenerateKVMRuntime()` to assign IDE bus/unit parameters
   - `man/gnt-instance.rst`: Document IDE bus slot assignment under `disk_type` and `cdrom_disk_type`

   ## QEMU output before/after (1 IDE disk + 1 IDE CDROM)

   **Before:**
   ```
   -device ide-hd,drive=disk-abc,write-cache=on
   -device ide-cd,drive=cdrom1
   ```

   **After:**
   ```
   -device ide-hd,bus=ide.0,unit=0,drive=disk-abc,write-cache=on
   -device ide-cd,bus=ide.0,unit=1,drive=cdrom1
   ```

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -167,6 +167,33 @@ _HOTPLUGGABLE_DEVICE_TYPES = {
 _PCI_BUS = "pci.0"
 _SCSI_BUS = "scsi.0"
 
+_MAX_IDE_DEVICES = 4
+
+
+def _GetIDEBusUnit(slot):
+  """Return the IDE bus and unit for a given slot index.
+
+  IDE controllers have two buses with two units each:
+    slot 0 -> ide.0, unit 0 (primary-master)
+    slot 1 -> ide.0, unit 1 (primary-slave)
+    slot 2 -> ide.1, unit 0 (secondary-master)
+    slot 3 -> ide.1, unit 1 (secondary-slave)
+
+  @type slot: int
+  @param slot: IDE slot index (0-3)
+  @rtype: tuple of (str, int)
+  @return: (bus, unit)
+  @raise errors.HypervisorError: if slot >= 4
+
+  """
+  if slot >= _MAX_IDE_DEVICES:
+    raise errors.HypervisorError(
+      "Too many IDE devices: slot %d exceeds maximum of %d"
+      % (slot, _MAX_IDE_DEVICES))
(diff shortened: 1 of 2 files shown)


rbott:
Hi @ksanislo,

I think with the recent introduction of the [bus manager module](https://github.com/ganeti/ganeti/blob/master/lib/hypervisor/hv_kvm/bus_manager.py) by @codefritzel the logic in this PR should be moved there. Not sure if the provided classes/abstractions really fit the IDE-related problem you are trying to solve but it would make sense to have that code live near the existing `SCSIAllocator` and `PCIAllocator` implementations.

What do you think? (both @codefritzel and @ksanislo)

ksanislo:
@rbott Thanks for the suggestion! I do agree that co-locating the IDE allocation logic in bus_manager.py makes sense from an organizational standpoint, and I've prototyped that approach here over here: https://github.com/ksanislo/ganeti/tree/kvm_ide_bus_untested

In this branch the IDEAllocator class now lives alongside PCIAllocator and SCSIAllocator in bus_manager.py. I did notice while putting it together that IDE doesn't fit very cleanly into the existing BusAllocator interface. It's not hotpluggable, has a fixed 4-slot topology, and its slot space is shared between hard drives and CDROMs which flow through separate code paths (_GenerateKVMBlockDevicesOptions vs _CdromOption). As such, in this prototype version I've kept it as a standalone class used directly rather than wiring it through the BusAllocatorManager.

I'm happy to go any direction you guys prefer here, of course. The original self-contained approach on kvm_ide_bus, or this version with the logic moved into the bus manager module. I would be curious to hear what @codefritzel thinks about whether the BusAllocator interface could/should be extended to cover IDE's constraints, or if keeping it as a separate class in the same file is a reasonable middle ground.

PR #1909, opened 2026-02-18
Make minimum VG size a configurable cluster parameter

ksanislo:
## Summary
- Add `min_vg_size` as a cluster-level parameter (default 20480 MiB) configurable via `gnt-cluster modify --min-vg-size`
- Replace hardcoded `MIN_VG_SIZE` constant in volume group checks with the cluster setting
- Include cfgupgrade support (upgrade adds default, downgrade strips it)
- Expose via QueryClusterInfo/RAPI

### DIFF
--- a/lib/cli_opts.py
+++ b/lib/cli_opts.py
@@ -144,6 +144,7 @@ __all__ = [
   "MAX_TRACK_OPT",
   "MC_OPT",
   "MIGRATION_MODE_OPT",
+  "MIN_VG_SIZE_OPT",
   "MODIFY_ETCHOSTS_OPT",
   "NET_OPT",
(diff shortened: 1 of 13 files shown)


ksanislo:
I still intend to fully patch things to play nice when an lvm vg doesn't exist at all, but implementing that would add scope creep and should probably just be a separate PR in the future. I've been running with a minvgsize=0 for years as it is though, and use a loopback mounted lvm "disk" that's something like 128k to keep ganeti happy. A complete patch will still use this same conf value set to zero to determine when it's safe to just work without lvm at all. Since I'm attempting to make PRs for all my previously existing patch sets, I fIgured I'd just include this for now too.

PR #1917, opened 2026-03-25
backport: qemu-10.2 compat: update RUNWITH_RE to support multiline help output

saschalucas:
needed for Ganeti-3.1 in upcoming Ubuntu 26.04.

Since QEMU 10.2 (commit 886898baad2), the `-run-with` option supports additional parameters, which causes the `chroot=` and `user=` parameters in the help output to appear on separate lines.

Update the corresponding regex to handle multiline output by enabling re.S and using a similar matching pattern as the regex arounds.

(cherry picked from commit 4bda73e9375a200839431c3a901ae25ac23ff0f7)


rbott (approved):
LGMT, thanks!

PR #1918, opened 2026-03-26
python-3.14 compat: drop FancyURLopener

saschalucas:
FancyURLopener has been deprecated for long time. Get rid of it and use urllib.request directly.

fixes #1916


rbott (approved):
LGTM, the QA passed, too (although there is not much going on with the burnin test right now). But I manually validated that at least the syntax parses well and I can start the burnin tool.

Issue #1913, opened 2026-03-10
Setting `link=` for routed network mode not possible


rbott:
With bridged network mode, Ganeti expects the `link` parameter to hold the name of bridge device. However with routed network mode, Ganeti uses the `link` parameter to specify a named routing table where it will add the route necessary to reach the instance. If the parameter is empty/not set, it will use the default routing table. 

See the following lines for the relevant shell code:

https://github.com/ganeti/ganeti/blob/9ade9886b04970add6ccb4dcd0106f76ed78eb6d/tools/net-common.in#L188-L196

Unfortunatly it is currently not possible to set `link` to an empty value:
```shell
# gnt-cluster modify --nic-parameters link=
Unhandled protocol error while talking to the master daemon:
Failed to parse request: SubmitJob: Can't parse value for 'nicparams': Can't parse value for 'link': Received empty value for non-empty list
```

The only workaround is to set it to `main` (which is the name of the default routing table in the Linux kernel). While it works, it will add an unnecessary `ip rule` for each tap interface. 

There are multiple solutions to this, e.g. (but of course not limited to:)
- allow `link` to be set to an empty value
- document the use of `main` and make the shell script check for `main` instead of an empty string

PR #1919, opened 2026-03-29
Redesign KVM ifup logic and add optional ifdown scripts

rbott:
The monolithic kvm-ifup script handled all network modes (bridged, routed, openvswitch) in a single file, making it difficult to customise individual modes without risk of side effects. This also meant there was no way to run teardown logic when an interface was removed.

Replace kvm-ifup/vif-ganeti with a dispatcher-based architecture:

- Add per-mode up/down scripts under /etc/ganeti/network/{bridged,routed,openvswitch,ext}/, each containing only the logic for that mode.
- Introduce kvm-nic-setup as a thin wrapper that sources net-common, validates the environment, and dispatches to the correct mode/action script.
- Add a new "ext" NIC mode (NMExt in Haskell, NIC_MODE_EXT in Python for fully custom network setups, blocked on Xen where it is not applicable.
- Add DeConfigureNIC() in hv_base.py and _DeConfigureAllNICs() in the KVM driver, called on instance shutdown, successful migration (source side), and failed migration (target side).
- Update vif-ganeti (Xen wrapper) to use the same dispatcher pattern.
- Move mode-specific functions (setup_bridge, setup_ovs, setup_route) out of net-common into the individual per-mode scripts.
- Add new NIC_ACTION_UP / NIC_ACTION_DOWN constants (Haskell + generated Python) and pass ACTION in the script environment.
- Add comprehensive documentation in doc/network.rst covering all modes, environment variables, and script call points.
- Update man pages to document the new "ext" mode.

Ideas for further improvements (within _this_ PR):
- expose the hypervisor type to the network scripts (so that the up/down scripts can act accordingly if required)

Ideas for separate PRs:
- add a `modify` action that triggers on modification of NIC parameters (e.g. change of VLAN, bridge etc.)

So far, this PR passes QA runs against Debian Trixie for KVM.


saschalucas:
>     * Add per-mode up/down scripts under /etc/ganeti/network/{bridged,routed,openvswitch,ext}/, each containing only the logic for that mode.

The old logic in the monolithic kvm-ifup script was: if a script in /etc exists use that, otherwise use the Ganeti shipped one.

The problem with /etc/ganeti is, that it is not versioned like the rest of the code. Installing a new Ganeti version besides an other will immediately overwrite the network scripts instead of waiting for `gnt-cluster upgrade` symlink switches. Downgrades won't be possible, the last installed Ganeti version wins.

Restoring old logic seems advisable here? Meaning shipping network scripts under $libdir with the possibility to override them in /etc ($confdir). The new network mode ext should just fail if no user supplied scripts can be found.

What do you think @rbott?

rbott:
> 
> The problem with /etc/ganeti is, that it is not versioned like the rest of the code. Installing a new Ganeti version besides an other will immediately overwrite the network scripts instead of waiting for `gnt-cluster upgrade` symlink switches. Downgrades won't be possible, the last installed Ganeti version wins.

My goal was to enable users to even modify/overwrite the default scripts in an easy way (without always resorting to "ext" mode). But I totally see your point. I will rework this PR so that the actual logic for the built-in network modes stays in the versioned folders.

There is also one thing I want to add/change for ext mode:

- with bridged/OVS mode, the ifup scripts are being executed *before* a live migration starts (on the destination node)
- with routed mode they are executed *after* the live migration has finished

With ext mode I want to be able to have both: e.g. create the `tap` interface pre-migration (and call a script of that name) and then call a second script "post-migration" after is has finished. But that needs a bit of a rework of the code. 

I'll be on vacation now and I'll look into that once I am back :-)

rbott:
I revisited the PR and integrated the following changes:

- scripts for the network modes `ovs`,`bridged` and `routed` are now stored again in the versioned file tree
- you can still override them with scripts dropped in `/etc/ganeti/network/$mode/$action`, *if* you want/need to
- scripts for the network mode `ext` are _only_ looked up in `/etc/ganeti/network/ext/up|down`
- in `ext` mode, the `up` script is now executed *before* and *after* a live migration
- there is a new environment variable `MIGRATION_PHASE` available in `ext` mode to distinguish:
  - if the `up` script is called during a live migration (as opposed to regular instance start or hotplugging of NICs)
  - if the `up` script is called at the start of a live migration
  - if the `up` script is called at the end of a (successful) live migration


I have also updated the documentation (doc/network.rst) to cover all variants.

*Update*: the QA failed due to a bug. Please save your time and do not review until I fix the issue :-)
*Update²*: QA works now for Fake, KVM and Xen hypervisor in bridged mode

saschalucas (approved):
Excellent work @rbott. 

I just have this on my todo list:

* add documentation for vlan aware bridge and the "default vlan 1"-problem ([link](https://groups.google.com/g/ganeti/c/cAbf2iRoIlA/m/gVm66rKqAQAJ)) and how to avoid it (`ip .... vlan_default_pvid 0`).
* OVS seems a bit broken (s. PR #1910)
* creating shelltests, that validate the up/down scripts (VLAN-handling) 

None of this needs to be done in this PR, but you may pickup something if you like. Otherwise please mere your changes.

rbott:
> Excellent work @rbott.

Thanks! :smile: 

>     * add documentation for vlan aware bridge and the "default vlan 1"-problem ([link](https://groups.google.com/g/ganeti/c/cAbf2iRoIlA/m/gVm66rKqAQAJ)) and how to avoid it (`ip .... vlan_default_pvid 0`).

I will incorporate that; since this PR also creates a bunch of missing documentation it fits in nicely.

>     * OVS seems a bit broken (s. PR #1910)

Yeah, I saw that too and wondered if I should integrate it here. But then I'll decided to only refactor and not try to add functional changes to the existing network types. On top of that, I do not have any OVS experience and there is no OVS support in the QA suite (yet). But AFAIU it is not that complicated. There is also #1913 which needs a fix, but I will also look into that separately as well.

> 
>     * creating shelltests, that validate the up/down scripts (VLAN-handling)

I'll look into that before merging as well, thanks for the hint!

rbott:
The shelltest part seems to blow up this PR, at least from a quick glance so I'll leave that to another time. But I have extended the documentation to cover the VID-1-related-problems as discussed on the [mailing list](https://groups.google.com/g/ganeti/c/cAbf2iRoIlA). I'll leave that here for @saschalucas to proof-read:

- https://github.com/ganeti/ganeti/blob/2a7d8ebc067b9b1d9a0bd569a7573e662d9575f4/doc/network.rst?plain=1#L67-L124
- https://github.com/ganeti/ganeti/blob/2a7d8ebc067b9b1d9a0bd569a7573e662d9575f4/doc/network.rst?plain=1#L224-L229

saschalucas:
> But I have extended the documentation to cover the VID-1-related-problems as discussed on the [mailing list](https://groups.google.com/g/ganeti/c/cAbf2iRoIlA). I'll leave that here for @saschalucas to proof-read:

Thanks, looks good!

Now that we have such a wonderful network documentation, it would be great if it were also visible. How about including it in `doc/index.rst under the "Using Ganeti" section? Once released, this should be visible like here: https://docs.ganeti.org/docs/ganeti/3.1/html/index.html#using-ganeti

rbott:
Oh wow, thanks for pointing that out @saschalucas. I completely forgot to link to the new documentation from various places (I only added it to the hidden toctree, which does not help at all :smile:). I will move forward and merge this now with proper documentation in place.

PR #1920, opened 2026-03-31
Replace the AsyncUDPSocket confd client with a synchronous client

rbott:
This PR removes the dependency of the confd client on the deprecated asyncore framework (further implementing the design doc introduced in #1903). Instead it replaces it with a stanard synchronous socket client.

Upon investigation it turned out that all uses of the confd client were synchronous anyways, so there will not be any observable performance drops.

The relevant tests have also been migrated from the legacy unit test format to the newer pytest based variant.

QA tests have passed for the KVM and fake hypervisor implementations with now visible impact.


saschalucas (approved):
Thanks @rbott. LGTM.

PR #1924, opened 2026-04-08
Improve QA Suite by clearing job queue and run some tests less often

rbott:
The QA suite has accumulated quite some runtime if you use e.g. KVM and enable most tests (> 6 hours). This PR improves this by two steps:

- the QA regularly tests/analyzes the job queue. in very long QA runs the queue accumulates thousands of jobs which do not expire during the QA's runtime. while the queue-related tests initially only take 1-2 seconds, they accumulate a runtime of over a minute near the end of the QA run. We now employ `gnt-job autoarchive` every now and then to keep the queue clean.
- we run the instance test block over and over again with iterations over possible HVparams. Some tests in there are not affected by the HVparams (e.g. TestInstanceList, TestNodeList and others). We can execute them in the first run and safely exclude them from further instance test iterations without negatively impacting the results of the QA suite

Overall this has brought the QAs runtime on Debian Forky down from 6:30 to 4:54 which is quite an improvement :-)

### DIFF
--- a/qa/ganeti-qa.py
+++ b/qa/ganeti-qa.py
@@ -867,6 +867,12 @@ def RunInstanceTests():
         else:
           test_desc = "Iterating through hypervisor parameter values"
           ReportTestSkip(test_desc, "instance-iterate-hvparams")
+
+        # Archive completed jobs to prevent large job queues across
+        # long QA runs.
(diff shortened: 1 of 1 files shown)


saschalucas (approved):
Thanks @rbott. LGTM.

PR #1927, opened 2026-04-10
Replace the asyncore-based daemon event loop with selectors

rbott:
With this PR we are removing the final bits of asyncore from Ganeti (see #1710, #1903). We replace the daemon event loop with a selectors-based approach (as outlined in the design document), move tests over to the pytest framework and drop all remaining references to asyncore.

With that (and the remaining Python 3.14 compatibility patches) we are closer to a 3.2 release than ever :smile:


saschalucas (approved):
Thanks @rbott, LGTM.

PR #1928, opened 2026-04-19
kvm: always enable debug-threads in QEMU -name argument

rbott:
QEMU supports a debug-threads sub-parameter appended comma-separated to the -name argument (e.g. `-name instance,debug-threads=on`), which causes QEMU to assign meaningful names to its threads instead of generic ones. This aids observability when inspecting the host with tools like top.

The feature was previously implemented as an optional HV parameter (#1528) but had to be reverted (#1822) because the instance name parser did not strip the comma-separated suffix, causing downgraded clusters to lose track of running instances (#1820).

The parser fix landed as part of #1822 and has been in place since. There is no performance impact, so the parameter is now unconditionally enabled rather than gated behind a new HV parameter (so there is no need for up/downgrade code).

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1248,8 +1248,14 @@ class KVMHypervisor(hv_base.BaseHypervisor):
     pidfile = self._InstancePidFile(instance.name)
     kvm = hvp[constants.HV_KVM_PATH]
     kvm_cmd = [kvm]
-    # used just by the vnc server, if enabled
-    kvm_cmd.extend(["-name", instance.name])
+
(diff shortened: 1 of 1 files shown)


rbott:
Oh I guess I need to revisit this and add a qemu version constraint. See the latest qemu 11.0 changelog:
> The "debug-threads" argument of "--name" is now deprecated. Thread naming is unconditionally enabled on all platforms where it is supported.

Source: https://wiki.qemu.org/ChangeLog/11.0

I will update this PR later so that debug-threads will only be configured on qemu versions < 11

rbott:
I have updated the PR so that debug-threads will be omitted for QEMU versions 11 and up

saschalucas (approved):
Thanks @rbott. I'm glad that we ended up without a new HV parameter :-).

Issue #846, opened 2017-06-24
Does not wait for DRBD connect and uses automatic split-brain resolving


Ganeti-Issues-Migrator:
Originally reported of Google Code with ID 790.

```
Ganeti's use of DRBD is completely unsafe.
It does not wait for the nodes to connect before promoting a node to primary, it uses automatic split-brain resolution and it is unaware of errors during operation.

Environment: Ganeti 2.10.1 on Debian Wheezy

| gnt-cluster (ganeti v2.10.1) 2.10.1

| Software version: 2.10.1
| Internode protocol: 2100000
| Configuration format: 2100000
| OS api version: 20
| Export interface: 0
| VCS version: (ganeti) version v2.10.1

| hspace (ganeti) version v2.10.1
| compiled with ghc 7.4
| running on linux x86_64

Ganeti's DRBD seettings:
| disk {
|         on-io-error             detach;
| }
| net {
|         after-sb-0pri           discard-zero-changes;
|         after-sb-1pri           consensus;
| }

Log for a "failed" attach:
| block drbd14: Starting worker thread (from drbdsetup [16187])
| block drbd14: disk( Diskless -> Attaching ) 
| block drbd14: Found 6 transactions (7 active extents) in activity log.
| block drbd14: Method to ensure write ordering: flush
| block drbd14: drbd_bm_resize called with capacity == 33554432
| block drbd14: resync bitmap: bits=4194304 words=65536 pages=128
| block drbd14: size = 16 GB (16777216 KB)
| block drbd14: bitmap READ of 128 pages took 1 jiffies
| block drbd14: recounting of set bits took additional 0 jiffies
| block drbd14: 0 KB (0 bits) marked out-of-sync by on disk bit-map.
| block drbd14: disk( Attaching -> UpToDate ) 
| block drbd14: attached to UUIDs E262D6FE3DC2D38C:0000000000000000:0001000000000004:0000000000000004
| block drbd14: conn( StandAlone -> Unconnected ) 
| block drbd14: Starting receiver thread (from drbd14_worker [16188])
| block drbd14: receiver (re)started
| block drbd14: conn( Unconnected -> WFConnection ) 
| block drbd14: role( Secondary -> Primary ) 
| block drbd14: new current UUID 7F0194F8DCDF78C5:E262D6FE3DC2D38C:0001000000000004:0000000000000004

Because of this we lost data in a way DRBD usually protects against.
This happened in a cluster with DRBD with two nodes (NodeA and NodeB).

[...]

PR #1929, opened 2026-04-19
Close DRBD handshake race in AssembleInstanceDisks

rbott:
This PR addresses #846 in a way that does not introduce new cluster parameters. 

The two-pass assemble only narrowed the window where a DRBD device could be promoted while still in WFConnection. Between passes, reuse the existing `drbd_wait_sync` RPC on the secondaries to wait for the handshake; failures are logged but do not abort. The logic does of course not apply to non-DRBD blockdevices.

To avoid new cluster parameters (for now), we resort to document `net-custom` as the supported way to override the hardcoded after-sb-*pri policies. With DRBD utils, the latest occurence of CLI parameters always "wins", hence it is possible to override the cluster default policies.

### DIFF
--- a/lib/cmdlib/instance_storage.py
+++ b/lib/cmdlib/instance_storage.py
@@ -1596,12 +1596,13 @@ def AssembleInstanceDisks(lu, instance, disks=None, ignore_secondaries=False,
 
   # With the two passes mechanism we try to reduce the window of
   # opportunity for the race condition of switching DRBD to primary
-  # before handshaking occured, but we do not eliminate it
-
-  # The proper fix would be to wait (with some limits) until the
-  # connection has been made and drbd transitions from WFConnection
-  # into any other network-connected state (Connected, SyncTarget,
-  # SyncSource, etc.)
+  # before handshaking occured. For DRBD instances with reachable
+  # secondaries we additionally poll the secondaries after pass 1 to
+  # confirm the DRBD handshake transitioned out of WFConnection before
+  # we promote the primary. On handshake failure we abort pass 2
+  # unless the caller has opted into the risk via ignore_secondaries
+  # (mapped from the CLI '--force' / '--ignore-secondaries' flags) or
+  # the secondary is administratively offline.
 
   # 1st pass, assemble on all nodes in secondary mode
   for idx, inst_disk in enumerate(disks):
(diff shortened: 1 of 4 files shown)


lib/cmdlib/instance_storage.py
```diff
@@ -1623,7 +1623,26 @@ def AssembleInstanceDisks(lu, instance, disks=None, ignore_secondaries=False,
...
+  # instances without secondaries (single-node / diskless).
+  if any(d.dev_type == constants.DT_DRBD8 for d in disks):
+    secondary_nodes = lu.cfg.GetInstanceSecondaryNodes(instance.uuid)
+    if secondary_nodes:
+      wait_result = lu.rpc.call_drbd_wait_sync(secondary_nodes,
+                                               (disks, instance))
+      for node_uuid, nres in wait_result.items():
+        node_name = lu.cfg.GetNodeName(node_uuid)
+        if nres.fail_msg:
+          lu.LogWarning("DRBD handshake wait failed on secondary %s: %s;"
```
apoikos: If we are to provide strong guarantees against data loss, this should be a hard failure rather than a simple warning,  or at least fail hard unless `ignore_secondaries` is `True`. If we were to turn this to a hard failure, what potential breakage would we expect?

apoikos (commented):
Thanks for looking into this @rbott!

rbott:
Thanks, trying to clear all items on the 3.2 roadmap/milestone list. Although I allowed myself to move the q35/UEFI stuff to 3.3 (as that is probably to big and will take some more time to implement properly).

lib/cmdlib/instance_storage.py
rbott: You're absolutely right, we should hard-fail here. I have added some code to distinguish the "non-error" cases and tried to add descriptive error/warning messages for each scenario. 

The QA suite is currently running, but that will take several hours to complete

rbott: (but it probably needs more manual testing anyways to cover corner cases)

rbott: The QA suite passed on Debian Trixie and Forky with KVM / DRBD backends. In general I'd say there is no breakage to expect with regular operations. Also the special cases (secondary node offline etc.) should be covered by the updated code and not break emergency operations.

apoikos (approved):
LGTM, thanks!

PR #1895, opened 2025-12-23
Extend RAPI authorization with granular role-based access control

codefritzel:
Replace the simple read/write permission model with a granular role-based access control system supporting ~50 individual permissions.

Key changes:
- Add granular permissions (e.g. jobs.list, instances.create, nodes.migrate) for fine-grained access control
- Introduce role system with @admin and `@readonly` predefined roles
- Support three permission formats in rapi_users file:
  - `@role` (e.g., `@admin`, `@readonly`)
  - Explicit permissions (e.g. jobs.list,instances.query)
  - Role with modifications (e.g. `@readonly,+instances.create,-jobs.list`)
- Add per-handler AUTH_REQUIRED attribute for public endpoints
- Update all RAPI handlers with appropriate granular permissions

Files changed:
- `lib/rapi/__init__.py`: Define permission constants and roles
- `lib/rapi/baserlib.py`: Add default permissions and `AUTH_REQUIRED`
- `lib/rapi/rlib2.py`: Assign granular permissions to all handlers
- `lib/http/auth.py`: Add _resolve_permissions() and has_permission()
- `lib/server/rapi.py`: Implement new authorization logic
- `lib/build/sphinx_ext.py`: Handle string permissions in docs
- `doc/rapi.rst`: Document new permission system
- `qa/qa_rapi.py`: Update test user to use `@admin` role
- Unit tests updated for new permission model

---
closes #1894

PR #1934, opened 2026-04-28
Add role-based access control for the Remote API

rbott:
This PR picks up where #1895 by @codefritzel left. The original PR does not apply cleanly any more because of the changes to the HTTP server implementation (transition from asyncore to threaded model). Changes were mostly related to unit tests though.

We replace the binary read/write permission model with a granular permission system featuring ~66 individual permissions and two predefined roles (`@admin`, `@readonly`). The rapi_users file now accepts three formats:

  - role reference: `@admin`
  - explicit permissions: `jobs.list,instances.query`
  - role with modifications: `@readonly,+jobs.cancel,-cluster.info`

Each RAPI handler declares the permission it requires via `*_ACCESS`, and may opt out of authentication entirely with `*_AUTH_REQUIRED=False` (used for /, /version, /2, /2/features). **Legacy `read`/`write`/`all` options keep working via a deprecation-mapped fallback** (with a deprecation warning in the logs).

The following additions have been made by me (analyzed and partially implemented by Claude Opus 4.7, as reflected in the commit message):

  - Code & tests adapted to the removal of asyncore (new HTTP server implementation)
  - _resolve_permissions() processes tokens in a fixed order so `@admin,-x` and `-x,@admin` produce the same result
  - 403 responses no longer echo the required permission name
  - Unknown `@role` tokens log a warning instead of failing silent
  - Deprecation warnings dedupe per file load
  - Sphinx ext asserts on RAPI_ACCESS_NOT_DEFINED
  - Permission tests

The commit message cites @codefritzel as Co-Author, I hope that is fine with you :-)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2043,6 +2043,7 @@ python_tests_unit = \
 	test/py/unit/test_daemon.py \
 	test/py/unit/hypervisor/hv_kvm/test_monitor.py \
 	test/py/unit/hypervisor/hv_kvm/test_bus_manager.py \
+	test/py/unit/http/test_auth.py \
 	test/py/unit/http/test_server.py \
 	test/py/unit/http/test_ssl.py \
 	test/py/unit/server/test_rapi.py
--- a/NEWS
+++ b/NEWS
@@ -3362,10 +3362,10 @@ New features
   just includes the DRBD data collector, that can be executed by calling
   ``mon-collector`` using the ``drbd`` parameter. See
   :manpage:`mon-collector(7)`.
-- A new user option, :pyeval:`rapi.RAPI_ACCESS_READ`, has been added
+- A new user option, ``rapi.RAPI_ACCESS_READ``, has been added
   for RAPI users. It allows granting permissions to query for
   information to a specific user without giving
-  :pyeval:`rapi.RAPI_ACCESS_WRITE` permissions.
+  ``rapi.RAPI_ACCESS_WRITE`` permissions.
 - A new tool named ``node-cleanup`` has been added. It cleans remains of
   a cluster from a machine by stopping all daemons, removing
   certificates and ssconf files. Unless the ``--no-backup`` option is
--- a/doc/rapi.rst
+++ b/doc/rapi.rst
@@ -43,53 +43,65 @@ Options control a user's access permissions. The section
 :ref:`rapi-access-permissions` lists the permissions required for each
 resource. If the ``--require-authentication`` command line option is
 given to the ``ganeti-rapi`` daemon, all requests require
-authentication. Available options:
+authentication.
 
-.. pyassert::
+Permissions can be specified in three formats:
 
-  rapi.RAPI_ACCESS_ALL == set([
-    rapi.RAPI_ACCESS_WRITE,
-    rapi.RAPI_ACCESS_READ,
-    ])
+Role-based access
+  Use ``@rolename`` to assign a predefined role. Available roles:
 
-.. pyassert::
+  ``@admin``
+    Full access to all RAPI resources (all permissions).
+  ``@readonly``
+    Read-only access to query information (e.g. ``jobs.list``,
+    ``instances.query``, ``nodes.list``, ``cluster.info``, etc.).
+  ``@operator``
+    Everything in ``@readonly`` plus instance operations that do not
+    change instance configuration: start/shutdown/reboot, migrate and
+    fail-over, replace disks, prepare/create/remove exports, and
+    console access. Cannot create, modify, remove or rename instances,
+    and cannot change cluster, node, network or group configuration.
 
-  rlib2.R_2_nodes_name_storage.GET_ACCESS == [rapi.RAPI_ACCESS_WRITE]
+Explicit permissions
+  Specify individual permissions separated by commas, e.g.
+  ``jobs.list,instances.create,nodes.query``. See
+  :ref:`rapi-access-permissions` for the full list of available
+  permissions.
 
-.. pyassert::
+Role with modifications
+  Start with a role and add/remove specific permissions using ``+`` and
+  ``-`` prefixes, e.g. ``@admin,-instances.remove`` (admin without
+  instance deletion) or ``@readonly,+jobs.cancel`` (readonly plus job
+  cancellation).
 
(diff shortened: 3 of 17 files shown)


rbott:
Note: QA passes 


The QA uses the `@admin` role and due to the available unit tests I dont think we have to add more tests to the QA. Although I guess it would be trivial to do so. I'll reconsider that before merging (if someone approves this PR 😃)

saschalucas:
Thanks @rbott for continuing @codefritzel's work.

Cloud it be possible, that the RAPI documentation needs further tuning? If looking for example at [this](https://docs.ganeti.org/docs/ganeti/3.1/html/rapi.html#groups-group-name-tags), there is a table listing available methods and their needed, now legacy, permissions like `write` for `PUT` or `DELETE`. Ideally they should be updated to `@admin` and the corresponding explicit permission, i.e. `groups.tag`.

Which leads to the second question: I got the impression, that for every RAPI resource a new explicit permission was created, but the group-tag example above seems not to have one? Other tag resources also, like instance-tags.

If every RAPI resource has an explicit permission, it could be somehow generated, instead of listed in `lib/rapi/__init__.py`?

Also the man page of `ganeti-rapi` should be updated under section `ACCESS CONTROLS` to reflect the new default required read permission for query / GET operations.

rbott:
> 
> Cloud it be possible, that the RAPI documentation needs further tuning? If looking for example at [this](https://docs.ganeti.org/docs/ganeti/3.1/html/rapi.html#groups-group-name-tags), there is a table listing available methods and their needed, now legacy, permissions like `write` for `PUT` or `DELETE`. Ideally they should be updated to `@admin` and the corresponding explicit permission, i.e. `groups.tag`.

I have to admit I mostly focussed on the code / merge part. I'll look into the documentation again and check if updates are necessary.

> Which leads to the second question: I got the impression, that for every RAPI resource a new explicit permission was created, but the group-tag example above seems not to have one? Other tag resources also, like instance-tags.

I'll check that, too!

> If every RAPI resource has an explicit permission, it could be somehow generated, instead of listed in `lib/rapi/__init__.py`?

I'll also check if that is possible. Maybe decorators/annotations might be a way to streamline this a bit more.

rbott:
I have updated the PR with a few things:

- documentation is now fixed/complete both in the man page as well in the RAPI docs
- I have removed the duplicated listing of permissions; the `RAPI_ACCESS_ALL` Frozenset is generated by enumerating all RAPI_ACCESS_ constants via introspection
  - adding new API routes / methods without specifying permissions should also be uncovered by tests/docs-generation because they will fail if the associated permission name can not be looked up (that one was already there, just double-checked)
- there is a new `@operator` role which is sort of a middle-ground between `@readonly` and `@admin` (allowing a user to manage instances but not modify/remove them)


Regarding the Group Tags: all of the tags-related API routes (network, instances, groups etc.) are handled by the same code and hence they are all collectively covered by the `tags.get`, `tags.set`, and `tags.delete` permissions. The documentation also covers that.

*Update*: I have also extended the QA suite to explicitly test that `@readonly` works and blocks access to modifying API routes. I'll re-run the QA suite and update this post once that has finished.
*Update2*: It works for the fake hypervisor (which already covers the newly added QA tests). The KVM hypervisor test is still running, but already passed the relevant QA steps succesfully so I guess it is safe to say that it does not break anything.

saschalucas:
Thanks @rbott.

>     * I have removed the duplicated listing of permissions; the `RAPI_ACCESS_ALL` Frozenset is generated by enumerating all RAPI_ACCESS_ constants via introspection

OK. Maybe I'm to confused or github is showing something wrong? I can see this by https://github.com/ganeti/ganeti/compare/28b1a14b5861899726f13149bb1bf1bc3ccd3b08..9afa982a8146cac18bd7800f6b22ea9c4b1c848e, but in current [changes](https://github.com/ganeti/ganeti/pull/1934/changesl) there is the list again visible. Can you please check?

If you find it better to keep the list, I would be fine.

>     * there is a new `@operator` role which is sort of a middle-ground between `@readonly` and `@admin` (allowing a user to manage instances but not modify/remove them)

Just in case having different operator roles, would it make sens to name them more precise? i.e. `instance-operator`, `node-operator` ... ?

> Regarding the Group Tags: all of the tags-related API routes (network, instances, groups etc.) are handled by the same code and hence they are all collectively covered by the `tags.get`, `tags.set`, and `tags.delete` permissions. The documentation also covers that.

Sorry, I missed that. From the RAPI docs I got the impression, that different kinds of tagging belong to different resources like instances, nodes, cluster etc. But it seems that the new permissions are for opcodes instead of RAPI-resources (which do not always correlate)??? And there tagging is a single opcode. Does this make sense?

rbott:
> OK. Maybe I'm to confused or github is showing something wrong? I can see this by https://github.com/ganeti/ganeti/compare/28b1a14b5861899726f13149bb1bf1bc3ccd3b08..9afa982a8146cac18bd7800f6b22ea9c4b1c848e, but in current [changes](https://github.com/ganeti/ganeti/pull/1934/changesl) there is the list again visible. Can you please check?

It looks correct to me (verified in the branch directly/locally).

> 
> Just in case having different operator roles, would it make sens to name them more precise? i.e. `instance-operator`, `node-operator` ... ?

I would assume that a node-operator would also need instance-operator level (because maintaining nodes also includes live-migrating and instance start/stop). I'd say we either keep the operator thing and see what people need/request or drop it again and start with just the two pre-defined roles (and see what people need/request :slightly_smiling_face: ). What do you think?

> Sorry, I missed that. From the RAPI docs I got the impression, that different kinds of tagging belong to different resources like instances, nodes, cluster etc. But it seems that the new permissions are for opcodes instead of RAPI-resources (which do not always correlate)??? And there tagging is a single opcode. Does this make sense?

The permissions are assigned in `lib/rapi/rlib2.py` which maps RAPI resources to classes and from there it is mapped to opcodes. If users really need fine-grained permissions for tags, I think we can add this later and deprecate (or keep) the existing broader tag permission levels.

saschalucas (approved):
Thanks @rbott for picking up @codefritzel work.

Looking twice and more precise I'm no longer confused regarding duplicated listing of permissions. Nothing is duplicated.

Regarding the name of the instance operator role I'm also fine to keep the name until further notice. The same holds true for permissions as is, specially regarding tags.

Please merge your changes.

rbott:
Thanks for reviewing! :-)

PR #1937, opened 2026-05-08
Close DRBD resync race for Grow Operation

rbott:
The xen-pvm QA suite has been failing for quite a while now. I tasked an Opus 4.7 agent with researching the issues. It turned out that there is a race in the DRBD grow operation that is however somewhat specific to the QA suite case.

At one point, the QA suite issues three grow operations in rapid succession (`TestInstanceGrowDisk`). `drbdsetup resize` returns before DRBD has coordinated the new size with the peer; for several seconds the connection state stays `Connected` while resync of the freshly-grown extent is being scheduled. During that window the LU's `WaitForSync` polls the primary's mirror status, sees `sync_percent=None`, and reports the disks as in sync. The follow-up gro request then races into the just-resized device and DRBD rejects it with `ERR_RESIZE_RESYNCING` (130). 

The fix is: after issuing the resize, wait up to `_GROW_RESYNC_TIMEOUT` seconds polling /proc/drbd until `is_in_resync` flips True; the caller's WaitForSync then correctly waits the resync out. If the wait times out the grow returns - no resync is required. For now, `_GROW_RESYNC_TIMEOUT` is set to `10`.

This has successfully been tested against xen-pvm and KVM with full QA runs. I cannot test xen-hvm right now because that is broken for other reasons.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -287,6 +287,7 @@ PYTHON_TEST_DIRS = \
 	test/py/unit/hypervisor \
 	test/py/unit/hypervisor/hv_kvm \
 	test/py/unit/server \
+	test/py/unit/storage \
 	test/py/unit/test_data \
 	test/py/integration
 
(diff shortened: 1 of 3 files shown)


saschalucas (approved):
Thanks @rbott. Nice catch.

rbott:
Thanks @saschalucas!

PR #1938, opened 2026-05-10
KVM: support q35 (PCIe) machine type

rbott:
This PR adds prelimenary support for the `q35` machine type and PCIe to Ganeti. It builds upon the work done by @codefritzel, thanks for that :-)

q35 splits cleanly into two device categories. Cold-boot static PCI devices (balloon, SCSI controller, qemu-xhci, SPICE/QGA virtio-serial channels, paravirtual CD-ROMs) are packed as functions of a single multifunction group on `pcie.0` slot `0x02` - same trick the q35 chipset uses for its own LPC/SATA/SMBus integrated devices. Hot-pluggable leafs (NICs, virtio disks) attach to a pre-allocated pool of 24 empty pcie-root-ports on `pcie.0`, the pool size matches Ganeti's `MAX_NICS + MAX_DISKS` hard cap, so an instance can never run out of slots. Pre-allocation is required because `pcie.0` itself is not hot-pluggable in QEMU, so hot-add and hot-del are single-step against an already-existing pool root-port.

To make life easier for users of predictable interface names, each pool root-port is tagged at cold-boot with a sequential ACPI index, so systemd assigns predictable `enoX` names. Surviving NICs always keep their names across hot-add/remove cycles. A NIC hot-added into a slot freed by an earlier hot-remove will inherit that slot's `enoX` name - name reuse on the slot-reuse path is a deliberate KISS tradeoff in exchange for not doing runtime acpi-index bookkeeping.

q35 machines do not have an explicit USB controller, so we add qemu-xhci for use with the USB tablet/mouse. It lives at function 2 of the static multifunction group on slot 0x02.

For CD-ROM, Floppy, VGA etc. there may be sub-optimal combinations now (e.g. IDE CD-ROM, LSI SCSI controller etc. are not compatible or at least not recommended to be used with q35). But instead of silently converting settings we resort to warning users of possible issues. The handling of this situation could be improved in a future PR.

There is also now a "resolver" for the machine type which only emits either `q35` or `i440fx` and persists that in the runtime (to ease future lookups during the lifetime of the instance). The resolver tries to be future proof since QEMU might change the `pc` identifier to resolve to a `q35` variant instead of a `i440fx` variant.

There are new tests for various components but ultimately I think we should move `GenerateKVMRuntime()` (and related helper functions) into its own module and refactor it so that the whole command line generation process can be tested. But this is out of scope for this PR :-)

Please feel free to review and / or comment on the design doc and also on the code itself. Of course I am also happy for any testing. But keep in mind the code might change over time as I run tests myself and fix problems. I just wanted to share the work early.

Spoiler: I have also done some very basic scouting / testing for UEFI support (only basic UEFI, not secure boot/tpm support). I can share a design doc soon I guess.


rbott:
Quick Update: it is basically working.

But I stumbled across various problems with IDE (because `q35` - by default - does not ship with an IDE controller) and decided to dump both IDE and Floppy support from `q35`. I am not sure if these are really required features for modern VMs and they make things way more complicated.

The QA suite has been updated to not check IDE disk types when a `q35`-class `machine_version` is configured.

saschalucas:
Thanks @rbott, I'm glade that you started working on q35. Will take a look at the design first and comment inline...

doc/admin.rst
```diff
@@ -1616,6 +1616,79 @@ NIC types like ``e1000``. Please check the :manpage:`gnt-instance(8)` man page f
...
+On q35, Ganeti tags each NIC's PCIe root port with an ACPI ``acpi-index``
+attribute. systemd-udev uses that attribute via the ``net_id`` builtin to
+assign an ``eno<N>`` name. The mapping is:
+
+  NIC enumeration position (0-based) + 1 → ``eno<N>``
+
+So a freshly created q35 instance with three NICs gets ``eno1``, ``eno2``
+and ``eno3`` inside the guest, mirroring the order Ganeti lists them in.
+
+On ``pc-i440fx-*``, no ``_DSM`` is exposed, so systemd falls back to its
```
saschalucas: `_DSM` seems not explained before. I assume it is related to `acpi-index`?

rbott: Yes, it is. I read the whole document again this morning and actually found a few ambigious / unclear paragraphs. I will try to rephrase them today.

doc/admin.rst
```diff
@@ -1616,6 +1616,79 @@ NIC types like ``e1000``. Please check the :manpage:`gnt-instance(8)` man page f
...
+Index gaps after NIC removal
+............................
+
+``acpi-index`` values are intentionally **not** renumbered when a NIC is
+removed. If an instance has three NICs (``eno1``, ``eno2``, ``eno3``) and
+``eno2`` is hot-removed, the surviving NICs keep their names: the guest
+sees ``eno1`` and ``eno3`` with no ``eno2`` in between. The next NIC added
+later gets ``eno4`` (``max(existing) + 1``). This is by design - silently
+shifting interface names of running NICs would break in-guest network
+configuration that is keyed on those names.
```
saschalucas: It should be noted, that the ugly Ganeti concept of NIC-hot-modify, i.e. if you are changing the link (bridge) a NIC is connected to or adding VLANs to a trunk, will hot-remove and hot-add a NIC. Other than with the pc-machine, the name won't be stable unless also using net.ifnames=0.

rbott: Yeah, I stumbled across this. This "modify" behaviour is especially annoying since hot-removing is now slower because it needs an acknowledgement from the guest (and will fail if the guest does not ack within 30 seconds for whatever reason). 

But with the current design on q35, the NIC should be placed on the exact same position after re-adding it so systemd should re-assign the previous `enoX` name. It might be worth clarifying that here. Or did I get you wrong here?

saschalucas:
Impressive work @rbott. Here are my fist quick comments:

Inline review of `doc/design-q35.rst` in github is not possible (300+ lines are to large for the web version). So I try to comment here:

> general things

Should we consider other changes in q35 introduction like `-nodefaults`

> no IDE

It seems common supporting SATA for legacy / Windows guests. I've also seen ISOs generated by [rear](https://relax-and-recover.org/) not booting from paravirtual. At least ICH9 seems to have a 6 port SATA controller by default. Are there any plans to use them?

> Chipset resolution

It's difficult to understand why it is necessary to store pc or q35 in the runtime besides already stored `machine_version`? Can you give an example?

> instance with many leaf devices can exhaust the available root-port slots faster than on i440fx

Does this hold true for Ganeti's limit of 24 leaf devices?

rbott:
Hi @saschalucas, thanks for your quick feedback! Reading the design doc once again myself I noticed some redundant and/or incomplete information. I have updated it accordingly.

> Should we consider other changes in q35 introduction like `-nodefaults`

I'll look into that. 

> 
> > no IDE
> 
> It seems common supporting SATA for legacy / Windows guests. I've also seen ISOs generated by [rear](https://relax-and-recover.org/) not booting from paravirtual. At least ICH9 seems to have a 6 port SATA controller by default. Are there any plans to use them?

I was not aware that booting from paravirtual might be an issue. I will put that on my manual-testing list :-) 

> 
> > Chipset resolution
> 
> It's difficult to understand why it is necessary to store pc or q35 in the runtime besides already stored `machine_version`? Can you give an example?

The section was slightly misphrased. The chipset resolver emits either `q35` or `i440fx` (to stay independent of the `pc` identifier which might change its meaning in the future). However, upon reading this and re-visiting the code I see that this actually causes more problems than it solves. I will drop the resolver from both the design doc and the current implementation. That will clean up the code a lot and we can deal with the problem _if_ Qemu really decides to change the meaning of "pc".

> 
> > instance with many leaf devices can exhaust the available root-port slots faster than on i440fx
> 
> Does this hold true for Ganeti's limit of 24 leaf devices?

What do you mean by that?

rbott:
So I _finally_ have a passing QA with `machine_version=q35`. I will now test again with `machine_version=""` to spot  regressions.

Testing took a while because random hot-add/-remove tasks would fail over the course of the 4-hour KVM QA run and it finally turned out to be the kernel+initrd combination I was using for the nested guests. They did not have proper PCIe hotplug support and for some reason it worked _most of the time_ but not everytime.

Instead of building the initrd myself I now switched to using dracut and that seems to be better at building working initrds :slightly_smiling_face: 

But that means we are now at a stage where it makes sense to also enter manual testing as the biggest blunders have been cleared :-)

lib/hypervisor/hv_kvm/monitor.py
```diff
@@ -609,13 +644,109 @@ def ResizeBlockDevice(self, disk_id: str, new_size: int):
...
+        if d.get("qdev_id") == devid:
+          return d
+        bridge = d.get("pci_bridge")
+        if bridge:
+          found = _walk(bridge.get("devices", []))
+          if found is not None:
+            return found
+      return None
+
+    return _walk(self._GetPCIDevices())
```
codefritzel, line 675: I would name the variable `devices` to use descriptive variable names.When using .get and the default values, you can omit one if statement:
```suggestion
    def _walk(devices):
      for device in devices:
        if device.get("qdev_id") == devid:
          return device

        found = walk(device.get("pci_bridge", {}).get("devices". []))
        if found is not None:
          return found
          
      return None

    return _walk(self._GetPCIDevices())
```

lib/hypervisor/hv_kvm/bus_manager.py
```diff
@@ -132,6 +148,74 @@ def initialize_from_device_info(self, device_infos: List[Dict]):
...
 
+class PCIeAllocator(BusAllocator):
+  """Allocator for q35's pre-allocated pcie-root-port pool.
+
+  Tracks which pool slots (L{Q35_RP_POOL_SLOTS}) are occupied. Leaf
+  devices attach as C{bus=rp<slot>,addr=0x0}. Uses C{bus_type="pci"} so
+  L{BusAllocatorManager} dispatches NICs and virtio disks unchanged.
+  """
+
+  BUS_TYPE = "pci"
```
codefritzel: ?
```suggestion
  BUS_TYPE = "pcie"
```

lib/hypervisor/hv_kvm/validation.py
```diff
@@ -55,6 +55,103 @@
...
+  Substring check on the user-supplied C{machine_version}; mirrors how
+  the rest of Ganeti treats this hvparam as a literal token (see
+  L{AssessParameters}). Versioned names (C{pc-q35-X.Y}) and the bare
+  C{q35} alias both match; C{pc-i440fx-X.Y} and the bare C{pc} alias
+  both do not.
+  """
+  return "q35" in (machine_version or "")
+
+
+def validate_q35_capability(hvparams, kvm_machine_output):
```
codefritzel, line 86: Please add type hints here

codefritzel (commented):
Hi @rbott,
It looks like a really good implementation so far. I've tried to do it too. But I've gotten completely lost in the rabbit hole of PCIe. :)
Originally, I wanted to treat everything in my data model as switches, multifunctional devices, and bridges, ...

I have suggestions to the code. As always, the decision is up to you.

rbott:
Thanks @codefritzel for your review and comments! For now I was mostly focussed on the design and a working prototype. I will now enter the next stage and do more manual testing and clean up the code (thanks for your hints, I am sure there's more 😀).

I also wanted to go down the PCIe switch route first, but some googling and "clauding" brought up the path with the multifunction device. And yes, it is a giant rabbit hole indeed 😂

I also managed to have a working KVM QA now (for both machine versions) but somehow borked all other types of QA runs. Once that is fixed, I'll continue with this PR.

rbott:
I wonder what to do with the `kvm_pci_reservations` setting. Right now, this has no effect on `q35`, just on `pc` machines. Instead of replicating the `pc`/PCI handling I'd pretty much like to stick with the multifunction device solution for the "built-in" devices (e.g. ballooning, sound, SCSI controller and so on). We could also address @saschalucas concern and statically add a S-ATA Controller as a second PCIe anchor/multifunction device so that we have even more possible slots for built-in devices (and just note in the code that there are more reserved slots for future use).

I would then just mark the setting as "pc-only" in the docs. What do you think?

//edit:
- I can't boot instances from `paravirtual`, but `scsi-cd` works fine
- I do not need to add a S-ATA controller, the PCI device is auto-created by qemu (and it was already in my overview in the design-doc :facepalm:)
- I'll see if I can make it usable as a CD option (for regular disks that would limit the total amount of disks to 6 minus any connected CD-ROM drives; that clashes with VirtIO/SCSI options which allow for 16).

rbott:
So here's a recap of what happend since my last comment:

# Getting rid of paravirtual CD types / freeing up PCIe slots

Following up on @saschalucas hint that paravirtual can not be used for booting I decided to drop that entirely from q35. There is not much to gain here because the performance win of virtio over IDE/SCSI/S-ATA CD ROM emulation is neglible. I am quite sure cd rom drives are mostly used for bootstrapping purposes (if at all). In short:

- `paravirtual` is not a supported value for `cdrom_disk_type` on q35 anymore
- I re-introduced `ide` but it uses S-ATA/AHCI internally on q35
  - interestingly enough, the 6-Port S-ATA controller of the q35 provides 6 "fake" IDE channels `ide.0...ide.5` and there's not much to change here
- as before, there is no legacy IDE available
- `scsi-cd` still works as well
- both `scsi-cd` and `ide` modes work fine for booting with SeaBIOS
- by default, instance **disks** are set to `paravirtual` and `cdrom_disk_type` inherits from that. To not leave users with unbootable instances, Ganeti will silently set `cdrom_disk_type` to `ide` if running on q35

There is one very important aspect of this change: the two (optional) VirtIO cdrom drives blocked two entire PCIe slots and we are short on PCIe slots. With the integrated q35 devices, VGA, 8 NIC + 16 disk slots and other optional periphery (see the PCIe bus layout in the design doc) there was not a single PCIe slot left.
Dropping the two `paravirtual` CD drives gives us back two PCIe multifunction slots!

# Adding virtio-rng

I have used one of the two empty slots to introduce a new PCIe device: [virtio-rng](https://wiki.qemu.org/Features/VirtIORNG) :-)

The other slot will be avilable for future extensions. I have also checked other means of extending the PCIe bus but none of them come for free:
- adding PCIe switches adds a whole new level of complexity
- adding a second PCIe bus is an option, but might confuse guests (as that seems to be a feature usually associated with NUMA systems)
- cutting down on maximum NICs (8) or DISKs (16) breaks compatibility with pc/i400fx VMs

# Fixing ACPI-Index / Predictive Interface Names and Ordering of Devices

There were several issues here. After some testing I figured out that systemd was naming the interfaces `enp...` instead of `enoX` because I had the `acpi-index` attached to the wrong PCIe device (it was tied to the pcie root port, but of course it needs to be attached to the actual device port one layer down). That is now fixed.

I also added static PCIe slot ranges for NICs and disk; previously they shared a single range. That lead to the situation that on cold boot you would see all NICs followed by all disks on the PCIe bus. Hot-Adding disks or NICs would place them on the PCIe bus in the order they were added, but on the next cold boot they were ordered by device type again. This is more of a cosmetic issue because the now properly placed `acpi-index` (see first paragraph) takes care of proper naming of the NICs.

With one exception: Having an instance with NICs `eno1`, `eno2` and `eno3` and then hot-removing `eno2` you will be left with `eno1` and `eno3`. However, after the next instance cold-boot, you will end up with `eno1` and `eno2`. To cover _that_ corner case as well, we'd need to store the NIC/PCIe-Slot-Assignments in the configuration database and I am not sure if that is worth the effort (at least in this preliminary stage).

# QA

The QA suite now toggles through `machine_version` values of `["", "pc", "q35"]` (testing both explicit and implicit set machine versions). This will again extend the runtime of the KVM QA by quite a bit, but with recent optimizations that should be fine.

So far I think we have reached a point where the features / goals are all layed out and it is now down to fixing / cleaning code & documentation :-)

saschalucas:
Thanks @rbott for your extensive work. Without looking at the code I have the following comments:

# PCI io port range
Looking at dmesg of an booted Trixie instance shows that on certain root-ports, no bridge window (io port range) can be assigned:

```
[Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: can't assign; no space
[Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: failed to assign
```

Maybe  [qemu:docs/pcie.txt](https://github.com/qemu/qemu/blob/master/docs/pcie.txt) section "3. IO space issues" and "7. Virtio devices" explain why. At least virtio devices should use disable-legacy=on, freeing IO-Ports from cold-plugged devices?

# pci bus tree
USB UHCI/EHCI: seems not present on a test instance? Maybe implicit removed by qemu when XHCI is used? Do we add always XHCI? usb_mouse can be tablet, mouse but not none.

# virtio CDROMs not bootable
Maybe the wording can be softened to "certain ISOs are not bootable via virtio". I just remember that some are working, while others don't. Maybe the working ones are ISOs that can be copied to USB storage as well and traditional BIOS El‑Torito works only with non virtio?

# no IDE disk
Once secure boot and TPM exists, we should figure out how to install Windows in a straight forward way. Former with IDE disks, no additional drivers were needed. Maybe it will work well, supplying the [virtio-win.iso](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.285-1/virtio-win.iso) in the second CDROM?

# spice
Ganeti's spice options may be incomplete. I.e. the USB-Passthrough or Drive-sharing are missing. But I don't know if this would consume additional PCIe slots?

# hotplug pools
You already mentioned grouping devices under functions later. Is this still the goal? Otherwise having 8 functions per root-port seems to fit well in the limits of 8 NICs and 2x8 disks. Probably also avoiding IO port issues?

# enoN vs. enpNs0
Is the understanding correct, that the advantage of using acpi-index will "only" be saving the suffix "s0", swap "p" to "o" and N could be a double-digit number?

# Unstable NIC names after hot remove and cold boot
... are fine, works like before. Ganeti traditionally does not save pci slot occupancy, also saving resulting conflicts / -resolution when changing the design later.

# virtio-rng
Isn't this obsolete since CPUs support RDRAND/RDSEED? In all my qemu-cpu-types I have this flags available inside the instances?

rbott:
Thanks for your feedback @saschalucas!

> # PCI io port range
> 
> Looking at dmesg of an booted Trixie instance shows that on certain root-ports, no bridge window (io port range) can be assigned:
> 
> ```
> [Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: can't assign; no space
> [Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: failed to assign
> ```
>
> Maybe [qemu:docs/pcie.txt](https://github.com/qemu/qemu/blob/master/docs/pcie.txt) section "3. IO space issues" and "7. Virtio devices" explain why. At least virtio devices should use disable-legacy=on, freeing IO-Ports from cold-plugged devices?

I have not observed these error messages, but maybe I simply overlooked them. I will dig into that, thanks for the hint!

> # pci bus tree
> 
> USB UHCI/EHCI: seems not present on a test instance? Maybe implicit removed by qemu when XHCI is used? Do we add always XHCI? usb_mouse can be tablet, mouse but not none.

I'll debug that as well. I remember seeing the devices, but I have not tested with different pointer-related settings. Currently XHCI is used whenever one of the USB devices is configured (mouse or tablet). XHCI will not be added, when no pointer device is configured.

> # virtio CDROMs not bootable
> 
> Maybe the wording can be softened to "certain ISOs are not bootable via virtio". I just remember that some are working, while others don't. Maybe the working ones are ISOs that can be copied to USB storage as well and traditional BIOS El‑Torito works only with non virtio?

In any case, IDE (S-ATA) or SCSI seem to work in all cases while virtio does not and it also blocks PCIe ports. I am not sure if the (seemingly) small performance gain of using virtio is worth spending more time on virtio here :-)

> # no IDE disk
> 
> Once secure boot and TPM exists, we should figure out how to install Windows in a straight forward way. Former with IDE disks, no additional drivers were needed. Maybe it will work well, supplying the [virtio-win.iso](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.285-1/virtio-win.iso) in the second CDROM?

We could still introduce IDE/S-ATA _disks_ at a later stage I guess (if users demand it). The only drawback will be that there's no hotplugging and the amount of virtual disks will be limited to four (as there are four remaining S-ATA channels).

> # spice
> 
> Ganeti's spice options may be incomplete. I.e. the USB-Passthrough or Drive-sharing are missing. But I don't know if this would consume additional PCIe slots?

I focussed on maintaining the current features - introducing new spice features would be out of scope for this PR I think.

> # hotplug pools
> 
> You already mentioned grouping devices under functions later. Is this still the goal? Otherwise having 8 functions per root-port seems to fit well in the limits of 8 NICs and 2x8 disks. Probably also avoiding IO port issues?

Unfortunatly multifunction devices are incompatible with hotplugging. You can not remove (or add) a single function from a multifunction port/device - only remove or add the entire device at once. So we won't be able to gain 

> # enoN vs. enpNs0
> 
> Is the understanding correct, that the advantage of using acpi-index will "only" be saving the suffix "s0", swap "p" to "o" and N could be a double-digit number?

It's not just that: even if we change the PCIe topology at a later point (e.g. introducing a PCIe switch to extend the number of ports) the NICs would maintain their stable name if we move the `acpi-index` with them to the new location in the PCIe topology.
Meanwhile, interface names starting with `enp...` reflect the absolute "position" on the PCIe bus / in the PCIe topology (also including function numbers if used) and hence they will change if we would re-shape the PCIe topology in the future.

> # Unstable NIC names after hot remove and cold boot
> 
> ... are fine, works like before. Ganeti traditionally does not save pci slot occupancy, also saving resulting conflicts / -resolution when changing the design later.

Agreed :-) 

> # virtio-rng
> 
> Isn't this obsolete since CPUs support RDRAND/RDSEED? In all my qemu-cpu-types I have this flags available inside the instances?

Actually I think I'll remove it again. a) because of RDRAND/RDSEED, b) because it occupies a precious PCIe slot (or rather: a function), and c) - which is the most important bit here - it has nothing to do with introducing Q35 support. We can have a separate PR/discussion for this feature :-)

saschalucas:
> > # hotplug pools
> > You already mentioned grouping devices under functions later. Is this still the goal? Otherwise having 8 functions per root-port seems to fit well in the limits of 8 NICs and 2x8 disks. Probably also avoiding IO port issues?
> 
> Unfortunatly multifunction devices are incompatible with hotplugging. You can not remove (or add) a single function from a multifunction port/device - only remove or add the entire device at once. So we won't be able to gain

I see. PCIe functions are not an option for the virtio-pools.

> > # enoN vs. enpNs0
> > Is the understanding correct, that the advantage of using acpi-index will "only" be saving the suffix "s0", swap "p" to "o" and N could be a double-digit number?
> 
> It's not just that: even if we change the PCIe topology at a later point (e.g. introducing a PCIe switch to extend the number of ports) the NICs would maintain their stable name if we move the `acpi-index` with them to the new location in the PCIe topology. Meanwhile, interface names starting with `enp...` reflect the absolute "position" on the PCIe bus / in the PCIe topology (also including function numbers if used) and hence they will change if we would re-shape the PCIe topology in the future.

Thanks, makes perfectly sense.

Overall, the design looks very good to me!

doc/admin.rst
```diff
@@ -1616,6 +1616,88 @@ NIC types like ``e1000``. Please check the :manpage:`gnt-instance(8)` man page f
...
+chipset for new x86 guests and it enables predictable
+``eno<N>`` interface naming in systemd guests, which makes it much easier
+to write stable in-guest network configuration.
+
+Predictable interface naming
+............................
+
+On q35, Ganeti tags each NIC's PCIe root port with an ACPI ``acpi-index``
+attribute. systemd-udev uses that attribute via the ``net_id`` builtin to
+assign an ``eno<N>`` name. The mapping is:
```
saschalucas, line 1638: This must be updated? The acpi index is on the device itself not the root-Port?

rbott: This is gone now, but for some reason Github does not tag it as "outdated"

rbott:
# PCI io port range
> > Looking at dmesg of an booted Trixie instance shows that on certain root-ports, no bridge window (io port range) can be assigned:
> > ```
> > [Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: can't assign; no space
> > [Tue May 19 08:14:14 2026] pci 0000:00:11.0: bridge window [io  size 0x1000]: failed to assign
> > ```

> >   
> > Maybe [qemu:docs/pcie.txt](https://github.com/qemu/qemu/blob/master/docs/pcie.txt) section "3. IO space issues" and "7. Virtio devices" explain why. At least virtio devices should use disable-legacy=on, freeing IO-Ports from cold-plugged devices?
> 
> I have not observed these error messages, but maybe I simply overlooked them. I will dig into that, thanks for the hint!

Well yes, I also see them. The explanation in the pcie.txt is also sound, but I can't get it fixed. It seems to be that SeaBIOS already does the right thing, but Linux goes its own way and ignores hints (like `io-reserve=0`). FWIW, the (seemingly cosmetic) problem goes away if Linux gets booted with `pci=hpiosize=0`.

This problem should in theory only affect hot-plugged PCI devices (not PCIe!). Since the 16 disks are all virtio-blk-pci (native PCIe) and the same also applies to paravirtual NICs the only affected devices would be emulated NICs (e1000, ne2k_pci etc.). I have tested hot-plugging up to 8 NICs of different types into instances booted with and without the aforementioned kernel parameter and it always worked.

So I *think* this is really just a cosmetic issue (although an anoying one :disappointed: ).

# pci bus tree
> > USB UHCI/EHCI: seems not present on a test instance? Maybe implicit removed by qemu when XHCI is used? Do we add always XHCI? usb_mouse can be tablet, mouse but not none.
> 
> I'll debug that as well. I remember seeing the devices, but I have not tested with different pointer-related settings. Currently XHCI is used whenever one of the USB devices is configured (mouse or tablet). XHCI will not be added, when no pointer device is configured.
> 

Apparently I have been fooled by (old) documentation, LLM hallucinations and poor validation skills: some of the devices that were originally in the PCIe topology in the design document are not really created by qemu. There is no UHCI/EHCI USB controller by default which means, more free PCIe slots for us :-) I have consolidated the existing slot ranges for NICs/disks to be in a continous space. Thanks for spotting this!

Also: I fixed that the XHCI controller will also be added if `usb_devices` are configured (but have not tested it yet).

saschalucas:
Thanks @rbott. This PR looks good to me for removing Draft and maybe also for merging. There are 2 pending comments by @codefritzel. Can you please check?

rbott:
> Thanks @rbott. This PR looks good to me for removing Draft and maybe also for merging. There are 2 pending comments by @codefritzel. Can you please check?

I was a bit busy in the last two weeks and I also need to fix the QA suite. I changed the initramfs system for the nested guests to properly support PCIe hotplug/hot-remove but unfortunately broke all kinds of other QA types along the way (QEMU shutdown signals for xen guests, hot-removal of devices on i440fx etc.) and have been moving back and forth between different problems :-)

As soon as the QA system is stable again I will pick up where I left with this PR (focus on code quality and then declare it ready for review).

rbott:
@saschalucas I did some cleanups and the PR is now ready for final review. QA passes cleanly for Debian Trixie and sometimes fails for Debian Forky (due to `sharedfile`/NFS quirks), but the latter also happens on the master branch and is not related to changes on this branch.

Regarding type hints (cc @codefritzel): The dev guidelines (`doc/dev-codestyle.rst`) currently mandate epydoc / docstrings. To keep things in line with the current reality I resorted to use whatever the current file already uses (e.g. use docstrings in a file that already uses them and use type hints in a file that does not use docstrings). That's at least consistent until we update the dev guidelines with a new policy.

saschalucas (approved):
Thanks @rbott. Well done!

PR #1941, opened 2026-06-03
rbd: add namespace parameter

### LINKED ISSUE #1936: rbd: support namespace
Since having rbd user support (#1680) it makes also sens to have rbd namespace support. I.e. when multiple Ganeti Clusters share a single ceph cluster, having different rbd users and hence different namespaces, where the actual user permission is granted only inside the namespace. So that different Ganeti clusters can not interfere in rbd image creation/deletion.

incomplete work at https://github.com/ganeti/ganeti/compare/master...saschalucas:ganeti:rbd_namespace

saschalucas:
To separate a single RBD Ceph pool for multiple consumers (i.e. multiple Ganeti clusters) RBD namespaces must be used. With that every namespace has its own space for RBD images and in combination with RBD `user-id` it's own permissions only inside this namespace.

### DIFF
--- a/lib/tools/cfgupgrade.py
+++ b/lib/tools/cfgupgrade.py
@@ -762,12 +762,12 @@ class CfgUpgrade(object):
       if variant in hvparams:
         hvparams[variant]["xen_cmd"] = "xl"
 
-  @OrFail("Removing the rbd/user-id parameter")
-  def DowngradeRbdUserId(self):
-    """Remove rbd/user-id parameters if set
+  @OrFail("Removing the rbd/user-id and namespace parameter")
+  def DowngradeRbdSettings(self):
+    """Remove rbd/user-id and namespace parameters if set
 
     """
-    def _removeRbdUserId(data):
+    def _removeRbdSetting(data, setting):
       diskparams = data.get("diskparams", None)
       if diskparams is None:
         return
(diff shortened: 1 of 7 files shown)


rbott:
Hi @saschalucas and thanks for your contribution; that looks like a useful addition :-) I have started a QA run (visible on https://stargazer.ganeti.org/ with tag `composed-maggie`).

While there is support for cycling through various hvparam values in the QA, there is currently no support for cycling through disk param values. This might be worth adding to the QA suite (but not as part of this PR) so that we automatically test with and without e.g. namespaces set.

lib/tools/cfgupgrade.py
rbott, line 1: When releasing 3.1 we should probably add a note that downgrading Ganeti *will* remove the namespace configurations (although it is kinda obvious). But I guess if you upgrade, create a new instance with a namespace configured and then downgrade you are in for a surprise :-)

rbott:
After adding my downgrade-comment another thing came into my mind: what happens when I upgrade an existing cluster (not using namespaces), then configure a cluster-wide namespace? Would that mean Ganeti will not be able to find the disks/volumes of existing instances any more?

saschalucas:
Setting/changing a namespace is like setting/changing the rbd pool: Ganeti will no longer see any RBD images. So this is only for new clusters or careful prepared migrations (i.e. shutting down all instances, clone all images into the new namespace, setting the new namespace inside Ganeti, start all instances).

On downgrade the same migration to the default RBD namespace applies.

Should be noted in the NEWS.

rbott:
I think we should even add a paragraph to the man page(s) that setting/chaning this parameter on an existing cluster may cause harm because it affects all existing instances. What do you think?

saschalucas:
> I think we should even add a paragraph to the man page(s) that setting/chaning this parameter on an existing cluster may cause harm because it affects all existing instances. What do you think?

I have updated the man page. PTAL.

rbott (approved):
LGTM, thanks @saschalucas!

PR #1946, opened 2026-06-09
rbd: add disk info

### LINKED ISSUE #1554: rbd/ceph: add metadata to disk images
rbd disks can be enriched by metadata in the form of key=value (see `rbd(8)` `image-meta set` etc.). Like with LVM tags one could add originstname=inst1.domain.tld. This makes association of instances and their disks possible for external tools (without looking up ganeti config data).

saschalucas:
Analogous to DRBD/LVM disks add information of the originating instance's name to RBD disks. This way it is possible to identify RBD images belonging to an instance without consulting Ganeti's config (i.e. for disaster recovery).

closes #1554.

### DIFF
--- a/lib/storage/bdev.py
+++ b/lib/storage/bdev.py
@@ -1319,6 +1319,23 @@ class RADOSBlockDevice(base.BlockDev):
       base.ThrowError("Hypervisor %s doesn't support RBD userspace access" %
                       hypervisor)
 
+  def SetInfo(self, text):
+    """Update metadata with info text.
+
+    """
+    # Replace invalid characters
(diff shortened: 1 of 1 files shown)


rbott (approved):
LGTM, thanks!

The only thing to consider: is there a maximum length on metadata values? I could not find anything in documention and did not dive into ceph/rbd's codebase to figure out more.

But I guess thats negligible...?

saschalucas:
> The only thing to consider: is there a maximum length on metadata values? I could not find anything in documention and did not dive into ceph/rbd's codebase to figure out more.

Also didn't find any limits, that's why I decided not to cap at 128 chars like LVM tags.
 
> But I guess thats negligible...?

Let's find out :-).

PR #1948, opened 2026-06-19
KVM: tolerate transient QMP socket errors when reading instance info

rbott:
During parallel instance creation/removal (QA test `TestParallelMaxInstanceCreationPerformance`) an instance may be queried while starting up or shutting down, when its QMP socket does not yet/no longer exist. `socket.connect()` then raises `FileNotFoundError`, which `GetInstanceInfo` did not catch (it is not a `ConnectionError`). This failed the whole `all_instances_info` RPC and made the master's IAllocator raise `OpExecError`, killing unrelated concurrent create jobs.

Catch `OSError` in `GetInstanceInfo` (covering `FileNotFoundError`, `BrokenPipeError` and `ConnectionError`) and fall back to the pid-derived values. Also skip an instance raising `OSError` in `GetAllInstancesInfo` so one instance in flux cannot abort info gathering for the rest.

This error surfaces every now and then in QA runs and has seemingly been introduced with the switch from HMP to QMP. It will probably not be relevant to regular Ganeti use.

Note: I have used AI to analyze various failing QA runs and the codebase to pinpoint the error.

### DIFF
--- a/lib/hypervisor/hv_kvm/__init__.py
+++ b/lib/hypervisor/hv_kvm/__init__.py
@@ -1030,14 +1030,12 @@ class KVMHypervisor(hv_base.BaseHypervisor):
       qmp.close()
     except errors.HypervisorError:
       pass
-    # during instance shutdown it may happen, that qemu has already exited
-    # in the middle of talking to it via QMP
-    except BrokenPipeError:
-      logging.debug("Error: Broken pipe. The QMP socket %s has shut down." %
-                    socket_path)
-    except ConnectionError:
-      logging.debug("Error: Unable to connect to the QMP socket %s." %
-                    socket_path)
+    # QMP may be transiently unavailable while the instance is starting up or
+    # shutting down (socket missing, refused, or broken pipe). Harmless here:
+    # fall back to the pid-derived values above.
+    except OSError as err:
+      logging.debug(f"Ignoring transient QMP error on socket {socket_path}:"
+                    f" {err}")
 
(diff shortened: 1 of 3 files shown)


lib/hypervisor/hv_kvm/__init__.py
```diff
@@ -1030,14 +1030,12 @@ def GetInstanceInfo(self, instance_name, hvparams=None):
...
-    except BrokenPipeError:
-      logging.debug("Error: Broken pipe. The QMP socket %s has shut down." %
-                    socket_path)
-    except ConnectionError:
-      logging.debug("Error: Unable to connect to the QMP socket %s." %
-                    socket_path)
+    # QMP may be transiently unavailable while the instance is starting up or
+    # shutting down (socket missing, refused, or broken pipe). Harmless here:
+    # fall back to the pid-derived values above.
+    except OSError as err:
```
saschalucas, line 1036: I assume OSError also covers BrokenPipeError and ConnectionError as before?

saschalucas (approved):
Thanks @rbott. Just a minor query regarding `OSError` (see comment). Otherwise please merge your changes.

lib/hypervisor/hv_kvm/__init__.py
saschalucas: Sorry, I read your PR description before, but understand it fully just now.

rbott:
Thanks for your review :-)

PR #1949, opened 2026-06-25
pvh_xen: Added hypervisor parameter to run Xen domains in PVH mode.

wesleyCampbell:
New hypervisor parameter, `pvh_mode`, that allows configuring Xen domains to run in PVH mode instead of PV mode.

When `pvh_mode=1`, the domain will run in legacy PVH mode by writing 'pvh = 1' in the Xen domain configuration file.

When `pvh_mode=2`, the domain will run in modern PVH mode by writing "type = 'pvh'" in the Xen domain configuration file.

Based on the patche files found at https://gist.github.com/flumeware/7bfa3e270712ba5bac4ccdaa815c092c.

### DIFF
--- a/lib/hypervisor/hv_xen.py
+++ b/lib/hypervisor/hv_xen.py
@@ -1486,10 +1486,8 @@ class XenPvmHypervisor(XenHypervisor):
     config.write("name = '%s'\n" % instance.name)
 
     pvh_mode = hvp[constants.HV_PVH_MODE]
-    if (pvh_mode == "1"):
-        config.write("pvh = 1\n")
-    elif (pvh_mode == "2"):
-        config.write("type = 'pvh'\n")
+    if (pvh_mode is True):
+        config.write("type = 'pvh'\n)
 
     self._WriteNicConfig(config, instance, hvp)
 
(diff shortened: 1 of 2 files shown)


rbott:
Hi @wesleyCampbell,

thanks for your PR! I am not a Xen expert but from what I get is that Ganeti currently creates configuration the "old" way using the legacy `builder` option (at least for `hvm`, as `generic` aka PV is the implicit default). I also understand that this should be removed in favor of the newer `type` option which holds either `pv`, `pvh` or `hvm`.

But since both existing types (pv/hvm) are currently split across two different hypervisor implementations (xen-pvm vs. xen-hvm) and `pvh` seems to be a sort of hybrid, I am not even sure where this would go or if we _actually_ need a new hypervisor implementation called xen-pvh.

What do you think?

wesleyCampbell:
Hi @rbott,

thanks for the reply. From our tests running the `pvh` code on our machines, it appears that Xen's `pvh` mode works just fine on the xen-pvm hypervisor, so it would not be necessary to create a xen-pvh hypervisor. It appears that setting `type = pvh` and pointing the domain to the correct `xen-pvh` grub path is sufficient for the machines to run. This has proved very helpful for us as it allows our Xen domains to run modern zstd-compressed kernels (such as Debian 13) without any modifications.

wesleyCampbell:
Hi, @rbott.

Given our experience running the patch on our machines, we have decided to change the PVH mode hypervisor parameter into a boolean. We found this approach much simpler and works quite well for enabling PVH mode on Xen guests. We have pushed these changes in our most recent commit.

We are also seeking further feedback on what we can do to get this patch officially accepted by the Ganeti team. Are there further changes that you would suggest?

Thank you for your consideration.

PR #1951, opened 2026-07-14
Vcluster iproute2 support

cpasqualini:
While building the Debian package, autopkgtest failed because vcluster-setup invokes the route command.

The route and ifconfig commands are provided by the legacy net-tools package, which is no longer installed by default on recent Linux distributions. Ganeti already depends on iproute2, which provides the modern ip command.

This patch replaces the remaining net-tools commands with their ip equivalents. The changes are limited to two occurrences: one in vcluster-setup.in and one in vif-ganeti-metad.

Signed-off-by: Carlos R. Pasqualini carlos@carlospasqualini.com.ar

### DIFF
--- a/tools/vcluster-setup.in
+++ b/tools/vcluster-setup.in
@@ -269,7 +269,7 @@ setup_network_interfaces() {
     ip addr add "$ipaddr" dev "$netdev"
   done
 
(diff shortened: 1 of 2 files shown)


saschalucas (approved):
Thanks @cpasqualini for caring about the Debian package.

Surprisingly there are still parts in the code not converted to iproute2.

PR #1953, opened 2026-07-21
introduce IOThreads support

saschalucas:
Qemu by default emulates devices in vCPU threads or the main event loop. This can become a scalability bottleneck, where vCPU time competes with I/O, slowing each other. IOThreads allow device emulation and I/O to run on other host CPUs.

IOThreads only work with paravirtual disks (virtio-blk, virtio-scsi) and *not* with `aio=threads`. Ganeti silently adjusts the number of IOThreads to a meaningful value. It disables IOThreads for unsupported `disk_type` / `scsi_controller_type` and unsupported `aio`. It also checks Qemu version for supporting mapping between IOThreads and virtqueues. If virtqueue-mapping is supported, values greater than `1` are possible (Qemu-9.0 for virtio-blk and 10.0 for virtio-scsi). In this case IOThreads are assigned round robin to every instance's disks.

The default is to set to `1` IOThread and should be enough for most workloads. Usually 4-8 IOThreads are sufficient to saturate disks (beyond 100k, up to 500k IOPS). Other than with `virtio_net_queues` it does not make sense to have an `auto` mode that follows the instance vCPU count. However an appropriate number of vCPUs is necessary to exploit multiple I/O threads. Adding IOThreads beyond saturation does not increase performance and may harm it.

As usual, IOThreads can not be changed during live migration. When downgrading Ganeti to a version not supporting IOThreads, instance reboots are necessary.

https://developers.redhat.com/articles/2024/09/05/scaling-virtio-blk-disk-io-iothread-virtqueue-mapping

Assisted-by: Gemini (Google AI) <gemini@google.com>


saschalucas:
currently in draft, needs more testing.

saschalucas:
basic testing was successful (live migration, hotplug, upgrades) @rbott can you please run QA?

rbott:
QA run started, i'll update here once I have results

*Update*: QA passend on [Debian Trixie](https://stargazer.ganeti.org/df90980e40ea235eb9615381d749e09e21824c84/) with the default `virtio_disk_iothreads` value of `1`

The QA suite knows how to cycle through hvparam values. Would it make sense to run it with the default of `1` and have it run tests with any other value? If that does make sense, you can add this within this PR, simply extend this list with the parameter name and the values to cycle through:

https://github.com/ganeti/ganeti/blob/012bc6820e8448dc9a2a3432f67bdb2746a946f0/qa/qa_cluster.py#L158-L165

man/gnt-instance.rst
```diff
@@ -1030,6 +1030,31 @@ virtio\_net\_queues
...
     It is set to ``1`` by default.
 
+virtio\_disk\_iothreads
+    Valid for the KVM hypervisor.
+
+    Set a number of IOThreads shared among *all* instance's disks. Qemu
+    by default emulates devices in vCPU threads or the main event loop. This
+    can become a scalability bottleneck. IOThreads allow device emulation
+    and I/O to run on other host CPUs. This only works with paravirtual
+    disks (virtio-blk, virtio-scsi) and *not* with ``aio=threads`` The
```
rbott: There is a dot missing at the end of the sentence

man/gnt-instance.rst
```diff
@@ -1030,6 +1030,31 @@ virtio\_net\_queues
...
+
+    Ganeti silently adjusts the number of IOThreads to a meaningful
+    value. It disables IOThreads for unsupported ``disk_type`` /
+    ``scsi_controller_type`` and unsupported ``aio``. It also checks Qemu
+    version for supporting mapping between IOThreads and virtqueues. If
+    virtqueue-mapping is supported, values greater than ``1`` are possible
+    (Qemu-9.0 for virtio-blk and 10.0 for virtio-scsi). In this case
+    IOThreads are assigned round robin to every instance's disks.
+
+    The default is to set to ``1`` IOThread and should be enough for
```
rbott: Just to understand this correctly: IOThreads will now _always_ be enabled unless users have configured an incompatible `disk_type` or `aio` mode? Or is the setting of `1` equivalent to the current mode of operation?

man/gnt-instance.rst
```diff
@@ -1030,6 +1030,31 @@ virtio\_net\_queues
...
+    disks (virtio-blk, virtio-scsi) and *not* with ``aio=threads`` The
+    maximum number of IOThreads is limited to ``8``.
+
+    Ganeti silently adjusts the number of IOThreads to a meaningful
+    value. It disables IOThreads for unsupported ``disk_type`` /
+    ``scsi_controller_type`` and unsupported ``aio``. It also checks Qemu
+    version for supporting mapping between IOThreads and virtqueues. If
+    virtqueue-mapping is supported, values greater than ``1`` are possible
+    (Qemu-9.0 for virtio-blk and 10.0 for virtio-scsi). In this case
+    IOThreads are assigned round robin to every instance's disks.
```
rbott: Does an instance with a single disk also benefit from `virtio_disk_iothreads > 1`? Will that simply be a multi-queue device from the guest point of view? 

I was just wondering if we could also implement some kind of `auto` magic similiar to the NIC queues (which scales the number of iothreads with `len(disks)` or with `len(disks) * 2` or something similar.

man/gnt-instance.rst
```diff
@@ -1030,6 +1030,31 @@ virtio\_net\_queues
...
+    Valid for the KVM hypervisor.
+
+    Set a number of IOThreads shared among *all* instance's disks. Qemu
+    by default emulates devices in vCPU threads or the main event loop. This
+    can become a scalability bottleneck. IOThreads allow device emulation
+    and I/O to run on other host CPUs. This only works with paravirtual
+    disks (virtio-blk, virtio-scsi) and *not* with ``aio=threads`` The
+    maximum number of IOThreads is limited to ``8``.
+
+    Ganeti silently adjusts the number of IOThreads to a meaningful
```
rbott: Wait, I thought the number of threads will be configured through this HV param? :-)

saschalucas:
Thanks @rbott for taking review, I'll address your comments shortly.

While trying to explain iothreads vs. number of queues I discovered a flaw in the current implantation. Here is a table comparing current Ganeti behavior compared to this PR as of now and as intended:

| what | current Ganeti | this PR as of now | this PR as intended |
|---|---|---|---|
| number of iothreads | 0, device I/O is done in vCPU threads or the main event loop | 1 dedicated iothread by default, up to 8 configurable, but always only if virtio and not aio-threads  | same as of now |
| number of I/O queues | equal to the number of vCPUs, implicit by qemu | ⚡identical to number of iothreads | identical to number of vCPUs |

For better understanding: With current Ganeti every virtio-disk has as many queues as vCPUs (this is implicit done by qemu). All this queues are executed in a single shared qemu thread (vCPU/main) competing with other tasks.

So the biggest win and new default is to use one dedicated iothread to decouple from vCPU/main thread. But still every virtio-disk with its multiple queues is done by a single iothread.

Only for extreme cases, more than 1 iothread can be set. In this case iothreads must be mapped to queues. This is done by instructing qemu to distribute all iothreads over available queues, for every virtio-disk.

man/gnt-instance.rst
saschalucas: Silently adjusting means, if virtio_disk_iothreads is set, but does not meat the the requirements (virtio and not aio-threads), Ganeti will silently override user settings.

man/gnt-instance.rst
saschalucas: iothreads and multiqueue are two distinct properties. virtio disks already have multiple queues if the instances has 2 ore more vCPUs (see /sys/block/vda/mq/). iothreads represents the node side view, how qemu will do I/O: weather shared on vCPU/main-thread or on a dedicated iothread.

More iothreads won't help / do harm if disk is already saturated, so auto magic does not make sense. The biggest improvement is to decouple the I/O with 1 thread from vCPU/main-thread.

man/gnt-instance.rst
saschalucas: yes it is the intention to always enable 1 IOThread, it is not equal to current mode of operation.

saschalucas:
Reworked man page, hopefully leaving no questions open. Fixed logic with number of queues equal to vCPUs. Limit iothreads to <= vCPUs. Discovered during testing #1963.

regarding qa: don't know how much load more hv-param cycles will generate, but in theory one could test for `virtio_disk_iothreads = [0, 1, 2]`. Where 2 should have also at least 2 vCPUs.

This PR is ready for review/qa again. @rbott PTAL.

rbott:
Thanks for your additions / clarifications @saschalucas, sounds very good to me!

I have recently reworked the QA a bit so that parameter iterations only run a very limited number of tests (instance creation, reboot, destruction, hot-plugging/-removing etc) so that the impact of new iterations is not that big any more. I'd say let's add `virtio_disk_iothreads = [0, 1, 2]` to the tests and then I'll re-run the QA!

saschalucas:
> I have recently reworked the QA a bit so that parameter iterations only run a very limited number of tests (instance creation, reboot, destruction, hot-plugging/-removing etc) so that the impact of new iterations is not that big any more. I'd say let's add `virtio_disk_iothreads = [0, 1, 2]` to the tests and then I'll re-run the QA!

Thanks @rbott, I have added `virtio_disk_iothreads = [0, 1, 2]` to the QA. How many vCPUs do instances in the QA have? Regarding `virtio_disk_iothreads = 2` only makes sense when the instance has 2 or more.

qa/qa_cluster.py
```diff
@@ -162,6 +162,7 @@ def PrepareHvParameterSets():
       "disk_type": ["ide", "paravirtual"],
       "soundhw": ["ac97", "hda"],
       "machine_version": ["", "pc", "q35"],
+      "virtio_disk_iothreads" = [0, 1, 2],
```
rbott: That should be `"virtio_disk_iothreads": [0, 1, 2],`:

```
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 44, in <module>
    from qa import qa_cluster
  File "/usr/src/ganeti/qa/qa_cluster.py", line 165
    "virtio_disk_iothreads" = [0, 1, 2],
                          ^
SyntaxError: ':' expected after dictionary key
```

saschalucas: Oh, sorry. Fixed now.

rbott:
The [run](https://stargazer.ganeti.org/623c74a3207e09d1977f554dd6ef56cf6c4e5ae1/qa.log) failed at a different stage:

```
---- Starting reduced number of instance tests with hypervisor parameter virtio_disk_iothreads=0
Command: ssh maggie.staging.ganeti.org 'gnt-cluster modify -H kvm:virtio_disk_iothreads=0'
Stderr was:
Unhandled Ganeti error: Parameter 'virtio_disk_iothreads' fails syntax check: invalid number of iothreads (current value: '0')
---- BLOCK FAILED [RunInstanceTests] Create and exercise instances: Command 'gnt-cluster modify -H kvm:virtio_disk_iothreads=0' on node maggie.staging.ganeti.org failed, exit code 1
^^^^ BLOCK 2026-09-09 18:54:54.694197 time=0:46:05.652286 [RunInstanceTests] Create and exercise instances
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 1216, in <module>
    main()
    ~~~~^^
  File "/usr/src/ganeti/ganeti/rapi/client.py", line 274, in wrapper
    return fn(*args, **kwargs)
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 1211, in main
    RunQa()
    ~~~~~^^
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 1106, in RunQa
    RunTestBlock(RunInstanceTests)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 186, in RunTestBlock
    return fn(*args, **kwargs)
  File "/usr/src/ganeti/qa/ganeti-qa.py", line 880, in RunInstanceTests
    qa_cluster.AssertClusterHvParameterModify(param, value)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/usr/src/ganeti/qa/qa_cluster.py", line 214, in AssertClusterHvParameterModify
    AssertCommand(["gnt-cluster", "modify", "-H", "%s:%s=%s" %
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Before the modify operation, the value was at `virtio_disk_iothreads: 1`

saschalucas:
> Thanks for discovering my copy&paste error from virtio_net_queues validation and sorry for not testing myself.

no worries, that's what the QA suite is for :-)

> Should be fixed, PTAL.

I have started another run: https://stargazer.ganeti.org/fc46e3e60bb734456e42a14b9af657996187bc3e/

rbott:
oh wow, I really don't know how I ended up *editing* your post instead of quoting it @saschalucas. I am very sorry for that :confused: 

However, the QA run passed and I will take a deeper look into the code during the next days :-)

rbott (approved):
LGTM, please merge your changes. Thanks for implementing this @saschalucas 🙂

PR #1954, opened 2026-07-23
Add connected_node_groups to network API responses

codefritzel:
This change adds a new connected_node_groups field to the network API response for `/2/networks` and `/2/networks/<network_name>`.

The goal is to expose the network-to-nodegroup relation in a structured, JSON-friendly form instead of the current opaque group_list tuple representation. Each entry now contains:
- name
- mode
- link
- vlan

This makes the API much easier to consume by clients and tooling, and avoids forcing downstream users to decode a positional list format.

It also normalizes empty string values to `None`, so missing optional fields are represented cleanly in JSON.

This is an important API improvement because a well-designed endpoint should return data in a stable, self-describing, and machine-friendly structure. That makes integrations more robust and reduces the amount of client-side parsing logic needed.

See #1889

@saschalucas or @rbott please check if this is relevant to you; if not, please close the PR.

### DIFF
--- a/lib/rapi/rlib2.py
+++ b/lib/rapi/rlib2.py
@@ -215,6 +215,28 @@ def _UpdateBeparams(inst):
   return inst
 
 
+def _NormalizeConnectedNodeGroups(group_list):
+  """Normalizes the network's group_list into a JSON-friendly view.
+
+  @param group_list: list of tuples from the query layer
+  @type group_list: list
+  @return: list of dictionaries with name/mode/link/vlan
+
+  """
+  if not group_list:
+    return []
+
+  def _NormalizeValue(value):
+    return None if value == "" else value
+
+  return [{
+    "name": group[0],
+    "mode": _NormalizeValue(group[1]),
+    "link": _NormalizeValue(group[2]),
+    "vlan": _NormalizeValue(group[3]),
+  } for group in group_list]
+
+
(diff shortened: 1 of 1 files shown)


rbott:
Thanks for bringing this up @codefritzel. Unfortunately there are a lot of API design smells hidden in RAPI responses (I think by far the worst offender is the jobs resource).

I have started a quick QA run but I do not expect any side effects from looking at the code. I'l update this comment once the results are in :-)

*Update*: QA [looks good](https://stargazer.ganeti.org/31fb2ad848147a299c1834e2e713ac9ac8752851/), Thanks and please merge your changes :-)

rbott (approved):

codefritzel:
@rbott thanks for your review and your approve! :+1: 



> Unfortunately there are a lot of API design smells hidden in RAPI responses (I think by far the worst offender is the jobs resource).

Not only the jobs object, but also networks and disks are in an array structure that the client must interpret on its own. I found this when I was trying to create a Python-RAPI-Client and a Ganeti-Ansible-Collection.




> _Update_: QA [looks good](https://stargazer.ganeti.org/31fb2ad848147a299c1834e2e713ac9ac8752851/), Thanks and please merge your changes :-)

Please merge it by yourself. Since I left the project, I've gave up these permissions.

PR #1955, opened 2026-07-24
KVM: add `boot_type` hvparam and UEFI/OVMF boot support [and bump development cycle to 4.0]

rbott:
This PR implements UEFI/OVMF boot for the KVM hypervisor (works both i440fx and q35 instances). It contains quite some substantial changes so we also bump the Ganeti version to 4.0.0~alpha (following an internal discussion).

Here's a short summary; for the full picture please read the included design doc:

- replace the current boot type logic ("`kernel_path` set" -> direct kernel boot; "unset" -> BIOS boot) with a new HV param `boot_type` (values `direct_kernel`, `bios`, `uefi`) 
- cluster upgrade derives the `boot_type` from the deprecated logic based on `kernel_path`
- OVMF firmware and vars along are stored along with each instance and retain instance mobility/availability (live migration, failover, move/copy etc.); to achieve that we create an extra instance disk for `UEFI` instances so we can re-use the existing code paths for all mobility operations
- regular data disks are distinguishable from the firmware disk through a new database field `role` (per disk)
- firmware disks are exempt from ipolicy checks, spindle counts and certain operations (grow, delete-if-active etc.)
- firmware disks are created / initialized when an instance boots for the first time or `boot_type` is changed to `uefi` on an existing instance

## DRBD Caveat

One notable aspect with DRBD:
While the [official 8.4 documentation](https://linbit.com/drbd-user-guide/users-guide-drbd-8-4/#s-recent-changes-defaults-minor-count) states:

> The maximum number of configurable DRBD devices (previously 255) is 1,048,576 (220) in DRBD 8.4. This is more of a theoretical limit that is unlikely to be reached in production systems.

... `modinfo drbd` on Debian Forky states:

> version:        8.4.11
> [...]
> parm:           minor_count:Approximate number of drbd devices (1U-255U) (uint)

If there actually is a device limit of 255, adding an extra disk for each UEFI instance slashes the number of allocatable instances in half (e.g. 64 primary and 64 secondary instances). 

## Kernelspace Access Caveat

While Ganeti supports userspace access for certain disk templates (notably RBD and gluster) Ganeti will silently override it to kernelspace for the firmware disk (all data disks will keep the configured access type). This is required because a `pflash` device cannot be backed by userspace QEMU drivers and must either be a file or a blockdevice (see design doc for more details).

## Future Improvements

There are still some missing parts here:
- export/import of UEFI instances is not supported (neither local nor between clusters)
- OVMF maintenance helpers/commands are missing (e.g. detect an updated OVMF firmware on the node, update instances' OVMF firmware)
- secure boot or TPM is also not covered by this PR

The code has passed QA runs but still needs extensive testing (as part of this PR and also after release of the 4.0.0~alpha).

Disclaimer: AI has been used in the discovery / research phase for this PR and also partially for code generation.


rbott:
# What has been tested so far

On a single-node "cluster", I have successfully UEFI-booted:

- [GRML Linux](https://grml.org/) ad-hoc UEFI boot
- Debian Trixie (installed through the interactive installer from the ISO), including XFCE desktop environment
- Windows 8.1 with VirtIO Driver Disk (did not have anything newer available)

All of them worked fine with both `machine_version=pc` and `machine_version=q35`. Using `q35` and `soundhw=hda` I was also able to get sound through SPICE and `virt-viewer` / `remmina`. So far I have only tested the `file` disk template.

# Whats left to test

- build a multi node cluster and test other disk templates
- test live migration and failover on various disk templates
- TBD

Sidenote: While Microsoft seems to offer Windows versions for evaluation purposes through their [Evaluation Center](https://www.microsoft.com/en-us/evalcenter), they only offer the current version Windows 11. Since that requires secure boot and a TPM device, this is currently of no use. They do not seem to offer Windows 10 there any more (but then again; why should they in 2026?) and I do not have an installation media at hand.

If anybody wants to help out with testing, please feel free :smile:

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+instance's data disks** - mirrored ``DT_DRBD8`` on DRBD instances, and so
+on. Because it lives in ``inst.disks``, every movement and failure path
+that iterates ``cfg.GetInstanceDisks`` carries it automatically through
+already-battle-tested code. There is no bespoke movement code and
+therefore no "forgot a path" risk - which is exactly the surface where
+data loss would brick an instance.
+
+The firmware code is pinned per-instance at creation, so updating the
+node's OVMF package does **not** change a running instance's firmware
+(no silent change that could alter boot behaviour).
```
saschalucas, line 118: There are pros and cons whether to make a copy of the OVMF CODE or to use a shared instance from the nodes ovmf-package.

| scenario | shared | copy |
| ---- | ----| ---- |
| Host-OS / QEMU Upgrade | QEMU and OVMF matches / are tested together (is there a risk of altered boot behavior?) | old CODE may hit new QEMU (could possible not play together?) |
| Live-Migration during (security) Updates | care must be taken to keep Live-Migration stable (i.e use old 2M instead of 4M) | no CODE changes / stable |
| expense of maintenance (i.e. security) | for free | high (not implemented yet) |

Do we know, how do other projects (openstack, proxmox, incus) handle this?

I tend to believe it might be simpler to start the initial implementation with the `shared` variant?

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+headroom for future blobs, without oversizing the mirrored volume) laid
+out as:
+
+- a small self-describing **superblock** in sector 0 (magic string, version,
+  a region table of ``name → (offset, size)``);
+- the **code** region (read-only, seeded from ``OVMF_CODE_TEMPLATE``);
+- the **vars** region (writable, seeded from ``OVMF_VARS_TEMPLATE``).
+
+Region starts are aligned to 1 MiB. The exact sizes come from the seed
+template files and are recorded in the superblock at seed time; the disk
```
saschalucas, line 134: History teaches, that the size of the CODE might change (2M to 4M). Theoretically the VARS might, too. Should the layout of the firmware disk be sized to have some fixed headroom for the future? For example:
* 0-1M superblock with exact mappings of CODE, VARS etc. 
* 1-9M CODE
* 9-11M VARS
This would enable upgrading to a new size without overwriting neighbor slices.

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+size stays a fixed constant. The layout logic lives in an I/O-free module
+(``hv_kvm/firmware.py``) usable both master-side (to size the disk) and
+node-side (to seed and to expose the regions).
+
+The layout is designed to hold additional firmware/UEFI-related data regions
+in the future (e.g. swtpm).
+
+Configure-time templates
```
saschalucas, line 142: Pinning templates at configure time would make further development (i.e. secure boot) harder. Is there a chance to have some additional level of indirection, i.e. pointing to a config file holding the pahtes to CODE and VARS?

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+driver has no offset/size option - yet code and vars are two regions of
+one disk. Each region is therefore presented as a standalone whole
+device via a device-mapper ``linear`` target
+(``dmsetup create <name> --table "0 <sectors> linear <base> <offset>"``,
+``--readonly`` for the code region):
+
+- **block-backed templates** (drbd/plain/rbd/ext): the firmware disk's
+  node-local symlink already is a ``/dev`` block node, mapped directly -
+  no loop devices. For RBD this requires the firmware volume to be
+  ``rbd map``-ed, which is why access is forced to kernelspace.
```
saschalucas, line 210: I don't know if it is possible, but there might be a chance, that [qemu-fuseblk](https://www.qemu.org/2021/08/22/fuse-blkexport/) can represent a rbd-userspace image as a raw file? But maybe not for initial implementation.

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+LXC), never ``dmsetup``.
+
+Command generation vs. side effects
+-----------------------------------
+
+``_GenerateKVMRuntime`` stays side-effect-free: it only emits the two
+``-blockdev`` nodes (before the ``-machine`` line that references them)
+and the ``pflash0``/``pflash1`` machine properties, pointing at
+**deterministic** dm paths (``<instance>-ovmf-code`` /
+``<instance>-ovmf-vars``). The actual ``dmsetup``/``losetup`` happens in
```
saschalucas, line 229: the dm path could use the instance uuid instead of the name, which prepares for renaming without downtime, but makes it less human friendly.

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+Movement and failure paths
+===========================
+
+Most paths carry the firmware disk **for free** because they iterate
+``cfg.GetInstanceDisks`` generically:
+
+- **Live migration** - ``_OpenInstanceDisks`` / ``CheckDiskConsistency`` /
+  the DRBD standalone/reconnect/sync dance all include it; QEMU migrates
+  the pflash state to the target's vars region. Full support for DRBD and
+  shared-storage templates.
```
saschalucas, line 265: maybe more precise wording here: from my understanding only one can be true: either QEMU migrates a state, or the disks are shared-storage to carry the state. If both are "active" the state migration can destroy the source if migration fails.

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+  rejected while ``boot_type=uefi`` (would brick the instance).
+- **``MAX_DISKS`` (16) counts the firmware disk**, so a UEFI instance
+  supports at most 15 data disks. It consumes no PCI slot (pflash is a
+  machine property), so only the config-DB count is affected.
+- **Switching ``boot_type``** is a disk lifecycle transition, modeled on
+  the instance-communication NIC DDM. Switching a **stopped** instance to
+  ``uefi`` creates and seeds a firmware disk (node resource locks are
+  acquired for it, as for a template change). Switching **away** from
+  ``uefi`` keeps the firmware disk in place but inert (no pflash emitted);
+  it is never silently destroyed, and can be removed explicitly later.
```
saschalucas, line 346: Haven't tested this PR yet, but switching `boot_type` can also be made at cluster level (effecting all / running instances), so something like instance start must seed CODE and VARS, not the switch directly?

doc/design-uefi-boot.rst
rbott: I am not sure if that will ever happen and if it does, I'd tend to add code that resizes/changes the data structure. I would build an "update-code" function in that way anyways (e.g. check if the new CODE fits into the current allocation or move stuff around and update the superblock accordingly).

doc/design-uefi-boot.rst
rbott: These are just the default values the cluster uses. "In the wild" the paths differ between Linux distributions and this is just a means for package mainainers to package ganeti with a default path that matches their distribution/environment.

In the end, the cluster operator can switch to a different CODE path using the `ovmf_code` hvparam (so it can be set on cluster or instance level).

I see your point regarding VARS path. An hvparam felt wrong here because the VARS "source" is only read once at instance creation (or when transitioning from non-UEFI to UEFI boot mode) and changing that parameter afterwards has no effect whatsoever. Would it make sense to have a separate `gnt-instance` CLI param (and corresponding RAPI parameter) to overwrite the VARS path upon instance creation?

doc/design-uefi-boot.rst
rbott: I agree, that sounds a bit complicated for the initial implementation. I'd rather wait and see if there is any negative feedback on the "requires kernelspace" part.

doc/design-uefi-boot.rst
rbott: That's true, but at the same time I'd vote for keeping the "naming things in the filesystem" part stable (e.g. keep using instance names as identifiers) and migrate all of that at once when we really want to implement online-instance-renaming.

doc/design-uefi-boot.rst
rbott: It meant to say: QEMU migrates the in-memory state, if that even exists with pflash/VARS. I'll try to rephrase that so  it's less confusing!

doc/design-uefi-boot.rst
rbott: Altering the `boot_type` hvparam does not trigger any disk operations - as you assumed that will be handled during the next instance start operation when Ganeti detects that no firmware disk has been created so far.

rbott: I'll rephrase that a bit to be more explicit/clear

doc/design-uefi-boot.rst
saschalucas: The easiest would be to have a `ovmf_vars` hvparam, too. Both (`ovmf_code` and `ovmf_vars`) are only read once at instance creation (or when transitioning from non-UEFI to UEFI boot mode). This way one can easily test various combinations:

```
$ grep -h filename /usr/share/qemu/firmware/* | sort -u
            "filename": "/usr/share/OVMF/OVMF_CODE_4M.fd",
            "filename": "/usr/share/OVMF/OVMF_CODE_4M.ms.fd",
            "filename": "/usr/share/OVMF/OVMF_CODE_4M.secboot.fd",
            "filename": "/usr/share/OVMF/OVMF_VARS_4M.fd",
            "filename": "/usr/share/OVMF/OVMF_VARS_4M.ms.fd",
```

rbott: You're right, I'll add that parameter. It just needs to be documented that this parameter will not have any effect whatsoever after the VARS data section has been initially created (at least not until additional vars/code maintenance commands are introduced).

saschalucas:
Just a small update: I've started manual testing (or mostly prepared for it [1](https://github.com/saschalucas/instance-guestfish/commits/master/?since=2026-09-08&until=2026-09-08), [2](https://github.com/saschalucas/ganeti-test-bed/commit/14f9653b46177174c99fef4a0d7b17ab4e283c0b), so that I can create bios and uefi instances from the same OSI).

Appending the firmware disk at the tail also does not break OSI assumptions, that the first disk is for the OS.

Generally UEFI seems to work, for my surprise also with pc-i440fx machine type, which may not be recommended?

There are some workflows, which needs improvement:
```
gnt-cluster modify -H kvm:boot_type=uefi
gnt-instance reboot test.vm 
Waiting for job 12 for test.vm ...
Job 12 for test.vm has failed: Failure: command execution error:
Could not reboot instance: Hypervisor error: Instance test.vm requests UEFI boot but has no firmware disk

# notice: the instance is now in ERROR_down

# gnt-instance modify -H boot_type=uefi test
Tue Sep  8 13:49:39 2026  - WARNING: Error while executing backend function: [Errno 2] No such file or directory
Tue Sep  8 13:49:39 2026  - INFO: Modification will take place without hotplugging.
```
It seems the "conversion / adding and seeding firmware disk" from bios to uefi still takes place in `gnt-instance modify` and not at `start`? Also a `--force` switch sounds reasonable, because it would break OS boot, without special preparation.

Live migration (and disk-template conversion) works for: DRBD, sharedfile, rbd (kernelspace), rbd (userspace). The last one, rbd userspace, seems to map all instance disks via kRBD, not only the firemware disk, right? If so, can only firmware be kRBD mapped?

doc/design-uefi-boot.rst
```diff
@@ -0,0 +1,392 @@
...
+    UEFI/OVMF boot driven by ``boot_order`` (see below).
+
+``boot_type`` is authoritative: ``kernel_path`` / ``initrd_path`` /
+``kernel_args`` are honored only under ``direct_kernel``, and
+``boot_order`` only under ``bios`` / ``uefi``. Using an empty
+``kernel_path`` to toggle disk boot is **deprecated** and documented as
+such in the man page, but not enforced with a runtime warning - because
+``kernel_path`` has a non-empty default, such a warning would fire for
+essentially every ``bios``/``uefi`` instance.
+
```
saschalucas, line 67: It may be noted, that UEFI  + direct_kernel boot is supported, if kernel and initrd are EFI stub binary (which is true for modern distros). See example below.

I agree to have a single single-source-of-truth parameter regarding the `boot_type`, however there is no need to cut of direct kernel boot with uefi. On the contrary, it would enable some easier booting of UEFI instances, skipping all the clutter from ESP, Bootloader to NVRAM. The QA suite may be a good candidate to benefit from this way, or ganeti-instance-debootstrap can keep working with UEFI.

In the below example, I bypassed the firmware disk and added an UEFI BIOS via `kvm_extra`. The kernel/initrd supplied are the one from Ganeti nodes. Console Output is appended to verify EFI boot:
```
# gnt-instance add -t plain -o noop --disk 0:size=5G --net 0:network=vm-net,ip=pool -B memory=1G,vcpus=1 -H kvm:boot_type=direct_kernel,kvm_extra='-bios /usr/share/ovmf/OVMF.fd',kernel_path=/boot/vmlinuz-6.12.90+deb13.1-amd64,initrd_path=/boot/initrd.img-6.12.90+deb13.1-amd64 test-direct.vm && gnt-instance console test-direct.vm
...
Fri Sep 11 07:32:42 2026 * starting instance...

EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
[    0.000000] Linux version 6.12.90+deb13.1-amd64 (debian-kernel@lists.debian.org) (x86_64-linux-gnu-gcc-14 (Debian 14.2.0-19) 14.2.0, GNU ld (GNU Binutils for Debian) 2.44) #1 SMP PREEMPT_DYNAMIC Debian 6.12.90-2 (2026-05-27)
[    0.000000] Command line: root=/dev/vda1 ro console=ttyS0,38400 initrd=initrd
...
[    0.000000] efi: EFI v2.7 by Debian distribution of EDK II
[    0.000000] efi: SMBIOS=0x3f988000 ACPI=0x3fb7e000 ACPI 2.0=0x3fb7e014 MEMATTR=0x3dd6d518 INITRD=0x3dd62f18 RNG=0x3fb74018 
[    0.000000] random: crng init done
[    0.000000] secureboot: Secure boot disabled
[    0.000000] SMBIOS 2.8 present.
[    0.000000] DMI: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 2025.02-8+deb13u1 11/08/2025
[    0.000000] DMI: Memory slots populated: 1/1
[    0.000000] Hypervisor detected: KVM
```

doc/design-uefi-boot.rst
rbott: Err, I was entirely wrong here (should not have added the response without proper checking, sorry!). Creating the firmware disk actually *does* happen during `gnt-instance modify` and totally misses `gnt-cluster modify` and other situations.

I played through multiple scenarios and decided against moving firmware-disk-creation to instance startup time (as that might add unexpected blockdevice-mingeling to the instance startup which usually does not happen at that point) - and it might break in corner cases (e.g. an operator changes `boot_type` to `uefi`, then the primary node dies and the instance cannot be failovered because it now wants to set up a new DRBD pair first).

However, I added a guard to the `gnt-cluster modify -H...boot_type=uefi` operation so that it requires a `--force` parameter because it might create a large amount of blockdevice operations on a large cluster.

I will push the changes once I have tested them at least superficially.

doc/design-uefi-boot.rst
rbott: That's true; kind of forgot about the direct kernel boot with EFI. I'll add that variant as well.

rbott:
> Generally UEFI seems to work, for my surprise also with pc-i440fx machine type, which may not be recommended?

I assumed so as well but apparently there's nothing wrong about it and it works (at least all my tests worked fine) so I did not see any reason to stop users from booting that combination. We could add a warning in the "parameter assessment" section of `gnt-cluster verify` but right now I don't have any valid arguments against using that combination (other than "in theory, UEFI predates the emulated hardware by several years" :slightly_smiling_face:).

> 
> It seems the "conversion / adding and seeding firmware disk" from bios to uefi still takes place in `gnt-instance modify` and not at `start`? Also a `--force` switch sounds reasonable, because it would break OS boot, without special preparation.

Yes, that was a misconception on my end about my own code, sorry for that confusion. 
Ganeti now refuses to set `boot_type` to `uefi` on cluster level *if* it affects any instance (if there are none *or* all existing instances are already set to `uefi`, the command will succeed). If you provide the `--force` argument, it will succeed even if there are instances affected by the change and add jobs to the queue to create the missing firmware disks for all instances *that are currently not running*. That is the same behaviour as with `gnt-instance modify -H boot_type=uefi`.

Why does it not work on *running* instances: the storage-level operations acquire a set of locks which prevents these operations from execution while the instance is running. This is simply not possible with Ganeti's current job/lock design.

However, (re-)starting an instance which is set to `boot_type=uefi` but is missing the firmware disk now produces a clear/actionable error message so that the operator knows how to fix this situation.

I also looked at the option to lazily create (missing) firmware disks at instance start but that is even more complex (because currently there are no storage-level operations involved at that stage; that all happens earlier in the process). Since I committed to re-use the existing storage-layer of Ganeti for the firmware disks I think we should use it as intended.

> If so, can only firmware be kRBD mapped?

From looking at the code I cannot confirm that non-firmware disks are krbd-mapped when userspace RBD is configured. Could you recreate the situation and copy/paste the QEMU commandline from `ps`? That would be the easiest way to check if a mix of user- and kernelspace access is used. 

I am also working on an extension to the QA suite so that several UEFI-related features are tested. My QA environment only uses direct kernel boot so far because it relies on a kernel + initrd image for the nested guests, but it should be possible to test many aspects of the UEFI implemention even on instances that are never started.

Regarding direct-kernel-boot-with-UEFI: I still need to look into that. 

Oh and one more thing: I would clearly mark UEFI boot as an experimental feature (because critical features like proper export/import and ovmf maintanenance commands are still missing) and keep it that way until e.g. a 4.1 release (or whenever we declare it stable). I will add some hints about that to the man pages and it will be clearly visible in the release notes.

saschalucas:
Thanks @rbott for your estimation regarding "create (missing) firmware disks at instance start". So staying with `gnt-instance modify` sounds reasonable.

The enforced workflow via `gnt-instance modify -H boot_type=uefi` seems to work (also when changing at cluster level). OTOH instances created with `uefi` and later switching at cluster level to `bios` renders all instances unbootable. The user will notice this only when rebooting instances, which could have much time in between.

Therefore I wonder if changing `boot_type` at cluster level (regardless of the direction) can be improved somehow?

I'm thinking that the parameter `boot_type` should serve as a seeding value for newly added instances. Meaning, that `boot_type` is not inherited from cluster to instance, but explicitly set at instance level at creation time. This way:

* no instance will break at reboot
* the admin is enforced to modify each instance individually (if desired)
* the `gnt-cluster modify -H kvm:boot_type=uefi --force` and followed firmware disk addition to down instances would be obsolete

I don't know if this is possible. Maybe just checking at `instance add` weather `boot_type` is user-supplied and when not inject it from cluster seed value as it would have been user-supplied. What do you think?

rbott:
For reference: the last push just rebased on current master and fixed conflicts; I have not yet had the time to address the other open points.

rbott:
> Therefore I wonder if changing `boot_type` at cluster level (regardless of the direction) can be improved somehow?

There are already quite some lines of code accumulated that deal with these situations (and as you point out - it's not even enough yet :slightly_smiling_face:). Especially when you think about the practical value: how often will operators actually *flip*  `boot_type` on an existing instance? Going vom `bios` to `uefi` might make sense, but still involves quite some work on the guest side. Going back is even less likely (unless you fat-fingered and went from `bios` to `uefi` by accident). It does not feel right to spend even more lines of code on this problem.

> I'm thinking that the parameter `boot_type` should serve as a seeding value for newly added instances. Meaning, that `boot_type` is not inherited from cluster to instance, but explicitly set at instance level at creation time. 

While that would make `boot_type` an absolute snowflake compared to other hvparams, it sounds like the much simpler option than the current (incomplete) implementation. I will think about that a bit more, thanks! And honestly, even with the current solution `boot_type` is already a snowflake, at least when modified on cluster level.

saschalucas:
> > I'm thinking that the parameter `boot_type` should serve as a seeding value for newly added instances. Meaning, that `boot_type` is not inherited from cluster to instance, but explicitly set at instance level at creation time.
> 
> While that would make `boot_type` an absolute snowflake compared to other hvparams, it sounds like the much simpler option than the current (incomplete) implementation. I will think about that a bit more, thanks! And honestly, even with the current solution `boot_type` is already a snowflake, at least when modified on cluster level.

Sorry for my doubts. Maybe I overestimate the cluster defaults and inheritance at instance level. There are already things like `boot_order` acting in the same way (breaking instance boot). So don't waste much time, I think we can go with it as is (your argument "how often a admin will change this cluster settings" seems right, too).

Probably what remains to think of is, if the "firmware disk addition to down instances" by changing `boot_type` with `--force` at cluster level can be omitted? It already just modifies down instances and all other get a warning at start about missing firmware disk and `gnt-instance modify -H boot_type=uefi` invocation (probably removing the snowflake character of `boot_type` at cluster level)? Was it my $previous "doubts" that lead to adding this feature? Sorry for spin in a circle :-(.

Other than that above, there seems only "UEFI direct kernel boot" to think about.

I already verified that only the firmware disk is rbd-mapped (kernelspace), while the other stay userspace. And noticed that ovmf_vars is its own HV param.

rbott:
I decided to switch to the "seed model". This really takes away a bunch of code related to `gnt-cluster modify -H kvm:boot_type...` which is a) complex and b) rarely needed (as discussed above).

I hope I have updated all relevant documentation parts. There's still missing support for UEFI direct kernel boot, but that's still on the list.

PR #1956, opened 2026-07-25
fix KVM runtime upgrades for discard and chroot

saschalucas:
With commit 377cfd5 the kvm_runtime has been separated into a class, but with that a python indentation error causes updates regarding disk_discard to be skipped, if VNC is not enabled. Later chroot fixes (commit 9e56387) start erroneously at the same indentation level.

For master this is addressed in PR #1953 where the problem was first discovered.


rbott:
QA run started, I'll update here with the results

*Update*: Passed [here](https://stargazer.ganeti.org/a6e28e73c129a5e1a80a41d6fa829642996558cc/). Please merge your changes :-)

rbott (approved):

PR #1957, opened 2026-07-27
Add JSON output support to hcheck and surface hcheck findings in cluster verify

### LINKED ISSUE #1952: gnt-cluster verify does not report N+1 violation detected by hcheck
**Summary**
hcheck -L and gnt-cluster verify report different results regarding N+1 redundancy violations in the same cluster state.
In the observed scenario, hcheck -L reports that one node group contains nodes which are not N+1 compliant, while gnt-cluster verify completes the N+1 memory redundancy verification without reporting any issue.
This makes it difficult to determine the actual redundancy status of the cluster and raises questions about the consistency of the underlying verification logic.

**Expected Behaviour**
Both tools should report a consistent N+1 redundancy status.
If a node group contains a node that is "not N+1 happy" according to hcheck, either gnt-cluster verify should report the same issue.

codefritzel:
This change adds JSON output support to hcheck and keeps the implementation scoped to hcheck itself. It also extends gnt-cluster verify to run hcheck when available and report its findings as errros, while preserving the existing verify logic.

This provides a structured way to consume hcheck results and makes cluster verification reflect the checks performed by hcheck more closely. With this change, the issue is #1952 addressed and closed.

@saschalucas or @rbott please check if this is relevant to you; if not, please close this PR.

closes #1952

> [!NOTE] 
> I've only tested this against unittests and shell tests (make hs-shell-hcheck), not against a real cluster!

### DIFF
--- a/lib/cmdlib/cluster/verify.py
+++ b/lib/cmdlib/cluster/verify.py
@@ -30,6 +30,7 @@
 """Logical units for cluster verification."""
 
 import itertools
+import json
 import logging
 import operator
 import re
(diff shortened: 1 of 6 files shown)


codefritzel:
I think `--machine-readable` and `--output json` should remain separate concepts.

--machine-readable is an existing, htools-wide compatibility mode that produces key-value pairs like:
```text
HCHECK_INIT_CLUSTER_N1_FAIL=0
HCHECK_OK=1
....
```
The `--machine-readable` parameter supports only `yes` and `no` as boolean choices.
I think this format should be preserved for existing scripts and monitoring integrations.

For the new structured output, I believe `--output json` is the better choice. JSON is easier to parse reliably, represents nested cluster and group data more naturally, and is better for consumption e.g. gnt-cluster verify, RAPI, and other programmatic clients.

My suggestion would be:
- Keep `--machine-readable` unchanged for backwards compatibility.
- Use `--output text` for the default human-readable output.
- Use `--output json` for structured output.
- Treat the JSON structure as a documented and stable schema if it is exposed through RAPI or consumed by external clients.


Therefore, I would recommend using --output json for the new integration while retaining `--machine-readable` as the existing legacy format.


@rbott  and @saschalucas  I haven't heard back from you in three weeks. If I don't hear from you within another week, I will close this PR unless you let me know that you want the change implemented in the project.

codefritzel:
Closed because of no response.

saschalucas:
> Closed because of no response.

Damn, missed the deadline by 2 days.

It seems that a new check `_VerifyWithHcheck()` was added to `cluster verify`, where instead it should replace the old and wrong python N+1 check `_VerifyNPlusOneMemory()`?

PR #1961, opened 2026-08-28
Fix disk wiping for small disks

peixian:
WipeDisks derives the chunk size from a percentage of the disk size. For disks smaller than 20 MiB, rounding down produces a zero-sized chunk, so the wipe loop never advances.

Clamp the chunk size to at least 1 MiB while retaining the existing maximum. Add regression tests for small and zero-sized disks, start offsets, the maximum chunk size, and synchronization recovery after a wipe failure.

Assisted-by: OpenAI Codex

### DIFF
--- a/lib/cmdlib/instance_storage.py
+++ b/lib/cmdlib/instance_storage.py
@@ -1266,11 +1266,14 @@ def WipeDisks(lu, instance, disks=None):
 
   try:
     for (idx, device, offset) in disks:
-      # The wipe size is MIN_WIPE_CHUNK_PERCENT % of the instance disk but
-      # MAX_WIPE_CHUNK at max. Truncating to integer to avoid rounding errors.
+      # The wipe size is MIN_WIPE_CHUNK_PERCENT % of the instance disk, at
+      # least 1 MiB but MAX_WIPE_CHUNK at max. Truncating to integer to avoid
+      # rounding errors.
       wipe_chunk_size = \
-        int(min(constants.MAX_WIPE_CHUNK,
-                device.size / 100.0 * constants.MIN_WIPE_CHUNK_PERCENT))
+        max(1,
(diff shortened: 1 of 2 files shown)


rbott:
Did you actually hit that on a real-world Ganeti Cluster?

peixian:
@rbott yep. We hit this during VM creation of some with a 4MB disk, specifically

```
Result: 
        - OpExecError
        - - float division by zero
```

in gnt-job show. This occurs for a silly reason in `_CalcEta`, which you can "repro" with:

```python

MAX_WIPE_CHUNK = 1024
MIN_WIPE_CHUNK_PERCENT = 10

size = 4
offset = 0

wipe_chunk_size = int(min(MAX_WIPE_CHUNK, size / 100.0 * MIN_WIPE_CHUNK_PERCENT)) # int(0.4)  => 0

wipe_size = min(wipe_chunk_size, size - offset) # 0
...
offset += wipe_size # still zero
...
# enter _CalcEta(..., offset, ...)

avg_time = time_taken / float (offset) -> # blows up here :( 
```

rbott:
Hi @peixian, sorry for the misunderstanding. At first I thought this to be some sort of drive-by-AI-PR without real-world usefulness. But then I found your blog and figured that you actually _do_ stuff with virtualisation :slightly_smiling_face: 

I'll take a closer look at this in the next days, thanks for your contribution!

peixian:
@rbott all good, I probably should've added more context on the original PR to make it seem less llm-y. Let me know if you have any feedback. I'm in the middle of setting up a personal ganeti cluster so these things are more easily verified.

rbott:
I tested it and it looks good, thanks for your contribution!

rbott (approved):

yousefbahrambeigi:
Run

PR #1962, opened 2026-09-02
feat: add hypervisor field to rapi

codefritzel:
add existing 'hypervisor' field to rapi endpoint
GET /2/instances/<name>


saschalucas (approved):
Thanks @codefritzel.

PR #1965, opened 2026-09-04
logrotate: match today’s requirements

saschalucas:
Switch from weekly to daily rotation, but keep only 90 days instead of 52 weeks. Add create mode and su directive. Without the last, modern logrotete refuses to rotate logfiles not owned by root.


rbott (approved):
LGTM, please merge!

PR #1966, opened 2026-09-04
cluster verify: replace python N+1 by hcheck

saschalucas:
Htools are fist class tools i.e. for instance allocation (`hail`) and cluster balancing (`hbal`). They track resources like RAM and others to guarantee N+1 redundancy (from the state of record). Historically cluster verify has its own simple version of N+1 check (state of world). Correctly monitoring N+1 redundancy requires usage of `hcheck` to compensate for state of world/record discrepancies: i.e. an instance with 10G RAM assigned (recorded state) vs. 5G currently using (real world usage state).

Replace the simple python algorithm by calling `hcheck` an evaluating its output.

Inspired by #1957, but replacing old python algorithm and using existing machine readable output.


rbott:
QA run failed with this:
```
Mon Sep  7 01:06:10 2026 * Verifying cluster config
Mon Sep  7 01:06:10 2026 * Verifying cluster certificate files
Mon Sep  7 01:06:10 2026 * Verifying hypervisor parameters
Mon Sep  7 01:06:10 2026 * Verifying all nodes belong to an existing group
Waiting for job 4821 ...
Mon Sep  7 01:06:11 2026 * Verifying group 'default'
Mon Sep  7 01:06:11 2026 * Gathering data (3 nodes)
Mon Sep  7 01:06:11 2026 * Gathering information about nodes (3 nodes)
Mon Sep  7 01:06:12 2026 * Gathering disk information (3 nodes)
Mon Sep  7 01:06:12 2026 * Verifying configuration file consistency
Mon Sep  7 01:06:12 2026 * Verifying node status
Mon Sep  7 01:06:12 2026 * Verifying instance status
Mon Sep  7 01:06:12 2026   - WARNING:EINSTANCEPOLICY:instance:kvm-test-instance02.staging.ganeti.org:memory-size value 768 is not in range [772, 772]
Mon Sep  7 01:06:12 2026 * Verifying orphan volumes
Mon Sep  7 01:06:12 2026 * Verifying N+1 Memory redundancy
Mon Sep  7 01:06:12 2026   - ERROR:ENODEN1:node:None:There are 0 nodes not N+1 happy and 1 nodes not directly evacuateable. Use hbal --print-nodes to find out which nodes and how to possibly solve the problem
Mon Sep  7 01:06:12 2026 * Assessing cluster hypervisor parameters
Mon Sep  7 01:06:12 2026   - cpu_type is currently set to 'host', please make sure all your cluster nodes have the exact same CPU type to allow live migrations.
Mon Sep  7 01:06:12 2026   - Spice is configured but without TLS, please consider setting spice_use_tls to 'true' for additional security.
Mon Sep  7 01:06:12 2026   - machine_version is not explicitly configured so Ganeti will autodetect the default value from 'kvm -M ?', this might produce unexpected results.
Mon Sep  7 01:06:12 2026   - Please refer to the gnt-instance man page for detailed information on the usage of each hypervisor parameter
Mon Sep  7 01:06:12 2026 * Other Notes
Mon Sep  7 01:06:12 2026   - NOTICE: 1 non-redundant instance(s) found.
Mon Sep  7 01:06:12 2026 * Hooks Results
```

Entire output plus logs can be found here: https://stargazer.ganeti.org/733d8e7b7f62ceb5e57c37bc35250ed9531a615d/ (for the next 14 days, that is :smile:)

saschalucas:
> QA run failed with this:
> 
> ```
> Mon Sep  7 01:06:12 2026 * Verifying N+1 Memory redundancy
> Mon Sep  7 01:06:12 2026   - ERROR:ENODEN1:node:None:There are 0 nodes not N+1 happy and 1 nodes not directly evacuateable. Use hbal --print-nodes to find out which nodes and how to possibly solve the problem
> ```
> 
> Entire output plus logs can be found here: https://stargazer.ganeti.org/733d8e7b7f62ceb5e57c37bc35250ed9531a615d/ (for the next 14 days, that is 😄)

Nice catch. It might be possible, that the cluster is indeed not N+1 redundant. Any chance to get the output of `hcheck -L --no-simulation` or `hbal -L --print-nodes` from such a cluster? I'll check the output logs if there are any traces of this.

OTOH, if the QA-Cluster is not N+1 redundant to not waste RAM, it is possible to run `gnt-cluster verify --no-nplus1-mem`

saschalucas:
> Nice catch. It might be possible, that the cluster is indeed not N+1 redundant.

No it's not. The QA just test ipolicy violation and sadly `hcheck` also triggers a global N+1 redundancy failure while instances exists, that violate ipolicy constrains. Since this is Haskell, chances to correct the behavior got reduce by some orders of magnitude.

rbott:
So it's partially a QA thing (because it deliberately causes this situation) but also `hcheck` should not error out in this scenario, right?

saschalucas:
> So it's partially a QA thing (because it deliberately causes this situation) but also `hcheck` should not error out in this scenario, right?

Yes, `hcheck` should not throw global N+1 violation in this QA scenario.

ATM I can't think of how ipolicy bounds spec should effect N+1 redundancy. Only soft errors like vcpu ratio etc. count towards solving N+1 issues. But this is one step before solving, just reporting (what `--no-simulation` does). Adding `--ignore-soft-errors` does not help either.

saschalucas:
> QA run failed with this:
> ```
> Mon Sep  7 01:06:12 2026 * Verifying N+1 Memory redundancy
> Mon Sep  7 01:06:12 2026   - ERROR:ENODEN1:node:None:There are 0 nodes not N+1 happy and 1 nodes not directly evacuateable. Use hbal --print-nodes to find out which nodes and how to possibly solve the problem
> ```

Fixed after some hours asking AI. For some more explanations see commit message.

rbott:
Quick Update: QA finished successfully, I'll take a closer look at the code within the next days.

rbott (approved):
LGTM, thanks for fixing this.

There is only one *minor* aspect: we are using `RunCmd` without the `timeout` parameter (which defaults to `None` / no timeout). I don't have much experience running `hcheck` but I wonder if it would be useful to cap the maximum runtime of the process to something like 30 or 60 seconds just to avoid a broken `hceck` blocking verify jobs. Feel free to ignore this or add a timeout, you may merge either way :-)

test/py/unit/cmdlib/cluster/test_verify.py
```diff
@@ -0,0 +1,56 @@
...
+  return lu_cluster_verify_group
+
+@pytest.mark.parametrize("n1_fail, gn1_fail, expected_error_calls", [
+  (0, 0, 0),
+  (2, 0, 1),
+  (0, 3, 1),
+  (4, 5, 1)
+])
+def test_VerifyNPlusOneMemory(lu_cluster_verify_group, n1_fail, gn1_fail,
+                              expected_error_calls):
```
paspflue: Please convert to snake_case:
```suggestion
def test_verify_n_plus_one_memory(lu_cluster_verify_group, n1_fail, gn1_fail,
                              expected_error_calls):
```

saschalucas:
> There is only one _minor_ aspect: we are using `RunCmd` without the `timeout` parameter (which defaults to `None` / no timeout). I don't have much experience running `hcheck` but I wonder if it would be useful to cap the maximum runtime of the process to something like 30 or 60 seconds just to avoid a broken `hceck` blocking verify jobs.

Thanks @rbott for the timeout hint. Normally when `hcheck` is run with `--no-simulation` it should be always fast. To be safe, I added 30 seconds.

test/py/unit/cmdlib/cluster/test_verify.py
saschalucas: done

saschalucas:
The #1968 issue has landed in Debian testing CI images. From now on they are failing.

PR #1970, opened 2026-09-05
docs: Specify amount unit for instance disk grow

codefritzel:
see /2/instances/[instance_name]/disk/[disk_index]/grow


saschalucas (approved):
Thanks @codefritzel.

Issue #1654, opened 2022-02-16
Jobs slowed down by RPC read loop


rowlap:
In rpc.transport.Transport.Recv the critical inner loop is
- read up to 4096 bytes from the socket
- scan for `LUXI_EOM`
- break / return when any complete message is received

###  Problem 1

4096 bytes is rather small, and has been there [since the beginning](https://github.com/ganeti/ganeti/commit/c2a03789be0213c07e56464222066adccfa5f051
).

The number of loop iterations would be vastly reduced for a modest increase in read size. On my Linux box, `net.core.wmem_default=212992` is the most data which could be sitting in the socket buffer waiting to be read.

### Problem 2

When we scan `self._buffer` for `LUXI_EOM`, we're checking data which has already been checked.

```
new_msgs = (self._buffer + data).split(constants.LUXI_EOM)
```

We can assert that `self._buffer` does not contain `LUXI_EOM`, yet the whole string is re-scanned each time.

For reasons, my config.data file has grown to over 4MB, taking 1000 loops just to receive the result of each `readConfig` call, of which there are several from each `INSTANCE_STARTUP` job. It appears from profiling that most of the time is spent in `Recv()`.

There is also a good amount of code duplication in FdTransport which might also benefit from any performance improvements.

PR #1972, opened 2026-09-09
Report WConfd update timeouts in job errors

peixian:
Often, we will see errors in `gnt-job info` with 

```
     Result:
        - OpExecError
        - - 
```

This is quite frustrating since it gives you nothing to go on. After some back and forth, I believe this is caused by `OpExecError(str(err))`. The `err` gets raised from `utils.RetryAgain()`, which gets parsed as an empty string. 

This adds a nonlocal poll counter to track how many times the retry has occurred, as well as some additional debugging information. 

-----

Testing wise this is a basic dummy one:

```
Python 3.11.14 (main, Oct  9 2025, 16:16:55) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> import json
>>> from ganeti import config, jqueue, objects
>>>
>>> class BusyWConfd:
...   def UpdateInstance(self, instance):
...     return None
...
>>> try:
...   config.ConfigWriter(wconfd=BusyWConfd()).Update(
...     objects.Instance(uuid="demo-instance"), None)
... except Exception as err:
...   print("Raised:", type(err).__name__)
...   print(json.dumps(jqueue._EncodeOpError(err)))
...
Raised: RetryTimeout
["OpExecError", ["WConfd config-lock timeout - UpdateInstance for object UUID demo-instance; config lock unavailable on all 301 polls (30-second retry budget)"]]
```

I also tested this live on a Ganeti cluster and got 

```
Status: error
      Result:
        - OpExecError
        - - WConfd config-lock timeout - UpdateInstance for object UUID <UUID>; config lock unavailable on all 300 polls (30-second retry budget)
      Execution log:
```

Assisted-by: OpenAI Codex

### DIFF
--- a/lib/config/__init__.py
+++ b/lib/config/__init__.py
@@ -3033,12 +3033,19 @@ class ConfigWriter(object):
       raise errors.ProgrammerError("Invalid object type (%s) passed to"
                                    " ConfigWriter.Update" % type(target))
 
+    polls = 0
+
     def WithRetry():
+      nonlocal polls
+      polls += 1
       result = update_function(target.ToDict())
       self.OutDate()
 
       if result is None:
-        raise utils.RetryAgain()
+        raise utils.RetryAgain(
+          "WConfd config-lock timeout - %s for object UUID %s;"
+          " config lock unavailable on all %s polls (30-second retry budget)" %
+          (update_function.__name__, target.uuid, polls))
       else:
         return result
     vals = utils.Retry(WithRetry, 0.1, 30)
--- a/lib/jqueue/__init__.py
+++ b/lib/jqueue/__init__.py
@@ -670,7 +670,7 @@ def _EncodeOpError(err):
   if isinstance(err, errors.GenericError):
     to_encode = err
   else:
-    to_encode = errors.OpExecError(str(err))
+    to_encode = errors.OpExecError(str(err) or err.__class__.__name__)
 
   return errors.EncodeException(to_encode)
 
--- a/test/py/legacy/ganeti.jqueue_unittest.py
+++ b/test/py/legacy/ganeti.jqueue_unittest.py
@@ -78,6 +78,15 @@ class _FakeJob:
 
 
 class TestEncodeOpError(unittest.TestCase):
+  def testEmptyException(self):
+    self.assertEqual(jqueue._EncodeOpError(utils.RetryTimeout()),
+                     ("OpExecError", ("RetryTimeout",)))
+
+  def testRetryTimeout(self):
+    message = "WConfd config-lock timeout - UpdateInstance"
+    self.assertEqual(jqueue._EncodeOpError(utils.RetryTimeout(message)),
+                     ("OpExecError", (message,)))
+
   def test(self):
     encerr = jqueue._EncodeOpError(errors.LockError("Test 1"))
     self.assertTrue(isinstance(encerr, tuple))


rbott:
Hi @peixian,

could you elaborate a bit more when and in what scenario these errors occured? I have never seen that so far.

What helps:
- (redacted) `gnt-cluster info` output
- what environment are you running Ganeti in (e.g. Debian, nix-os, Ubuntu etc.) and which version of Ganeti?
- which kind of jobs come up with this error? any/random jobs? or specific job types?

I have started multiple QA runs against this branch and so far there are no problems, but I'd like to understand what the root cause of the problems you observe is.

peixian:
hey @rbott,

We're running a patched version of Ganeti that's slightly older than master. These are running on RHEL-based distributions

Unforunately I don't think `gnt-cluster info` will have anything super interesting, but we can see it occur fairly often for replace-disk jobs. Specifically on large Ganeti clusters (when config sizes are 4-5MB), there's many concurrent DRBD disk replaces. 

If we trigger some actions that dump a lot of disk-replaces on a single node (INSTANCE_REPLACE_DISKS) and lots of INSTANCE_MIGRATES, this causes this particular behavior to occur.

We can somewhat reproduce this if you have many VMs on a single node, then run INSTANCE_REPLACE_DISKS on every secondary and INSTANCE_MIGRATE on every primary on that node

rbott:
oh wow - I have not yet myself operated a cluster where `/var/lib/ganeti/config.data` exceeded 1M in size. I'll see what I can come up with from the pointers you gave me (no need for more information). Thanks!

peixian:
Btw in my personal testing DRBD metadata takes up a ton of space in the config metadata. And wconfd processing the file seems to be quadratic with file size

saschalucas:
That reminds me of issue #1654 where also a large config.data is used. Haven't noticed this problem yet on DRBD cluster with 2.3M.

rbott:
From what I understand, wcfond's performance degrades linearly (or maybe even worse) with growing `config.data` size.

It uses the somewhat imperformant Haskell JSON implementation (which could possibly be replaced by the much faster [aeson](https://hackage.haskell.org/package/aeson) library). We could implement that as configure-time alternative - so that users / maintainers can choose between built-in JSON or aeson (and maybe flip to aeson as the default in the future if it works well).

However, I am not proficient enough in Haskell to do that myself. I guess we have to look out for someone more competent in Haskell :-)

I am also sure there are other optimization paths as well, but this seems to be the low-hanging fruit of possible changes.

saschalucas:
I wonder how a big `config.data` can be simulate to do performance testing?

In #1654 it's observed, that python in the reading/receive phase is causing slowdowns on a big config... Maybe the problem has also a writing end in Haskell as @rbott pointed out?

rbott:
I am currently (load-) testing wconfd to have some numbers ready. I'll come back will the results soon (hopefully)

rbott:
So I fiddled around with a bit of AI help and build a wconfd-test-setup:

## Setup

- Ganeti built from source in the `ganeti/ci:trixie-py3` container
  (`--init`, `--privileged`, `nofile=1024`).
- vcluster (`tools/vcluster-setup`): single node for the idle sweep, 8 nodes
  (1 master + 6 master candidates) for the job-load sweep; fake hypervisor,
  diskless instances.
- `config.data` grown **offline** (stop wconfd, clone a template instance,
  restart wconfd/luxid) - reaches 5MB in seconds and distributes instances
  across node UUIDs.
- Latency measured directly at the wconfd unix socket: `Echo`, `ReadConfig`,
  `TryUpdateLocks` (shared BGL acquire + release).
- Load phase: 8 concurrent `gnt-instance modify -B memory=N` jobs (same
  exclusive-config-lock + full-serialize path as `INSTANCE_MIGRATE` /
  `INSTANCE_REPLACE_DISKS`), sampling wconfd latency while the queue drains.

Numbers are relative (container CPU scheduling); the signal is the trend.

## Idle latency vs. config size

| Config | Instances | Echo p50 | ReadConfig p50 | ReadConfig p95 | TryUpdateLocks p50 |
|---|---|---|---|---|---|
| 7.8 KB | 0 | 0.1 ms | 0.2 ms | 2.0 ms | 1.9 ms |
| 102 KB | 223 | 0.2 ms | 2.8 ms | 3.6 ms | 2.2 ms |
| 335 KB | 833 | 0.1 ms | 10.2 ms | 11.2 ms | 1.8 ms |
| 532 KB | 1133 | 0.1 ms | 17.5 ms | 18.9 ms | 1.7 ms |
| 1.0 MB | 2233 | 0.1 ms | 46.1 ms | 55.8 ms | 1.7 ms |
| 2.0 MB | 4433 | 0.1 ms | 154.7 ms | 155.5 ms | 1.7 ms |
| 3.0 MB | 6533 | 0.0 ms | 317.9 ms | 328.0 ms | 1.7 ms |
| 5.0 MB | 10933 | 0.2 ms | 882.0 ms | 899.2 ms | 1.9 ms |

`ReadConfig` is **superlinear** (per-100KB cost grows ~50x across the sweep);
`Echo` and `TryUpdateLocks` stay flat - the config-lock structure is O(1), the
cost is reading/serializing the config via `Text.JSON`.

## Under job load (8 concurrent modify jobs, 8-node vcluster)

| Config | ReadConfig p50 idle→load | ReadConfig max load | lock max load | lock timeouts |
|---|---|---|---|---|
| 2 MB | 148.6 → 177.5 ms | 516.0 ms | 753.6 ms | 0 |
| 3 MB | 318.0 → 722.2 ms | 2070.4 ms | 1526.3 ms | 0 |
| 5 MB | 886.6 → 2114.7 ms | 3779.6 ms | 1364.3 ms | 0 |

Under load, `ReadConfig` reaches **3.8s at 5MB** (4x idle) and lock latency
crosses **1.5s at 3MB**. No lock timeouts at 8 concurrent jobs, but the trend
is steep - higher concurrency or larger configs would plausibly breach the 30s
retry budget

**Sidenotes:** apparently at the size of the test luxid is also a problem in itself and has performance problems with e.g. `gnt-instance list` once there a few thousand instances (since my vCluster setup sports instances without disks and without nics, a lot of instances are needed to scale the DB up to 5MB). Due to that (and also some other reasons) the test always queried the WConfD socket directly.

I can add the code to run the tests as development tooling if you want. For reference: I ran these tests locally on a fairly recent Intel Core Ultra X9 388H.

saschalucas:
Your tests indicate that pure WConfD performance seems not the problem. However what about this phrase:
https://github.com/ganeti/ganeti/blob/b89d932273d6cdb6626a8004d8395055adc09b99/doc/design-configlock.rst?plain=1#L23-L34
Isn't it possible that the read takes to long?

peixian:
I think there might be two issues underneath:
- the exclusive lock can have a lot of contention under lots of operations, and since each lock is eagerly spinlocking, it can get pretty bad
- config reads get progressively slower as size increases, which interacts poorly with the lock.

The purpose of this specific PR is to make the error message when these conditions occur slightly more useful, so it can be caught. Right now, the OpExec empty error is frustrating to diagnose. 

Although I'm in agreement that we should consider some improvements to wconfd anyways.

rbott:
Yeah sorry @peixian, I went a little off-topic here. I'll open an issue and transfer the wconfd-related information over.

@saschalucas the worst case example (just reading a large config takes wconfd 2 seconds or almost 4 seconds under load) does show a problem: if that happens under high job pressure I am quite sure that things can pile up so that a 30 second retry window is not enough.

rbott (approved):

Issue #1960, opened 2026-08-10
Feature: Support Memory Hotplug for KVM Instances


paspflue:
**Summary**
Add runtime memory hotplug (adding memory to running VMs) for KVM guests. 
This issue is focused on hotplug (add) only; hot-unplug (remove) can be staged later.

**Motivation**
Allow increasing guest memory without reboot.

**Options**
1. DIMM-based
`qemu-system-x86_64 -m 8G,slots=16,maxmem=128G`
- less flexible
- requires preconfigured slots/maxmem.
2. virtio-mem (my preference)
`-object memory-backend-ram,id=mem0,size=64G -device virtio-mem-pci,memdev=mem0`
- designed for dynamic allocation 
- cleaner future support for unplug
but needs modern QEMU/guest kernel and integration to work.

**Out of Scope (Future Extensions)**
- Memory hot-unplug support
- A configurable hypervisor parameter to switch beetween the two options

saschalucas:
> **Options**
>     1. DIMM-based
>        `qemu-system-x86_64 -m 8G,slots=16,maxmem=128G`
>     * less flexible
>     * requires preconfigured slots/maxmem.
>     2. virtio-mem (my preference)
>        `-object memory-backend-ram,id=mem0,size=64G -device virtio-mem-pci,memdev=mem0`    
>     * designed for dynamic allocation
>     * cleaner future support for unplug
>       but needs modern QEMU/guest kernel and integration to work.

An implementation should explain this options in a design doc, so that someone not familiar with memory hotplug can understand the ideas behind.

PR #1974, opened 2026-09-16
docs: Add memory-hotplug design document

paspflue:
see #1960


doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+   device IDs, slot mappings, or DIMM configurations in the runtime
+   file. The migration path is identical to cold-boot: the target
+   QEMU re-creates the device and restores its state from the migration
+   stream.
+
+3. **Future-proof for hot-unplug** — virtio-mem supports memory
+   shrinking by reducing ``requested-size``. When hot-unplug is
+   needed in the future, virtio-mem can implement it without
+   unplugging PCI devices. pc-dimm hot-unplug is more complex as it
+   requires guest cooperation and device removal.
```
rbott, line 120: I entirely agree with the argument but I do have one question: I assume virtio-mem still needs some sort of guest cooperation (aside from having virtio-mem support in the first place) when reducing the memory as it needs to actually free the amount of memory before it can be removed/returned to the host, right?

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+A new Python constant defines the node's hotplug ceiling:
+
+``KVM_VIRTIO_MEM_MAX_SIZE``
+   The maximum amount of memory (in MiB) that can be provided via
+   virtio-mem on this node. This is a fixed value, analogous to
+   ``maxcpus`` for vCPU hotplug. All instances on a node share the
+   same ceiling.
+
+   For the initial implementation, the ceiling is fixed (e.g.
+   ``65536`` for 64 GiB). Per-instance ceilings can be added later
```
rbott, line 154: Just out of curiosity: where does that number come from? 🙂

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+   (the node's hotplug ceiling).
+
+2. A virtio-mem-pci device is added, referencing the backend:
+
+   ::
+
+     -device virtio-mem-pci,id=vmem0,\
+     requested-size=0,block-size=1M
+
+   The device is placed on a free PCI slot by the existing bus
```
rbott, line 268: For Q35 there's a designated area where these types of devices can be placed (similiar to other devices like serial ports, sound cards etc.)

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+
+Other hypervisors
+   KVM only. Xen and LXC raise ``NotImplementedError`` when hotplug
+   is requested.
+
+Future work
+===========
+
+- pc-dimm support — Implement pc-dimm as an alternative to virtio-mem.
+  A hypervisor parameter (e.g. ``memory_hotplug_method``, with values
```
rbott, line 438: I would suggest to start directly with `memory_hotplug_method` with values `['none', 'virtio-mem']`

Pros:
- we can set all existing instances to `none` on cluster upgrade, making it explicit
- it leaves room for future options (like `pc-dimm`)
- it will be easier to e.g. disallow/allow ballooning based on memory hotplug being enabled or disabled

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+  ``virtio-mem`` or ``pc-dimm``) could control which mechanism is
+  used. pc-dimm is needed for scenarios where virtio-mem is not
+  suitable (e.g., guests without a virtio-mem driver, or when
+  discrete DIMM boundaries are required). pc-dimm requires PCI slot
+  reservation, device persistence in the runtime file, and migration
+  state reconstruction.
+
+- Memory hot-unplug — Support reducing the ``requested-size`` of the
+  virtio-mem device to free memory. Shrinking may fail if the guest
```
rbott, line 447: ah, that already answers my question from above. nevermind :-)

rbott (commented):
Hi @paspflue,

thanks for contributing the design doc. I mostly agree with it (and like the overall design/choices!), only very few comments/questions 🙂

doc/design-memory-hotplug.rst
paspflue: Yes, exactly. Hot-unplug still requires guest cooperation.

The advantage of virtio-mem is that this process is largely handled by the virtio-mem driver itself. When requested-size is reduced, the guest tries to migrate pages away from the affected blocks, offline them, and return them to QEMU automatically.

This is much more flexible than pc-dimm, where an entire DIMM usually has to be offlined before it can be removed.

doc/design-memory-hotplug.rst
paspflue: The number is somewhat arbitrary at the moment. The current idea is simply to add a single virtio-mem device with a fixed size (currently 64 GiB) to every instance where memory hotplug is enabled. 
At this stage, it's still unclear whether larger virtio-mem devices have any drawbacks.
For that reason, I picked 64 GiB as a conservative default to get the implementation started.

doc/design-memory-hotplug.rst
paspflue: What do you think of `0x02.6`?

rbott: sounds good!

doc/design-memory-hotplug.rst
paspflue: My main concern is keeping the initial implementation simple as possible, so I'd prefer not to introduce a new parameter yet.

For the current Ganeti use case, this likely isn't needed anyway, as we primarily target Linux guests and not Windows. If we later identify a real need for supporting multiple hotplug mechanisms (e.g. pc-dimm), we can still introduce a memory_hotplug_method parameter at that point.

For now, I'd keep it out to minimize complexity and avoid adding configuration options that don't provide immediate value.

rbott: I see the point of reduced complexity in Ganeti. OTOH I am also a fan of explicit configuration so that users/operators know what is happening. Having a variable that can be set to `none` (and defaults to `none`) would mean that operators will not see a change in virtual hardware without explicetly opting in.

saschalucas: > ... OTOH I am also a fan of explicit configuration so that users/operators know what is happening. Having a variable that can be set to `none` (and defaults to `none`) would mean that operators will not see a change in virtual hardware without explicetly opting in.

Good point: introducing memory-hotplug via virtio-mem, there are already two implementations to chose: none and virtio-mem. In this case I would vote for having a new HV parameter. Enabling unconditionally virtio-mem for all instances sounds not the right way... @paspflue: didn't think so far in our offline discussion.

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+=============
+
+The maximum guest memory currently requires a reboot because the
+QEMU memory topology is static after startup. Ganeti already exposes
+``minmem``, ``memory`` and ``maxmem`` as backend parameters. These are
+exclusively used for ballooning: the balloon driver adjusts the
+guest-visible memory within the ``minmem``..``maxmem`` range at runtime.
+Ballooning, however, only changes how much of the already available
+memory the guest actually uses — it cannot make *additional*
+guest-usable memory available beyond the original ``-m`` value. To
```
saschalucas, line 24: ... beyond the qemu ``-m`` value

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+provide more guest-visible memory, Ganeti needs memory hotplug.
+
+QEMU/KVM provides two mechanisms for memory hotplug:
+
+1. **pc-dimm** — Uses discrete memory devices (DAX devices on Windows)
+   that are hotplugged as individual DIMMs with fixed sizes. Requires
+   PCI slot reservation, explicit slot assignment, and the hypervisor
+   must persist DIMM device IDs and slot mappings for migration.
+
```
saschalucas, line 33: According to https://github.com/qemu/qemu/blob/c1c18d1e640b64292859ce9f30f3c344edfb0294/docs/memory-hotplug.txt#L9-L45 there is no mention of requiring a PCI slot?

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+Here a comparison:
+
+================== ================================== =========================
+Aspect             pc-dimm                            virtio-mem
+================== ================================== =========================
+Device model       Multiple fixed-size DIMMs          Single expandable device
+Hot-unplug         Supported (remove DIMM device)     Possible (reduce
+                                                      requested-size)
+PCI slot usage     One slot per DIMM; slots must be   One slot for the
+                   pre-reserved at boot time          virtio-mem-pci device
```
saschalucas, line 83: please check PCI slot usage for pc-dimm (see above).

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+
+Rationale for virtio-mem
+------------------------
+
+Virtio-mem is the preferred approach for the initial implementation
+for the following reasons:
+
+1. **Simpler implementation** — Only one device is needed at boot.
+   Memory growth is a single QMP call to adjust ``requested-size``.
+   No PCI slot reservation beyond one slot is required.
```
saschalucas, line 106: again the PCI slot argument WRT pc-dimm. Please check.

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+  ``requested-size`` to ``16384``. The guest now sees **24 GiB**.
+- The maximum guest-visible memory is ``KVM_VIRTIO_MEM_MAX_SIZE``.
+
+Choosing the block size
+-----------------------
+
+The virtio-mem device uses a ``block-size`` parameter that determines
+the granularity of memory plug/unplug operations. The initial
+implementation uses a fixed block size of 1 MiB.
+
```
saschalucas, line 200: 1M sounds very small. What about 32, 64 or 128M? Do we have some evidence what is reasonable here?

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+   gnt-instance modify -B memory=X instance-name
+
+This is distinct from the existing ballooning path which uses
+``maxmem``:
+
+::
+
+   gnt-instance modify -B maxmem=X instance-name
+
```
saschalucas, line 216: I think with ballooning `minmen` and `maxmem` are a bounds spec (min/max). The actual setting of runtime memory must be done via `gnt-instance modify --runtime-memory`.

So the interface for virto-mem could stay `-B memory=X`

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+  Ganeti does not reboot the instance on its own.
+- ``X < memory`` (reduction): the running instance is not modified
+  (hot-unplug is out of scope); the new value applies at the next
+  cold boot.
+
+Instances that are not running are updated as today.
+
+``gnt-instance modify -B maxmem=X`` continues to work as before
+and the instances are not available for that mechanism.
+
```
saschalucas, line 239: The only option to verify that no ballooning can happen is when memory=minmem=maxmem. And you can not get rid of the ballooning device (a hard qemu requirement).

ATM I'm to confused about the interface/interaction of ballooning/virtio-mem... give me some time...

doc/design-memory-hotplug.rst
```diff
@@ -0,0 +1,456 @@
...
+Future work
+===========
+
+- pc-dimm support — Implement pc-dimm as an alternative to virtio-mem.
+  A hypervisor parameter (e.g. ``memory_hotplug_method``, with values
+  ``virtio-mem`` or ``pc-dimm``) could control which mechanism is
+  used. pc-dimm is needed for scenarios where virtio-mem is not
+  suitable (e.g., guests without a virtio-mem driver, or when
+  discrete DIMM boundaries are required). pc-dimm requires PCI slot
+  reservation, device persistence in the runtime file, and migration
```
saschalucas, line 443: repetitive "pc-dimm requires PCI slot" argument, please check.

doc/design-memory-hotplug.rst
paspflue: I added the parameter to the design document and adjusted the rest accordingly.

doc/design-memory-hotplug.rst
paspflue: That's right, I confused memslots with PCI slots.

PR #1976, opened 2026-09-16
QA: honor `modify_ssh_setup: false` properly

rbott:
The QA suite already knows a config knob `modify_ssh_setup` to stop mingling with SSH keys when that breaks the QA environment (or the QA environment has been set up with `--no-ssh-init`). Unfortunatly it was not honored in all places, which is fixed by this PR.

This should unbreak `cluster-renew-crypto` tests which had been blocked by this issue on my QA clusters in the past years. 

Prequesite for my branch that fixes #1968 (will follow when this is merged without any problems).

PR #1977, opened 2026-09-17
Replace OpenSSL.crypto with the cryptography package

### LINKED ISSUE #1968: gnt-cluster renew-crypto fails - module 'OpenSSL.crypto' has no attribute 'X509Req'
Gentoo's Ganeti 3.1.1; feels like this codepath was missed in other python openssl updates

```
2026-09-05 17:00:19,656: gnt-cluster renew-crypto pid=126464 cli:1259 ERROR Error during command processing
  File "/usr/share/ganeti/3.1/ganeti/cli.py", line 1602, in Call
    return fn(self, *args)
           ^^^^^^^^^^^^^^^
  File "/usr/share/ganeti/3.1/ganeti/client/gnt_cluster.py", line 1185, in _RenewServerAndClientCerts
    bootstrap.GenerateClusterCrypto(True, # cluster cert
  File "/usr/share/ganeti/3.1/ganeti/bootstrap.py", line 139, in GenerateClusterCrypto
    utils.GenerateNewClientSslCert(clientcert_file, nodecert_file,
  File "/usr/share/ganeti/3.1/ganeti/utils/security.py", line 103, in GenerateNewClientSslCert
    x509.GenerateSignedSslCert(cert_filename, serial_no, signing_cert_filename,
  File "/usr/share/ganeti/3.1/ganeti/utils/x509.py", line 423, in GenerateSignedSslCert
    (key_pem, cert_pem) = GenerateSignedX509Cert(
                          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ganeti/3.1/ganeti/utils/x509.py", line 387, in GenerateSignedX509Cert
    req = OpenSSL.crypto.X509Req()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/cryptography/utils.py", line 76, in __getattr__
    obj = getattr(self._module, name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'OpenSSL.crypto' has no attribute 'X509Req'
```

rbott:
pyOpenSSL 26.3.0 removed the deprecated OpenSSL.crypto.X509Req (and friends), breaking `gnt-cluster init` and `gnt-cluster renew-crypto --new-node-certificates` on distributions shipping current pyOpenSSL. This change migrates all certificate and key handling to the `cryptography` package.

## Changes

 - `GenerateSignedX509Cert` builds certificates directly via `CertificateBuilder`; the CSR step is dropped (subject and public key were copied out of the CSR anyway).
 - Generated certificates are X.509 v3 with basicConstraints (CA only for the cluster certificate), keyUsage, so chain validation by OpenSSL/GnuTLS succeeds (legacy v1 certs were only accepted implicitly).
 - `X509CertKeyCheck` compares public-key DER bytes and returns a bool; the previous SSL.Context.check_privatekey() approach is removed (
 - `GetX509CertValidity` uses not_valid_before_utc/not_valid_after_utc where available (cryptography ≥ 42) and falls back to the naive properties otherwise - compatible with Debian Bookworm (38.x) and Trixie. This avoids the `CryptographyDeprecationWarning` on newer releases and maintains comaptibility for older distributions
 - The certificate digest keeps the pyOpenSSL format (XX:XX:…, SHA1 colon-hex) for ssconf compatibility; shared implementation in `utils.x509.FormatCertificateDigest`.
 - `tools/check-cert-expired` now also loads certificates via cryptography - not sure if that script/tool is of any relevance any more
 - The old code leaked PyOpenSSL Exceptions into regular Ganeti code outside of the x509 module - all occurences have been adapted (that's why there are also code changes outside of the x509 module and the related tests).

## Dependencies

 - pyOpenSSL remains required for the SSL/TLS connection layer (`OpenSSL.SSL`, `lib/http/__init__.py`).
 - `configure` now additionally/explicetely requires the cryptography module (`AC_PYTHON_MODULE(cryptography, t))

## Tests

 - New pytest suite `test/py/unit/utils/test_x509.py`: CA/key-usage properties, issuer/subject correctness, and an end-to-end mutual-TLS handshake (server requires client certificate, both sides validate against the cluster certificate).
 - Legacy tests (`ganeti.utils.x509_unittest`, `ganeti.tools.common_unittest`, `cmdlib/cluster_unittest`) migrated to `cryptography`; mocking now targets `security.VerifyCertificate` instead of `OpenSSL.crypto` internals.

Also with #1976 in place the QA suite now validates `gnt-cluster renew-crypto` (minus SSH-related knobs). I have already successfully tested this branch against Debian Trixie and Forky (earlier attempts uncovered problems with the new certificates like missing CA flags, X.509 v1 vs. v3 etc.).

## Backporting

I'd suggest to backport this fix to 3.1 as this is already relevant for current distributions. Since the next release will be 4.0, I think we will have to support `stable-3.1` for a while. I have not yet tested this but I am not sure if this will apply cleanly to 3.1. With #1905 and friends there might be some additional changes required for this PR to work.

Closes #1968


lib/utils/security.py
```diff
@@ -152,13 +171,17 @@ def IsCertificateSelfSigned(cert_filename):
...
   except Exception as err: # pylint: disable=W0703
     return (constants.CV_ERROR,
             "Failed to load X509 certificate %s: %s" % (cert_filename, err))
 
-  if cert.get_subject().CN == cert.get_issuer().CN:
+  subject_cn = _GetNameCn(cert.subject)
+  issuer_cn = _GetNameCn(cert.issuer)
+
+  if subject_cn == issuer_cn:
+
```
saschalucas: may be a newline to much here?

saschalucas (approved):
Thanks @rbott. Nice work.

I wonder if `IsCertificateSelfSigned()` should deserve it's own unit test, since the now fixed extra newline problem after `if subject_cn == issuer_cn:` wasn't spotted in unit-tests and QA?

Please decide on your own if an extra test would help. Other than that, LGTM.

rbott:
> Thanks @rbott. Nice work.
Thanks!

> 
> I wonder if `IsCertificateSelfSigned()` should deserve it's own unit test, since the now fixed extra newline problem after `if subject_cn == issuer_cn:` wasn't spotted in unit-tests and QA?

not sure I understand - the removed newline did not change the function's behaviour. Am I missing something here?

saschalucas:
> not sure I understand - the removed newline did not change the function's behaviour. Am I missing something here?

No you are not. I got the impression, that after `if something:` no new line is allowed. But this seems wrong. Sorry for the confusion.

Issue #1978, opened 2026-09-18
Large clusters impose problems on both luxid and wconfd


rbott:
As seen in #1972 and also #1654, larger clusters have negative impacts on both luxid and wconfd. Luxid seems to have trouble with e.g. very large lists of instances while wconfd chokes on config databases > 3MB and causes job timeouts (while waiting for config locks).

I created this issue to preverse the discussion that already happend in the (more or less related) PR #1972.

For completeness, here is the output of tests I ran against wconfd. Both with the current, built-in JSON handling and with the presumably much faster [aeson library](https://hackage.haskell.org/package/aeson). Aeson was implemented ("vibed") using AI and with my very limited Haskell knowledge I would neither say that it is production ready nor would I submit this as a PR. If we think aeson helps, someone else has to implement it. But nevertheless, we can see some numbers:

## Setup

- Ganeti built from source in the `ganeti/ci:trixie-py3` container
  (`--init`, `--privileged`, `nofile=1024`).
- vcluster (`tools/vcluster-setup`): single node for the idle sweep, 8 nodes
  (1 master + 6 master candidates) for the job-load sweep; fake hypervisor,
  diskless instances.
- `config.data` grown **offline** (stop wconfd, clone a template instance,
  restart wconfd/luxid) — reaches 5MB in seconds and distributes instances
  across node UUIDs.
- Latency measured directly at the wconfd unix socket: `Echo`, `ReadConfig`,
  `TryUpdateLocks` (shared BGL acquire + release).
- Load phase: 8 concurrent `gnt-instance modify -B memory=N` jobs (same
  exclusive-config-lock + full-serialize path as `INSTANCE_MIGRATE` /
  `INSTANCE_REPLACE_DISKS`), sampling wconfd latency while the queue drains.

Numbers are relative (container CPU scheduling); the signal is the trend.

## Idle latency vs. config size

| Config | Instances | Echo p50 | ReadConfig p50 | ReadConfig p95 | TryUpdateLocks p50 |
|---|---|---|---|---|---|
| 7.8 KB | 0 | 0.1 ms | 0.2 ms | 2.0 ms | 1.9 ms |
| 102 KB | 223 | 0.2 ms | 2.8 ms | 3.6 ms | 2.2 ms |
| 335 KB | 833 | 0.1 ms | 10.2 ms | 11.2 ms | 1.8 ms |
| 532 KB | 1133 | 0.1 ms | 17.5 ms | 18.9 ms | 1.7 ms |
[...]

PR #1980, opened 2026-09-21
Improve Python RPC communication

### LINKED ISSUE #1654: Jobs slowed down by RPC read loop

rbott:
This PR improves the Python side of RPC communication to backend daemons. It is based on findings and discussions in #1978 and #1654.

I used a local benchmark setup, combining vcluster with a Ganeti `config.data` database at around 5MB (a few thousand disk & nic-less instances running on the `fake` hypervisor) to compare numbers before and after the change.

## Improvements

I only addressed the Python-side of RPC communication here; so no new RPC calls or changed behaviours. 

1. **Transport receive path** (`lib/rpc/transport.py`): `Transport` and `FdTransport` now read 256K chunks instead of 4K and scan only newly read data for the LUXI message terminator instead of rescanning the whole accumulated buffer
2. **Config-lock retries** (`lib/config/__init__.py`): instead of hammering on wconfd we now use exponential backoff in a 30 second time window to reduce load and useless retries on a busy config daemon

## Measured impact

The numbers are taken from my local container/vcluster based tests:

| metric | before | after |
|---|---|---|
| recv syscalls per message, 1 MB / 3 MB / 5 MB | 57 / 186 / 305 | 29 / 75 / 110 |
| `lockConfig` (full-config download), 5 MB | 964 ms | 180 ms |
| mutate critical section e2e, 1 MB / 3 MB / 5 MB | 297 / 1389 / 3071 ms | 281 / 1047 / 2094 ms |
| 8-job batch wall time, 1 MB / 3 MB / 5 MB | 2.8 / 10.0 / 30.5 s | 2.3 / 5.7 / 12.1 s |
| 8-job per-job e2e p50, 5 MB | 26.7 s | 10.0 s |

We can see improvements across all tests/numbers. There is much more to gain in the future, but from here it starts getting more complicated (e.g. reducing the amount of times when Python clients lock & download the whole configuration just to compare the database serial). This will most likely lead to the addition of new RPC calls or alter the way existing RPC calls work and that has a much bigger blast radius then the changes in this PR.

## Tests

We now have tests for RPC message parsing and the backoff.

## AI Disclosure

AI was used to create the benchmark tooling around vcluster and also for the unit tests.


rbott:
I have started multiple QA runs to find possible side effects. Will report back once they finished.

*Update*: fake hypervisor on Trixie and Forky as well as KVM/DRBD/Plain/Sharedfile/File on Trixie finished successfully. The changes did have no impact on the QA runtimes (neither faster nor slower) but I also did not expect any changes here.

lib/config/__init__.py
```diff
@@ -72,6 +72,13 @@
 from ganeti import network
 
 
+# Exponential backoff (start, factor, limit) and budget for acquiring
+# WConfd's config lock; avoids hammering wconfd at peak contention
```
saschalucas, line 76: Wasn't before `time.sleep(random.random())` used in between? Than maybe that was not hammering?

Does it hurt to lose the random component? i.e. many clients started at the same time?

rbott: Actually there were two occurences of backend communication in this file. One used "hammer on wconfd every 100ms for 30 seconds straight" and the other one used "hammer on wconfd with a random delay of 100ms to 999ms **forever**" (it was an unbounded `while True` loop).

It seems that `wconfd` is stable enough to answer *at some point*, otherwise we should see quite some people report that jobs get seemling stuck (in that `while True` loop). OTOH it seems that this code path is only used via `ConfigManager` object and that again is only used within `gnt-cluster verify`.

In any case I think it's valid to align those two occurences and use the same retry mechanisms/settings. But I should have made clearer in the PR summary that one of the occurences changed its behaviour from "unbounded loop" to "bounded loop" 🙂

saschalucas: Thanks, for your explanation.

saschalucas (approved):
Thansk @rbott. LGTM.

